PR #2, opened 2026-05-14
feat: declare kind: memory for OpenClaw memory slot compatibility

rrriiiccckkk:
## Summary

Add `"kind": "memory"` to `openclaw.plugin.json` so the plugin can properly occupy OpenClaw's memory slot.

## Problem

OpenClaw uses `plugins.slots.memory` to select which plugin owns the memory slot. The slot resolution logic in `resolveMemorySlotDecisionShared` checks:

```javascript
if (!hasKind(params.kind, "memory")) return { enabled: true };
```

Without `kind: "memory"` in the manifest, the plugin is **silently skipped** — even if the user explicitly sets `plugins.slots.memory = "memory-tencentdb"`. This means the memory slot ends up empty (no auto-recall, no Dreaming, etc.).

## Fix

One-line addition to `openclaw.plugin.json`:

```json
{
  "id": "memory-tencentdb",
  "kind": "memory",  // ← added
  ...
}
```

## What this enables

Users can now replace the default memory stack with a single config:

```json
{
  "plugins": {
    "slots": {
      "memory": "memory-tencentdb",
      "contextEngine": "memory-tencentdb"
    }
  }
}
```

This gives them the full TencentDB Agent Memory pipeline (L0→L3 + Context Offload) as the sole memory provider, without needing Hindsight/memory-core/active-memory running in parallel.

## Compatibility

- `kind` is an optional field in the OpenClaw manifest spec
- Existing setups without `plugins.slots.memory` are unaffected (the default slot assignment still applies)
- No runtime code changes — manifest-only

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -1,12 +1,16 @@
 {
   "id": "memory-tencentdb",
+  "kind": "memory",
   "name": "Memory (TencentDB)",
   "description": "Four-layer memory system — auto-captures, structures, and profiles conversational knowledge",
   "activation": {
     "onStartup": true
   },
   "contracts": {
-    "tools": ["tdai_memory_search", "tdai_conversation_search"]
+    "tools": [
+      "tdai_memory_search",
+      "tdai_conversation_search"
+    ]
   },
   "configSchema": {
     "type": "object",
@@ -14,7 +18,10 @@
     "properties": {
       "storeBackend": {
         "type": "string",
-        "enum": ["sqlite", "tcvdb"],
+        "enum": [
+          "sqlite",
+          "tcvdb"
+        ],
         "default": "sqlite",
         "description": "存储后端：sqlite（本地 SQLite + sqlite-vec）或 tcvdb（腾讯云向量数据库）"
       },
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:
Hi @rrriiiccckkk, thanks for the investigation and the detailed analysis of the `resolveMemorySlotDecisionShared` code path! 👍

After internal discussion, we've decided to hold off on adding `kind: "memory"` to the manifest for now. Our current design intentionally does not claim the memory slot — the plugin registers its capabilities through hooks (`allowConversationAccess`) rather than occupying the `plugins.slots.memory` slot. This is a deliberate architectural choice to avoid conflicts with OpenClaw's built-in memory subsystem (memory-core).

That said, we understand the use case you're describing — if a user explicitly sets `plugins.slots.memory = "memory-tencentdb"`, they'd expect it to work. We'll evaluate whether to support this in a future version as an opt-in mode.

We'll keep this PR open for reference. Thank you for digging into the internals and raising this — it's a valuable discussion! 🙏

YOMXXX:
Marking this as superseded by #132 for #119.

#132 covers the same memory-slot compatibility issue and is more complete: it adds `kind: "memory"` to the manifest and feature-detects/calls `registerMemoryCapability({})` so OpenClaw can recognize the plugin as occupying the memory slot. This PR is manifest-only and currently conflicting.

Verification note for #132: `COREPACK_ENABLE_AUTO_PIN=0 pnpm build` and `git diff --check` passed locally. `COREPACK_ENABLE_AUTO_PIN=0 pnpm test` could not be used as a pass signal on that old branch because Vitest found no test files and exited 1.

YOMXXX:
Follow-up cleanup note: the newer canonical path is now #353.

The earlier triage pointed to #132, but #353 has since been locally verified and approved as the focused memory-capability PR. It covers the manifest-side memory capability declaration without keeping this old conflicting manifest-only branch open.

Recommendation: close this PR as superseded by #353.

PR #13, opened 2026-05-15
perf: eliminate N+1 metadata queries in vector search (searchL1Vector / searchL0Vector)

yuanrengu:
## Summary

Replace per-row metadata lookups with a single batch query in `searchL1Vector` and `searchL0Vector`.

Previously, after vec0 returned KNN results, each `record_id` was looked up individually via `stmtGetMeta.get(record_id)` — an N+1 pattern that issues up to `topK + buffer` (default 30) separate SQLite round-trips per vector search.

Now, all valid `record_id`s are collected and fetched in one `SELECT ... WHERE record_id IN (?, ...)` query via new `batchGetL1Meta` / `batchGetL0Meta` helpers, then resolved from a `Map` in O(1).

## Changes

- **`batchGetL1Meta(recordIds)`** — new private method, batch-fetches L1 metadata, returns `Map<string, meta>`
- **`batchGetL0Meta(recordIds)`** — new private method, batch-fetches L0 metadata, returns `Map<string, meta>`
- **`searchL1Vector`** — refactored to use batch lookup instead of N individual `stmtGetMeta.get()` calls
- **`searchL0Vector`** — refactored to use batch lookup instead of N individual `stmtL0GetMeta.get()` calls

## Behavior

No functional change. The zero-vector filtering, orphan detection, result ordering, and fault-tolerant behavior are all preserved:

| Aspect | Before | After |
|---|---|---|
| Metadata queries per search | N (up to 30) | 1 |
| Zero-vector filtering | `continue` in loop | `filter` before loop |
| Orphan handling | warn + skip | warn + skip (same) |
| Result ordering | vec0 distance order | vec0 distance order (same) |
| Error handling | catch → empty array | catch → empty Map → empty array (same) |

## Test plan

- [ ] Build passes (`npx tsdown`)
- [ ] Verify vector search returns correct results with existing data
- [ ] Verify orphan detection still logs warnings for records with vectors but no metadata
- [ ] Verify degraded mode (sqlite-vec unavailable) still returns empty results gracefully

### DIFF
--- a/src/core/store/sqlite.ts
+++ b/src/core/store/sqlite.ts
@@ -1140,32 +1140,27 @@ export class VectorStore implements IMemoryStore {
 
       if (rows.length === 0) return [];
 
-      const results: VectorSearchResult[] = [];
-
-      for (const { record_id, distance } of rows) {
-        // sqlite-vec returns null distance for zero vectors (cosine undefined when ‖v‖=0).
-        // Skip these — they are placeholder vectors from embedding-service-unavailable fallback.
+      // Filter out null/NaN distances (legacy zero-vector placeholders) first,
+      // then batch-fetch metadata in a single query to avoid N+1 round-trips.
+      const validRows = rows.filter(({ record_id, distance }) => {
         if (distance == null || Number.isNaN(distance)) {
           this.logger?.warn(
             `${TAG} [L1-search] record_id=${record_id} has null/NaN distance (likely zero vector) — skipping`,
           );
-          continue;
+          return false;
         }
+        return true;
+      });
 
-        const meta = this.stmtGetMeta.get(record_id) as
-          | {
-              content: string;
-              type: string;
-              priority: number;
-              scene_name: string;
-              session_key: string;
-              session_id: string;
-              timestamp_str: string;
-              timestamp_start: string;
-              timestamp_end: string;
-              metadata_json: string;
-            }
-          | undefined;
+      if (validRows.length === 0) return [];
+
+      // Batch metadata lookup: single query instead of N individual get() calls
+      const metaMap = this.batchGetL1Meta(validRows.map((r) => r.record_id));
+
+      const results: VectorSearchResult[] = [];
+
+      for (const { record_id, distance } of validRows) {
+        const meta = metaMap.get(record_id);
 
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:
Hi @yuanrengu, thanks for the PR! 👍

The N+1 query optimization in vector search is an interesting area to look at. We'll review the implementation details — including edge cases like empty `recordIds` and SQLite parameter limits — and get back to you shortly.

Thanks for the contribution! 🙏

yuanrengu:
@Maxwell-Code07 Thanks for taking a look! Let me know if any changes are needed.

withRiver:
Thanks for the PR! The idea of batching queries to avoid N+1 is a solid instinct, but in this specific case I don't believe it provides a real benefit. Here's why:

**1. This is embedded SQLite, not a network database**

The PR mentions reducing "round-trips," which is a valid concern for network databases (MySQL, PostgreSQL) where each query incurs network latency. However, we use better-sqlite3's synchronous in-process API — there are no network hops, no connection pool contention. Each `.get()` call is just a C++ binding invocation at the microsecond level.

**2. The original code already uses a pre-compiled prepared statement**

`this.stmtGetMeta` is compiled once at initialization and reused across all searches. Each call is just bind → step → reset with near-zero overhead.

The new batch implementation calls `db.prepare()` with a dynamically constructed SQL string on every search invocation. Since the number of placeholders varies, this means:
- String concatenation to build `IN (?, ?, ..., ?)` each time
- A fresh SQL parse/compile cycle on every call
- Loss of the prepared statement caching advantage

This is strictly worse than reusing a single pre-compiled statement.

**3. The actual scale is tiny**

`topK + buffer` defaults to 15. For local SQLite, 15 primary-key indexed lookups vs. a single `IN (...)` query yields a difference in the low-microsecond range. Internally, SQLite resolves `IN (...)` as individual rowid lookups anyway — there's no fundamental advantage.

---

Appreciate the effort and the clean implementation! If we ever move to a client-server database, this pattern would become highly relevant. For now, the existing approach is already well-optimized for our embedded use case.

yuanrengu:
@withRiver Appreciate the explanation! 🙏

PR #16, opened 2026-05-15
perf: replace O(n²) splice with O(n) filter in filterHeartbeatMessages

yuanrengu:
## Summary

Replace two levels of `splice()` calls in `filterHeartbeatMessages` with a filter-then-rebuild approach, reducing time complexity from O(n²) to O(n).

The function filters heartbeat tool calls from the messages array before LLM input. Previously it used `messages.splice(i, 1)` in the outer loop and `content.splice(j, 1)` in the inner loop — both O(n) per call due to element shifting.

### Changes

- Add `setMessageContentLocal(msg, newContent)` helper for consistent content reassignment across `msg.type === "message"` and normal message shapes
- Outer loop: replace reverse iteration + `splice(i, 1)` → forward iteration into `kept[]` array, skipping removed messages via `continue`
- Inner loop: replace reverse iteration + `splice(j, 1)` → `content.filter(block => !isHeartbeatToolUseBlock(block))`
- Rebuild `messages` array in-place from `kept` list at the end

### Behavior

No functional change — same heartbeat detection, same count semantics, same in-place mutation contract:

| Aspect | Before | After |
|--------|--------|-------|
| Heartbeat ID collection | O(n) scan | O(n) scan (same) |
| Message removal | splice(i, 1), O(n) each | skip via continue, O(1) |
| Content block removal | splice(j, 1), O(n) each | filter, O(n) total |
| Overall time | O(n²) worst-case | O(n) |
| Removal count | 1 per message | 1 per message (same) |

## Test plan

- [x] `npx tsdown` builds successfully
- [x] Heartbeat ID collection unchanged
- [x] Same messages are removed (tool/toolResult by tcId, assistant by content blocks)
- [x] Content block filtering preserves non-heartbeat blocks
- [x] In-place mutation contract preserved (callers receive mutated array)

### DIFF
--- a/src/offload/hooks/llm-input-l3.ts
+++ b/src/offload/hooks/llm-input-l3.ts
@@ -53,6 +53,14 @@ function getMessageRoleLocal(msg: any): string | undefined {
   return msg.role;
 }
 
+function setMessageContentLocal(msg: any, newContent: any[]): void {
+  if (msg.type === "message") {
+    msg.message.content = newContent;
+  } else {
+    msg.content = newContent;
+  }
+}
+
 function collectHeartbeatToolUseIds(msg: any): string[] {
   const role = getMessageRoleLocal(msg);
   if (role !== "assistant") return [];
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:
Hi @yuanrengu, thanks for the PR! 👍

The splice → filter optimization in `filterHeartbeatMessages` is an interesting direction. We'll review the implementation details and get back to you shortly.

Thanks for the contribution! 🙏

withRiver:
Thanks for the PR. The algorithmic analysis is correct, but this optimization has negligible real-world impact in our project. Here's why:

**a) Heartbeat messages are extremely rare**

`filterHeartbeatMessages` filters out tool_use messages containing `HEARTBEAT.md`. In a typical session, there are only single-digit to maybe a dozen such messages — not thousands. When n is this small, the difference between O(n²) and O(n) is negligible.

**b) `splice` has a very low constant factor**

V8 heavily optimizes `Array.splice` (internally it's essentially a `memmove`). For arrays of tens to hundreds of elements, splice can actually be faster than filter + rebuild because:
- `filter` allocates a new array (GC pressure)
- The final `messages.length = 0; for (...) messages.push(msg)` is itself an O(n) traversal and insertion pass

**c) This function is not on a hot path**

`filterHeartbeatMessages` runs once per LLM call. The LLM call itself takes seconds to tens of seconds, while this function processes an array of at most a few hundred to a few thousand messages. Even O(n²) splice on a few thousand messages completes in microseconds — entirely drowned out by LLM network latency.

**d) The new code introduces unconditional overhead**

```typescript
const kept: any[] = [];        // new allocation
// ...
messages.length = 0;           // clear original array
for (const msg of kept) messages.push(msg);  // refill
```

As long as any heartbeat ID exists (i.e. heartbeatIds.size > 0), the new code always iterates all messages and rebuilds the array — even if zero messages are actually removed. The old code only pays for splice when it actually hits a match, making its cost near-zero in the common case.

---

Given the above, this change doesn't deliver measurable performance gain. I'll pass on this one — closing without merge. Thanks for the effort though, and feel free to revisit if profiling data shows this function becoming a bottleneck in the future.

PR #25, opened 2026-05-16
docs: add seed (historical conversation import) section to README

leoge007:
### Summary

Documents the existing `seed` CLI command for importing historical conversations through the L0→L1→L2→L3 pipeline. The seed mechanism was implemented in `src/cli/commands/seed.ts` and `src/core/seed/` but had no user-facing documentation.

### What this adds

- **When to use seed** — backfill historical context after plugin install, migration from other systems
- **Input format** — Format A (`{ sessions: [...] }`) and Format B (top-level array) with examples
- **CLI usage** — command flags and options reference table
- **Seed pipeline** — what happens during execution (L0 capture → L1 extraction → storage)
- **Best practices** — selective import, start small, monitor L1 quality, LLM recommendations

Both English (README.md) and Chinese (README_CN.md) are updated.

### Related

The `memory-tencentdb-ctl` script also supports seed operations in Hermes mode. See `scripts/README.memory-tencentdb-ctl.md`.

---

Closes: (no related issue — this addresses an existing documentation gap)

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,95 @@ docker exec -it hermes-memory hermes
 
 ---
 
+### 3. Importing historical conversations (seed)
+
+Starting from v0.1.4, TencentDB Agent Memory ships with a `seed` command that processes historical conversations through the full L0→L1→L2→L3 pipeline, so previously exchanged conversations can produce structured memories, scene blocks, and persona refinements just like new ones.
+
+#### When to use seed
+
+- You installed the plugin after days of usage and want it to learn from past context.
+- You are migrating from another memory system.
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Hi @leoge007, thanks for this PR!

We appreciate you taking the time to document the seed command. However, we intentionally haven't included seed documentation in the public README yet — this feature is still in an early/experimental stage and the API surface may change. We'd like to stabilize it further before officially documenting it for end users. We'll keep this PR in mind and may revisit once the feature is more mature.
Thanks for your interest! 🙏

PR #27, opened 2026-05-16
feat(package): expose host-neutral library entrypoints

Dimon94:
## Description | 描述

Add a focused package/library-mode surface for host-neutral integrations:

- export typed package subpaths for `./core` and `./config`
- re-export `parseConfig` and config types from the core barrel
- generate declaration files for the package entrypoints
- remove stale package metadata that referenced missing CLI build outputs, so `npm pack` validates the publishable package
- keep the published OpenClaw extension pointed at the built runtime artifact and keep the source checkout unchanged

OpenClaw remains the default plugin path. This PR only makes the existing core easier to consume from package subpaths and keeps package metadata internally consistent.

## Related Issue | 关联 Issue

None.

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [x] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Local verification:

- `npm test -- src/core/public-exports.test.ts src/core/package-metadata.test.ts`
- `npm run build:plugin`
- `npm pack --dry-run` produced a 289.0 kB tarball, under the 512 kB CI size guard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,6 +334,37 @@ Debugging no longer means probing an opaque database — it becomes a determinis
 | Hybrid retrieval | BM25 + vector + RRF — supports both keyword and semantic recall |
 | Agent tools | `tdai_memory_search` / `tdai_conversation_search` |
 
+### Library Mode for Non-OpenClaw Hosts
+
+OpenClaw remains the default plugin entry point, but host-neutral embedders can
+use the same core without depending on OpenClaw runtime APIs:
+
+```ts
+import { TdaiCore, parseConfig } from "@tencentdb-agent-memory/memory-tencentdb/core"
+import type {
+  CompletedTurn,
+  HostAdapter,
+} from "@tencentdb-agent-memory/memory-tencentdb/core"
+
+const config = parseConfig({
+  capture: { enabled: true },
+  extraction: { enabled: false },
+})
+
+const core = new TdaiCore({
+  hostAdapter,
+  config,
+})
+await core.initialize()
+```
+
+`TdaiCore` obtains model execution through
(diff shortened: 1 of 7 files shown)


Maxwell-Code07:
Hi @Dimon94, thanks for this PR!

Since this changes the public package surface, we'll need to review it together with the product team to ensure it aligns with our intended API stability guarantees. Will get back to you after internal review.
Thanks for contributing! 🙏

Dimon94:
Thanks for the update. That makes sense. I agree this should be treated as a public API/stability decision rather than just an implementation detail.

I'll wait for your internal/product review before changing the PR further. If the preferred direction is to keep the public surface smaller, I'm happy to narrow this to the minimal subpath exports or adjust the package entry points to match the API stability boundary you want to support.

YOMXXX:
Triage note while cleaning up older PRs:

This still looks like a real, not-duplicated proposal: current main only exposes the root package export, so the host-neutral subpath export idea has not been superseded by the later adapter/checkpoint/FTS5 work.

I would not treat it as merge-ready yet, though:

- The branch is currently DIRTY and needs a rebase before any final review.
- The change expands the public package surface, so it still needs the product/API-boundary decision Maxwell called out earlier.
- After rebasing, it should get fresh verification for tests/build plus a package dry-run or equivalent check that the generated export paths resolve from the packed artifact.

Suggested cleanup path: either keep this open as the canonical public API proposal and rebase it for product review, or split the metadata/test cleanup from the actual subpath exports so maintainers can accept the mechanical pieces independently.

Issue #26, opened 2026-05-16
Discuss optional L2/L3 pipeline stage controls


Dimon94:
## Context

TencentDB Agent Memory currently runs the full L0 -> L1 -> L2 -> L3 pipeline by default, which is a good default for the OpenClaw plugin path.

Some host integrations may want to keep capture and L1 extraction enabled while deferring downstream L2 scene extraction or L3 persona generation. This is useful when the host wants to control when heavier summarization/profile work runs, or when it only needs structured L1 records.

## Proposal

Add two optional pipeline config fields:

- `pipeline.enableL2` defaults to `true`
- `pipeline.enableL3` defaults to `true`

When both are left unset, current behavior is preserved. When disabled, the scheduler skips those downstream stages and avoids creating a tool-enabled L2/L3 runner.

## Compatibility

- OpenClaw default behavior remains unchanged.
- Existing configs continue to work because both fields default to `true`.
- The change is host-neutral and not tied to any specific downstream project.

I opened this issue first because this is a product/behavior choice, not just a packaging or correctness fix.

Maxwell-Code07:
Hi @Dimon94, thanks for opening this discussion!
The idea of allowing hosts to selectively disable L2/L3 pipeline stages while keeping L0/L1 makes sense — we can see the use case for lightweight integrations that only need structured L1 records. We'll review this internally and discuss the product implications (e.g. how it interacts with existing offload and persona features). Will follow up once we have a direction.
Thanks! 🙏

Dimon94:
Thanks for the quick response. That direction makes sense.

The main requirement I want to preserve is for host integrations to be able to express a lightweight mode that stops after L1, without changing the default OpenClaw behavior. The exact API shape does not have to be `enableL2` / `enableL3` if that conflicts with product semantics around offload or persona features.

If the internal review prefers a narrower contract, I can adjust the PR accordingly. For example, this could be exposed as a single pipeline mode/preset instead of two independent booleans, or it could remain internal until the offload/persona interaction is fully settled.

I will wait for your direction before changing the implementation further.

Huang10q:
Thank you for your suggestion! However, in our design, L1, L2, and L3 constitute a single unified memory block, making it difficult to separate them within the official codebase. If needed, you may attempt modifying this section of the code.

Dimon94:
Thanks for confirming the product direction.

[...]

PR #28, opened 2026-05-16
feat(core): add optional pipeline stage controls

Dimon94:
## Description | 描述

Draft implementation for optional downstream pipeline controls discussed in #26:

- add `pipeline.enableL2` and `pipeline.enableL3`, both defaulting to `true`
- preserve current OpenClaw behavior when the options are unset
- skip L2/L3 scheduling when disabled
- avoid creating the tool-enabled L2/L3 runner when neither downstream stage is needed
- document the new config fields in the English and Chinese config tables

This is opened as a draft because it is a product/behavior choice. The implementation is intentionally small so maintainers can first decide whether this control belongs in the project.

## Related Issue | 关联 Issue

Discusses #26.

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [x] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Local verification:

- `npm test -- src/core/pipeline-controls.test.ts src/utils/pipeline-manager.test.ts`
- `npm run build:plugin`
- `git diff --cached --check` before commit

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,8 @@ docker exec -it hermes-memory hermes
 | `recall.strategy` | `"hybrid"` | Recall strategy: `keyword` / `embedding` / `hybrid` (RRF fusion, recommended) |
 | `recall.maxResults` | `5` | Number of items returned per recall |
 | `pipeline.everyNConversations` | `5` | Trigger an L1 memory extraction every N turns |
+| `pipeline.enableL2` | `true` | Enable L2 scene extraction after L1 |
+| `pipeline.enableL3` | `true` | Enable L3 persona generation after L2 |
 | `extraction.maxMemoriesPerSession` | `20` | Max memories extracted per L1 pass |
 | `persona.triggerEveryN` | `50` | Generate the user persona every N new memories |
 | `offload.enabled` | `false` | Whether to enable short-term compression |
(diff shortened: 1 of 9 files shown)


Maxwell-Code07:
Hi @Dimon94, thanks for this PR!

This involves a product/behavior decision, so we'll need to discuss it internally first. We'll align on the direction in #26, then review this implementation accordingly. Will get back to you once we have a conclusion.
Thanks! 🙏

Dimon94:
Thanks for clarifying. That makes sense: whether L2/L3 stages can be disabled is a product behavior decision, so aligning in #26 first is the right order.

I'll wait for the direction from #26 before updating this PR. Once the behavior is settled, I can adjust the implementation to match it, including narrowing/removing the config option if disabling these stages is not something the product should expose.

Maxwell-Code07:
Thank you for your contribution and proposal!
After internal discussion, in our design, L1, L2, and L3 form a unified memory block, and the official codebase does not support disabling individual downstream stages for now. Therefore, we will not merge this PR at this time.
If you have specific use cases, you are welcome to adapt the code on your own. We will continue to evaluate related capability planning in the future.
We’ll close this PR for now. Feel free to reach out if you have further questions.

Dimon94:
Thanks for the clear follow-up and for taking this through internal discussion.

Understood. If L1/L2/L3 are intended to remain one unified memory block in the official product model, then not merging this PR is the right call. I will keep this as a fork/local-integration adaptation rather than trying to push this behavior into the upstream codebase.

For future planning, the use case behind this PR was a host integration that wants lightweight capture plus L1 structured records while keeping downstream summarization/persona work under the host's own policy. But I agree that should not override the official memory semantics if the product boundary is unified L1-L3.

Thanks again for the review.

PR #29, opened 2026-05-16
fix(core): scope L1 memory search in stores

Dimon94:
## Description | 描述

Make L1 memory search scope filters apply before result truncation:

- accept `sessionKey` / `sessionId` in `searchMemories()` and Gateway `/search/memories`
- pass scope filters through the L1 store interfaces
- apply SQLite FTS filters in SQL and vector filters before final topK results are returned
- apply TCVDB filters through the query filter expression
- keep a caller-side scope check as a final defense
- add regression coverage for vector and FTS paths where unscoped topK truncation could otherwise hide in-scope memories

This is a general shared-backend correctness fix. It is not tied to any specific downstream host.

## Related Issue | 关联 Issue

None.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Local verification:

- `npm test -- src/core/tools/memory-search.test.ts`
- `npm test`
- `npm run build:plugin`
- `git diff --cached --check` before commit

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@
 
 ---
 
+## [Unreleased]
+
+### ✨ 改进
+
+- **L1 作用域过滤**：`searchMemories()` 支持 `sessionKey` / `sessionId`，Gateway `/search/memories` 支持 `session_key` / `session_id`，避免共享后端里的结构化记忆跨 session 召回。
+
+---
+
 ## [0.3.4] - 2026-05-12
 
 ### 🐛 修复
--- a/src/core/store/sqlite.ts
+++ b/src/core/store/sqlite.ts
@@ -28,6 +28,7 @@ import type {
   IMemoryStore,
   StoreCapabilities,
   L0Record,
+  L1QueryFilter,
   L1SearchResult,
   L1FtsResult,
   L0SearchResult,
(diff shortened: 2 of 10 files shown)


Maxwell-Code07:
Hi @Dimon94, thanks for this fix!

This is an important correctness improvement for scoped memory search. We'll review the implementation details and get back to you.
Thanks! 🙏

Dimon94:
Thanks for the quick response.

Agreed. The key correctness point here is that scope filtering should happen before topK truncation, otherwise valid in-scope memories can be hidden by out-of-scope results that happen to rank higher globally.

I tried to keep the fix store-local and defensive: each backend applies the scope as early as its query path allows, and the caller-side check remains as a final guard. I also added regression tests for both vector and FTS paths to cover the truncation-before-filtering failure mode.

I will wait for the implementation review and adjust anything that does not fit the backend conventions.

YOMXXX:
Triage note: #354 is now the canonical focused fix for #111 and has been locally verified/approved.

This PR overlaps on the same session-scoped L1 search problem, but it goes deeper by pushing scope filters into the store implementations and Gateway surface. That may still be useful as a follow-up optimization/correctness hardening, especially to avoid topK truncation before filtering, but it should be rebased after #354 and split away from the immediate #111 client-side isolation fix.

Suggested next step: re-scope this PR to the store-level pre-filtering improvement only, keep #354’s public/tool threading as the baseline, and add tests that demonstrate the pre-filtering behavior beyond what #354 already covers.

PR #31, opened 2026-05-16
fix(offload): preserve non-BMP characters in sanitizeText

### LINKED ISSUE #30: [Bug] sanitizeText silently strips emoji and CJK Extension B characters
### OpenClaw Version
N/A - code-trace bug, no runtime required.

### Plugin Version
0.3.4 (current main, package.json)

### Operating System
Any. The bug is in a JS regex and is OS/platform independent.

### Describe the bug
`UNSAFE_CHAR_RE` at `src/offload/storage.ts:165` includes the full surrogate range `[\uD800-\uDFFF]` but the regex has no `u` flag. Because JS strings are UTF-16, every non-BMP code point (emoji, CJK Extension B, math bold, etc.) is stored as a surrogate PAIR, and the regex strips each half independently. `sanitizeText` and `sanitizeJsonLine` therefore destroy any non-BMP character in tool params, tool results, and ref-md archives written by the offload pipeline.

### To Reproduce
```
node -e '
const re = /[�---\uD800-\uDFFF​-‏  ﻿]/g;
console.log(JSON.stringify("CJK ext-B \u{20BB7} here".replace(re, "")));'
// prints: "CJK ext-B  here"   (the 𠮷 character is gone)
```

Same problem for 🎉 (U+1F389), 𝐀 (U+1D400, math bold A), and every other supplementary character.

### Expected behavior
CJK Extension B 𠮷 and other non-BMP characters should pass through `sanitizeText` unchanged. The original intent of including `[\uD800-\uDFFF]` in the class is to strip LONE (malformed) surrogates, not to destroy well-formed supplementary characters.

### Error Logs / Screenshots
Silent data corruption - there is no log, the characters just disappear from the offloaded JSONL entries and ref-md files.

### Additional context
Affected callers (all in `src/offload/`):
- `sanitizeText`: `index.ts:429-430, 459-460` (every tool-call params/result)
- `sanitizeJsonLine`: `storage.ts:174` (every JSONL line written via `safeStringifyEntry`)
- `parseJsonlSafe`: `storage.ts:232` (second-pass strip on read)

Introduced in commit db8f3e51 (v0.3.3 release).

Suggested fix: add the `u` flag. With `u`, `[\uD800-\uDFFF]` matches only lone surrogates because paired surrogates have already been combined into a single code point.

akhilesharora:
`UNSAFE_CHAR_RE` at `src/offload/storage.ts:165` includes the full surrogate range `[\uD800-\uDFFF]` but is not declared with the `u` flag. Without `u`, JS regexes treat strings as UTF-16 code units, so every well-formed non-BMP code point (emoji, CJK Extension B, math bold, etc.) is a surrogate pair whose halves get stripped individually. `sanitizeText`, `sanitizeJsonLine`, and `parseJsonlSafe` therefore silently destroy these characters in tool params, tool results, and ref-md archives.

Reproduction with the exact regex copied from the file:

```
$ node -e '
const re = /[�---\uD800-\uDFFF​-‏  ﻿]/g;
console.log(JSON.stringify("CJK ext-B \u{20BB7} here".replace(re, "")));'
"CJK ext-B  here"
```

The 𠮷 (U+20BB7) is gone. Same problem for 🎉, 𝐀, and every other supplementary character.

Fix: add the `u` flag. With `u`, the surrogate range only matches lone (malformed) surrogates, which is the original intent. Well-formed pairs are treated as single code points and pass through.

Other entries in the class (replacement char, C0/C1 controls, zero-width chars, line separators, BOM) keep their behavior.

Added `src/offload/storage.test.ts` with six cases:
- plain ASCII survives
- emoji, CJK Extension B, math bold survive
- lone surrogates still stripped
- C0 and C1 controls still stripped
- zero-width and BOM still stripped
- non-string input passes through (existing typeof guard)

On main the non-BMP test fails (`expected 'emoji  here' to be 'emoji 🎉 here'`); with the fix all six pass.

Files touched:
- `src/offload/storage.ts`: one-character change (`/g` -> `/gu`).
- `src/offload/storage.test.ts`: new file, no runtime/network/db dependency.

Closes #30

### DIFF
--- a/src/offload/storage.test.ts
+++ b/src/offload/storage.test.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from "vitest";
+
+import { sanitizeText } from "./storage.js";
+
+describe("sanitizeText", () => {
+  it("preserves plain ASCII", () => {
+    expect(sanitizeText("hello world")).toBe("hello world");
+  });
+
+  it("preserves emoji and other non-BMP code points", () => {
+    // 🎉 = U+1F389, 𠮷 = U+20BB7 (CJK Extension B), 𝐀 = U+1D400 (math bold A).
+    // Each is a surrogate pair in UTF-16. Without the `u` flag, the
+    // [\uD800-\uDFFF] range in UNSAFE_CHAR_RE would strip each half
+    // independently and silently destroy these characters.
+    expect(sanitizeText("emoji \u{1F389} here")).toBe("emoji \u{1F389} here");
+    expect(sanitizeText("CJK ext-B \u{20BB7} here")).toBe(
+      "CJK ext-B \u{20BB7} here",
+    );
+    expect(sanitizeText("math bold \u{1D400} here")).toBe(
+      "math bold \u{1D400} here",
+    );
+  });
+
+  it("strips lone (malformed) surrogates", () => {
+    expect(sanitizeText("lone \uD800 surrogate")).toBe("lone  surrogate");
+    expect(sanitizeText("lone \uDC00 surrogate")).toBe("lone  surrogate");
+  });
+
+  it("strips C0 and C1 control characters", () => {
+    expect(sanitizeText("ctrlhere")).toBe("ctrlhere");
+    expect(sanitizeText("c1here")).toBe("c1here");
+  });
+
+  it("strips zero-width characters and BOM", () => {
+    expect(sanitizeText("a​b")).toBe("ab");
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Hi @akhilesharora, thanks for the fix!

This directly addresses #30 — adding the `u` flag is the correct minimal fix, and the test coverage (emoji, CJK Extension B, isolated surrogates, control chars) is thorough. We'll review and merge. 

Thanks! 🙏

PR #36, opened 2026-05-17
fix: include prepend context in gateway recall response

lazyayuan:
## Description | 描述

Fix a Gateway/Hermes `/recall` compatibility issue where L1 recall content in `prependContext` could be dropped.

The core recall logic already returns L1 memories in `prependContext`, but the Gateway `/recall` response previously only exposed `appendSystemContext` through the legacy `context` field. As a result, custom Agent integrations or the Hermes provider calling Gateway `/recall` directly could not receive the L1 memories returned in `prependContext`.

This PR:
- Adds `prependContext` and `appendSystemContext` to the Gateway `/recall` response.
- Preserves backward compatibility by building the legacy `context` field from `appendSystemContext` + `prependContext`.
- Updates Hermes `prefetch()` to prefer the split fields when present, and fall back to the legacy `context` field for old Gateway responses.
- Adds TypeScript and Python regression tests covering both new and legacy `/recall` response behavior.

## Related Issue | 关联 Issue

None

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

Verified with:
- `npm test`
- `npm run build:plugin`
- `python -m pytest hermes-plugin/memory/memory_tencentdb/tests -q`
- `git diff --check`

## Additional Notes | 其他说明

The legacy `context` field is preserved so existing clients can still read the full recall context. New clients can read `appendSystemContext` and `prependContext` separately.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,55 @@
+"""Pytest import helpers for the standalone plugin checkout."""
+
+from __future__ import annotations
+
+import importlib
+import importlib.util
+import os
+import pathlib
+import sys
+import types
+
+
+_PROJECT_ROOT = pathlib.Path(__file__).resolve().parent
+_HERMES_PLUGIN_ROOT = _PROJECT_ROOT / "hermes-plugin"
+if str(_HERMES_PLUGIN_ROOT) not in sys.path:
+    sys.path.insert(0, str(_HERMES_PLUGIN_ROOT))
+
+
+def _add_real_hermes_agent_root() -> None:
+    candidates = []
+    env_root = os.environ.get("HERMES_AGENT_ROOT")
+    if env_root:
+        candidates.append(pathlib.Path(env_root))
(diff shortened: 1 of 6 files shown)


Maxwell-Code07:
Hi @lazyayuan,

We've received your PR regarding the missing prependContext in Gateway recall response. Thanks for catching this! We'll review and follow up.

YOMXXX:
Triage note for duplicate/scope cleanup:

The Gateway recall-response compatibility part is now covered by #372, which is the verified focused MCP/Gateway bridge baseline. #372 adds split `prepend_context` / `append_system_context` recall response coverage, preserves Gateway compatibility, and was locally verified with focused tests, full test suite, build, and diff check.

This PR also includes Hermes Python prefetch fallback behavior, which may still be useful, but it should be split/rebased as a small Hermes-only follow-up on top of the shared Gateway recall-response shape rather than merged as a competing Gateway response implementation.

Issue #58, opened 2026-05-19
standalone L1 extraction can fail with DeepSeek due to thinking/tool interference


sirenexcelsior:
## Summary

In standalone/gateway mode, L1 extraction can fail against DeepSeek-compatible chat APIs even when the model itself is capable of returning valid JSON.

I reproduced this with DeepSeek chat-style endpoints and saw L1 extraction produce either:

- empty output (`output=0 chars`)
- non-JSON assistant text
- `The reasoning_content in the thinking mode must be passed back to the API.`

This leaves L0 recorded successfully while L1 stays empty for the same session.

## Reproduction

Environment:

- mode: standalone gateway
- LLM: DeepSeek-compatible OpenAI API
- model: `deepseek-chat`
- TDAI L1 extraction path (`StandaloneLLMRunner` with `enableTools=false`)

Observed behavior:

1. `POST /capture` records L0 successfully.
2. L1 runner starts for the same session.
3. L1 returns empty/non-JSON output, or errors with reasoning-related message.
4. No L1 records are stored.

Representative logs:

```text
[memory-tdai][l1-extractor] No JSON array found in extraction response
[memory-tdai][l1-extractor] [l1-debug] NO_JSON taskId=l1-extraction, rawLen=0, cleanedLen=0, rawFull=""
[memory-tdai] [standalone-runner] run() failed after 13911ms: The `reasoning_content` in the thinking mode must be passed back to the API.
```

## What seems to be happening

Two details in the standalone runner make L1 extraction brittle for DeepSeek:

1. When `enableTools=false`, the runner still exposes a read-only tool (`read_file`).
   For L1 extraction this is undesirable because the prompt expects pure JSON text output, but the model may still opportunistically attempt tool use.

2. There is no way on `origin/main` to configure standalone gateway requests with `disableThinking`, so DeepSeek-compatible backends may enter reasoning/thinking mode and produce output that is not compatible with the current L1 JSON parser.

## Suggested fix

I tested a minimal fix locally and it resolved the L1 write path for my DeepSeek setup:

1. In `src/adapters/standalone/llm-runner.ts`
   - when `enableTools=false`, do **not** pass tools at all
   - add optional `disableThinking?: boolean` to `StandaloneLLMConfig`
   - when `disableThinking=true`, inject `thinking: { type: "disabled" }` into the OpenAI-compatible request body

2. In `src/gateway/config.ts`
[...]

PR #59, opened 2026-05-19
Do not expose tools for standalone text-only LLMRunner tasks

sirenexcelsior:
## Summary

This PR removes tool exposure from the standalone `LLMRunner` when `enableTools=false`.

## Why

Some standalone text-only tasks, such as L1 extraction, expect plain text output and do not need tool access. Passing even a read-only tool subset can still encourage tool-calling behavior on OpenAI-compatible backends, which makes this path less stable than necessary.

With this change:

- `enableTools=true` keeps the existing sandboxed tool behavior
- `enableTools=false` sends no tools at all

## Scope

- `src/adapters/standalone/llm-runner.ts`

## Notes

This is intentionally narrow and only affects the standalone runner's text-only path.

### DIFF
--- a/src/adapters/standalone/llm-runner.ts
+++ b/src/adapters/standalone/llm-runner.ts
@@ -49,6 +49,8 @@ export interface StandaloneLLMConfig {
   maxTokens?: number;
   /** Request timeout in milliseconds (default: 120_000). */
   timeoutMs?: number;
+  /** Disable thinking/reasoning mode (DeepSeek: sets thinking.type=disabled in request body). */
+  disableThinking?: boolean;
 }
 
 // ============================
@@ -149,12 +151,6 @@ function createSandboxedTools(workspaceDir: string, logger?: Logger) {
   };
 }
 
-/** Read-only tool subset — used when enableTools=false to avoid empty tools rejection. */
-function createReadOnlyTools(workspaceDir: string, logger?: Logger) {
-  const all = createSandboxedTools(workspaceDir, logger);
-  return { read_file: all.read_file };
-}
-
 // ============================
 // StandaloneLLMRunner
 // ============================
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Thanks for this fix! We've noticed this issue as well. On certain reasoning models, the L1 extraction path can run into problems like what's described in [#58](https://github.com/Tencent/TencentDB-Agent-Memory/issues/58). We'll review this and follow up with you soon.

withRiver:
@sirenexcelsior 

After reviewing the relevant code, I can explain why a `read_file` tool was being passed in.

This project originally reused OpenClaw's `runEmbeddedPiAgent` to send requests to and receive responses from the LLM. If you pass `disableTools: true` to `runEmbeddedPiAgent`, it ends up sending `"tools": []` in the HTTP payload to the LLM API. However, some LLM providers (e.g. qwencode) reject requests where the `tools` field is an empty array. To work around this, the developer had no choice but to include a minimal, low-impact tool — `read_file` — so the array would never be empty.

`StandaloneLLMRunner` does not use OpenClaw's `runEmbeddedPiAgent`, yet it mistakenly carried over the `read_file` tool inclusion. For `StandaloneLLMRunner`, there is absolutely no need to pass `read_file` — doing so only degrades the LLM's output quality.

After I removed `read_file` based on your PR, the model was able to correctly extract L1 memories. So I believe removing `read_file` in the PR is the right fix. However, I don't think it's necessary to add the `disableThinking` configuration — that part of the PR feels a bit heavy-handed. For issue #58, simply removing `read_file` is sufficient.

sirenexcelsior:
Thanks a lot for the thorough investigation and explanation. This was very helpful, and I learned a lot from the background on why `read_file` was originally included in the first place.

Your clarification makes perfect sense: the workaround was necessary in the `runEmbeddedPiAgent` path to avoid sending `"tools": []`, but it should not have been carried over to `StandaloneLLMRunner`. I agree that, for the standalone path, removing `read_file` is the right and sufficient fix.

Based on your feedback, I have updated this PR to remove the `disableThinking` part and keep the change focused on the `read_file` removal only.

Thanks again for taking the time to trace this through so carefully.

withRiver:
@sirenexcelsior 
Could you please modify the title and content of this PR?  Focus on removing the tools for the standalone LLMRunner. After completing the above modifications, this PR can be merged.

Thanks for your contribution!

sirenexcelsior:
Updated the PR title and description to focus only on removing tools from the standalone LLMRunner text-only path. Thanks for the guidance.

PR #71, opened 2026-05-21
feat(recall): cap injected memory context

YOMXXX:
## Summary
- Add `recall.maxCharsPerMemory` and `recall.maxTotalRecallChars` with defaults of `0`, which do not alter existing behavior. Users can opt in by setting positive values to cap injected memory context.
- Apply the budget after L1 search and before `<relevant-memories>` injection, preserving score order while truncating oversized entries and dropping overflow.
- Document the new guards in README, README_CN, and `openclaw.plugin.json`.

## Test Plan
- `npm test`
- `npm run build`
- `npm pack --dry-run --json`
- `git diff --check`

## Notes
- Addresses #70 as an MVP focused on recall-time safeguards.
- `npx tsc --noEmit ...` still fails on current `upstream/main` due unrelated existing type issues in commander/OpenAI settings/offload strict initialization; this PR does not touch those areas.


Maxwell-Code07:
感谢您如此及时地解决了 #70 的问题！我们会内部评估，有结论后会尽快反馈。

package.json
withRiver, line 1: DO NOT modify this file, which might cause some problems. We have increased the limitation of Size Guard from 512KB to 2048KB.

YOMXXX:
Thanks for the review. I restored `package.json` to match `upstream/main` and removed the package manifest / extension entry changes from this PR.

Fresh local verification after the update:
- `npm test` ✅
- `npm run build` ✅
- `npm pack --dry-run` ✅ package size 647.5 KB, below the updated 2048 KB Size Guard limit.

package.json
YOMXXX: Resolved in 8031ec0: restored package.json to match upstream/main and removed the manifest / extension entry changes from this PR. Fresh checks are green, including Size Guard under the updated 2048KB limit.

src/config.ts
```diff
@@ -486,6 +490,8 @@ export function parseConfig(raw: Record<string, unknown> | undefined): MemoryTda
     recall: {
       enabled: bool(recallGroup, "enabled") ?? true,
       maxResults: num(recallGroup, "maxResults") ?? 5,
+      maxCharsPerMemory: normalizeNonNegativeInt(num(recallGroup, "maxCharsPerMemory"), 800),
```
withRiver: 1.  normalizeNonNegativeInt() is unnecessary here. 

All other config fields (maxResults, scoreThreshold, timeoutMs) use the simple num(...) ?? defaultValue pattern — these two fields don't need special treatment.

Defensive handling for negative/non-integer values is already done on the consumer side in auto-recall.ts's normalizeBudgetLimit() (<= 0 → treated as disabled, Math.floor() for rounding). No need to duplicate that at the parse layer. 


2. The defaults should be 0 (disabled), not 800 / 3000.

This is a new feature — the default behavior should remain backward-compatible. Existing users upgrading should not suddenly see their recall results being truncated or dropped. Those who need this guard can opt in by explicitly setting the values.

openclaw.plugin.json
```diff
@@ -69,6 +69,8 @@
         "properties": {
           "enabled": { "type": "boolean", "default": true, "description": "是否启用自动召回" },
           "maxResults": { "type": "number", "default": 5, "description": "召回最大结果数" },
+          "maxCharsPerMemory": { "type": "number", "default": 800, "description": "单条 L1 记忆注入的最大字符数；填 0 表示不限制" },
```
withRiver: default should be zero

README.md
```diff
@@ -259,6 +259,8 @@ docker exec -it hermes-memory hermes
 | `storeBackend` | `"sqlite"` | Storage backend: `sqlite` |
 | `recall.strategy` | `"hybrid"` | Recall strategy: `keyword` / `embedding` / `hybrid` (RRF fusion, recommended) |
 | `recall.maxResults` | `5` | Number of items returned per recall |
+| `recall.maxCharsPerMemory` | `800` | Max characters injected for one recalled L1 memory; `0` disables this guard |
```
withRiver: default should be zero

README_CN.md
```diff
@@ -262,6 +262,8 @@ docker exec -it hermes-memory hermes
 | `storeBackend` | `"sqlite"` | 存储后端：`sqlite` |
 | `recall.strategy` | `"hybrid"` | 召回策略：`keyword` / `embedding` / `hybrid`（RRF 融合，推荐） |
 | `recall.maxResults` | `5` | 每次召回条数 |
+| `recall.maxCharsPerMemory` | `800` | 单条 L1 记忆注入的最大字符数；`0` 表示不限制 |
```
withRiver: default should be zero

src/core/hooks/auto-recall.test.ts
withRiver, line 1: There are no existing test files under src/core/hooks/. The budget logic in normalizeBudgetLimit / truncateRecallLine is straightforward enough that the config parsing coverage in src/config.test.ts plus manual verification should suffice for this PR. Let's remove this file for now.

src/config.test.ts
withRiver, line 1: Remove this test file.

src/core/hooks/auto-recall.ts
```diff
@@ -706,6 +709,83 @@ function formatMemoryLine(m: FormatableMemory): string {
...
+      ? truncateRecallLine(line, maxCharsPerMemory)
+      : line;
+    if (perMemoryBounded !== line) truncatedCount++;
+
+    if (!maxTotalRecallChars) {
+      budgeted.push(perMemoryBounded);
+      continue;
+    }
+
+    const separatorChars = budgeted.length > 0 ? 1 : 0;
```
withRiver: The 1 here implicitly assumes the join separator is "\n" (line 212). Consider adding a brief comment like // "\n".length to make the coupling explicit

src/core/hooks/auto-recall.ts
```diff
@@ -706,6 +709,83 @@ function formatMemoryLine(m: FormatableMemory): string {
...
+      droppedCount += lines.length - i;
+      break;
+    }
+
+    if (perMemoryBounded.length > remainingChars) {
+      if (remainingChars >= MIN_TRUNCATED_RECALL_LINE_CHARS) {
+        const totalBounded = truncateRecallLine(perMemoryBounded, remainingChars);
+        budgeted.push(totalBounded);
+        usedChars += separatorChars + totalBounded.length;
+        if (totalBounded !== perMemoryBounded) truncatedCount++;
```
withRiver: truncatedCount can be incremented twice for the same memory — once at line 734 (per-memory truncation) and again here (total-budget truncation). This makes the debug log misleading (e.g. reporting truncated=3 when only 2 distinct lines were affected). Consider tracking truncated lines with a Set or a boolean flag per iteration instead.

src/core/hooks/auto-recall.ts
```diff
@@ -706,6 +709,83 @@ function formatMemoryLine(m: FormatableMemory): string {
...
+      } else {
+        droppedCount++;
+      }
+      droppedCount += lines.length - i - 1;
+      break;
```
withRiver, line 760: The droppedCount accumulation is split across the if/else branch and the line after it, making the logic harder to follow. Suggestion:
```ts
if (perMemoryBounded.length > remainingChars) {
  const canFit = remainingChars >= MIN_TRUNCATED_RECALL_LINE_CHARS;
  if (canFit) {
    budgeted.push(truncateRecallLine(perMemoryBounded, remainingChars));
  }
  droppedCount += lines.length - i - (canFit ? 1 : 0);
  break;
}
```
Single accumulation point, same semantics, easier to reason about.

README.md
YOMXXX: Resolved in c588558: README default is now `0`.

openclaw.plugin.json
YOMXXX: Resolved in c588558: plugin schema default is now `0`.

src/config.ts
YOMXXX: Resolved in c588558: recall budget config now follows the existing `num(...) ?? 0` pattern, defaults are disabled, and `normalizeNonNegativeInt()` was removed.

src/config.test.ts
YOMXXX: Resolved in c588558: removed `src/config.test.ts` from this PR.

src/core/hooks/auto-recall.ts
YOMXXX: Resolved in c588558: introduced `RECALL_LINE_SEPARATOR` and use its `.length` for budget accounting, matching the final join separator.

src/core/hooks/auto-recall.ts
YOMXXX: Resolved in c588558: `truncatedCount` now uses a per-iteration flag, so one emitted memory line is counted at most once even if both per-memory and total-budget truncation apply.

src/core/hooks/auto-recall.test.ts
YOMXXX: Resolved in c588558: removed `src/core/hooks/auto-recall.test.ts` from this PR.

src/core/hooks/auto-recall.ts
YOMXXX: Resolved in c588558: collapsed dropped-count handling to a single accumulation point in the total-budget overflow branch.

README_CN.md
YOMXXX: Resolved in c588558: README_CN default is now `0`.

withRiver:
LGTM. Thanks for your contribution!

By the way, there’s another issue #3 related to L1 recall. If you’re interested, feel free to take a look.

PR #76, opened 2026-05-23
feat(capture): make timestamps timezone configurable

### LINKED ISSUE #75: [Feature] Make timestamp timezone configurable in CaptureConfig
## Feature Description

TDAI stores  (L0) and  (L1) timestamps using  which always outputs UTC. These timestamps should respect a configurable timezone.

## Motivation

Timestamps in local time are more useful for debugging, user-facing display, and data analysis. UTC is correct for storage but the system should allow users to configure which timezone is used when generating these timestamps. Currently  already has a  helper with a hardcoded  offset — the natural next step is to expose this as a config field.

## Proposed Solution

Add a  field to  (in ):

Default to  (CST) to match the existing  behavior. Users in other timezones can override.

The three hardcoded UTC sites are:
-  —  → 
-  —  → 
-  —  → 

These would read from  instead of using UTC directly.

## Alternatives Considered

1. **Environment variable ** — Node.js respects this but requires system-level config, not user-configurable per-deployment.
2. **Hard-coding  into the three sites** — what the existing  helper already does, but not configurable.

## Additional Context

The  file already contains a  function (with ) that could serve as the basis for the implementation. This would be a minimal, low-risk addition with a clear default.

YOMXXX:
## Summary
- add `capture.timezoneOffsetMinutes` with UTC+08:00 default and schema/docs coverage
- apply the configured offset to L0 JSONL timestamps/shards, L0 vector metadata, L1 extraction prompts, and L2 scene extraction prompts
- make SQLite L0 cursor filtering instant-aware for offset timestamps
- keep the npm package under the existing size guard by shipping built runtime artifacts only

Closes #75.

## Verification
- `npm test`
- `npm run build`
- `npm pack --dry-run` (package size: 340.7 kB, unpacked: 1.3 MB)


Maxwell-Code07:
Thanks for the quick implementation addressing #75! We'll review the approach and get back to you.

Maxwell-Code07:
We have fixed this issue in https://github.com/TencentCloud/TencentDB-Agent-Memory/pull/129，Thank you for attention！

PR #79, opened 2026-05-23
fix(plugin): add fallback for runtime.state.resolveStateDir on OpenClaw >=2026.5.x

### LINKED ISSUE #78: fix: runtime.state.resolveStateDir undefined on OpenClaw >=2026.5.x
## Summary

TDB 0.3.5 (built with OpenClaw SDK 2026.3.13) fails to register on OpenClaw 2026.5.20 because `api.runtime.state` is `undefined` during plugin registration.

## Error

```
[plugins] memory-tencentdb failed during register: 
```

## Root Cause

`index.ts` calls `api.runtime.state.resolveStateDir()` at two points during registration:

1. L220: `const pluginDataDir = path.join(api.runtime.state.resolveStateDir(), "memory-tdai");`
2. L826: `stateDir: api.runtime.state.resolveStateDir()`

On OpenClaw 2026.5.20, the `runtime.state` object is not yet populated at the point TDB calls it, even though the TypeScript types declare it exists.

## Fix

Add optional chaining with a sensible fallback. Since `resolveStateDir()` is well-defined (returns `~/.openclaw` or `$OPENCLAW_STATE_DIR`), the fallback can be `os.homedir() + "/.openclaw"`:

```typescript
// Before

// After
const stateDir = api.runtime.state?.resolveStateDir?.() ?? 
  path.join(os.homedir(), ".openclaw");
const pluginDataDir = path.join(stateDir, "memory-tdai");
```

This is safe because `resolveStateDir()` is a pure directory resolution with no side effects. The call sites that don't have access to `process.env` can hardcode the homedir fallback.

## Environment

- TDB: 0.3.5
- OpenClaw: 2026.5.20
- macOS 15.5 (Apple Silicon)
- Plugin SDK compat declared: `pluginApi >=2026.3.13`

## Related

Similar to #57 (CLI registration break after OpenClaw upgrade) — this is the same class of SDK compat issue.

leoge007:
## Summary

Fixes #78

## Problem

TDB 0.3.5 fails to register on OpenClaw 2026.5.20 because `api.runtime.state` is `undefined` during plugin registration:

```
```

The TypeScript types declare `runtime.state` exists, but at registration time on 2026.5.x the object isn't yet populated.

## Fix

Add optional chaining with `os.homedir()` fallback at the two call sites.

`resolveStateDir()` returns `~/.openclaw` (or `$OPENCLAW_STATE_DIR`), so the fallback is well-defined and has no side effects.

## Testing

- ✅ Tested on OpenClaw 2026.5.20 with macOS 15.5 (Apple Silicon)
- ✅ Plugin registers successfully
- ✅ `tdai_memory_search` and `tdai_conversation_search` work correctly
- ✅ CLI commands register (`openclaw memory-tdai --help`)

## Changes

2 lines changed in `index.ts`:
- L220: `const pluginDataDir = path.join((api.runtime.state?.resolveStateDir ? ... : require('os').homedir() + '/.openclaw'), "memory-tdai");`
- L826: `stateDir: (api.runtime.state?.resolveStateDir ? ... : require('os').homedir() + '/.openclaw'),`

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -217,7 +217,7 @@ export default function register(api: OpenClawPluginApi) {
   }
 
   // Resolve plugin data directory via runtime API (avoid importing internal paths directly)
-  const pluginDataDir = path.join(api.runtime.state.resolveStateDir(), "memory-tdai");
+  const pluginDataDir = path.join((api.runtime.state?.resolveStateDir ? api.runtime.state.resolveStateDir() : require('os').homedir() + '/.openclaw'), "memory-tdai");
   initDataDirectories(pluginDataDir);
   api.logger.debug?.(`${TAG} Data dir: ${pluginDataDir} (all subdirectories initialized)`);
 
@@ -823,7 +823,7 @@ export default function register(api: OpenClawPluginApi) {
       registerMemoryTdaiCli(memoryTdai, {
         config,
         pluginConfig: api.pluginConfig,
-        stateDir: api.runtime.state.resolveStateDir(),
+        stateDir: (api.runtime.state?.resolveStateDir ? api.runtime.state.resolveStateDir() : require('os').homedir() + '/.openclaw'),
         logger: cliLogger,
       });
     },


Maxwell-Code07:
Thanks for the fix and the thorough testing! The root cause analysis is clear and the change is well-scoped. We'll review internally and get back to you with feedback.

YOMXXX:
Reviewer triage notes from a local pass:

Verification I ran locally on this branch:
- `npm test` ✅
- `npm run build` ✅
- `npm pack --dry-run` ⚠️ produces a 646.2 kB tarball / 139 files, so this would fail the current 512 KB Size Guard if the full CI matrix runs. Same package file-list cleanup as #76/#71 is needed unless that lands first.

Functional risk to address before merge:
- The fallback branch uses bare `require('os')` inside `index.ts`, which is an ESM source file and is also shipped as source today. This fallback is exactly the path that runs when `api.runtime.state` is missing, so source-mode execution can turn the original error into `ReferenceError: require is not defined`.

Suggested safer shape:
```ts
import { homedir } from "node:os";

function resolveOpenClawStateDir(api: OpenClawPluginApi): string {
  return api.runtime.state?.resolveStateDir?.() ?? path.join(homedir(), ".openclaw");
}
```

Then reuse that helper for both `pluginDataDir` and CLI `stateDir`. This keeps the fix small, avoids duplicated expressions, and is safe in ESM/source runtime.

YOMXXX:
I opened #85 as an ESM-safe alternative for the same #78 failure mode. It keeps the fallback small, adds a focused unit test, and CI is green including Size Guard. Reviewers can compare against this PR when deciding which path to merge.

Xuruida:
Thanks a lot for the fix and the detailed investigation! 🙏

We’ve reviewed both approaches (#79 vs #85) internally, and we’re going to move forward with **#85** as the canonical fix.

The main reasons are:
- **Test coverage:** #85 includes solid unit tests covering the `require()` / ESM edge cases, which gives us higher confidence for long-term stability.
- **Architecture alignment:** It provides a cleaner ESM-safe implementation and avoids mixing CommonJS semantics in our source ESM context.

That said, the diagnosis here was spot-on and really helped us confirm the root cause quickly.  
Closing this PR in favor of #85.

Thanks again @leoge007 and @YOMXXX for driving this forward — really appreciate the collaboration! 🚀

PR #85, opened 2026-05-24
fix(plugin): add OpenClaw state dir fallback

### LINKED ISSUE #78: fix: runtime.state.resolveStateDir undefined on OpenClaw >=2026.5.x

YOMXXX:
## Summary
- add an ESM-safe helper for resolving the OpenClaw state directory
- fall back to `~/.openclaw` when `api.runtime.state` is unavailable during plugin registration
- reuse the same resolved state dir for plugin data and CLI registration
- keep the npm tarball under Size Guard by publishing built runtime artifacts only

Fixes #78.

This is intended as a safer alternative to #79: it avoids bare `require("os")` in the ESM source path and adds a focused unit test for the fallback behavior.

## Verification
- `npm test`
- `npm run build`
- `npm pack --dry-run` (package size: 340.2 kB, unpacked: 1.3 MB)


Xuruida:
Adopting this fix over #79 mainly because of the included test coverage and ESM-safe structure.  
Thanks to @leoge007 for the initial diagnosis and to @YOMXXX for the complete fix! 🚀

PR #86, opened 2026-05-24
fix(runner): avoid duplicate system prompt on new OpenClaw

YOMXXX:
## Summary
- avoid passing `extraSystemPrompt` on OpenClaw versions that support `agents.defaults.systemPromptOverride`
- preserve the legacy `extraSystemPrompt` fallback for unknown or older OpenClaw versions
- add unit coverage for the version gate
- keep the npm tarball under Size Guard by publishing built runtime artifacts only

This is intended as a compatibility-preserving alternative to #84.

## Verification
- `npm test`
- `npm run build`
- `npm pack --dry-run` (package size: 340.4 kB, unpacked: 1.3 MB)


Maxwell-Code07:
Thanks for this compatibility-aware fix! We'll review and run the test suite internally. Will update here with feedback.

YOMXXX:
Closing this older duplicate-system-prompt PR in favor of #241. #241 carries the same fix on the current code path, keeps the review surface smaller, and is already clean against main. This PR remains useful as the original compatibility context, but #241 is the better merge target now.

PR #109, opened 2026-05-29
fix(recall): preserve non-BMP characters when truncating recalled context

### LINKED ISSUE #108: [Bug] truncateRecallLine splits surrogate pairs, corrupting non-BMP chars in recalled context
N/A - static code defect in string truncation, independent of the OpenClaw runtime version.

### Plugin Version | 插件版本
0.3.6 (current `main`, 438869b)

### Operating System | 操作系统
Any - pure JavaScript string handling, platform-independent.

### Describe the bug | 问题描述
`truncateRecallLine` (`src/core/hooks/auto-recall.ts`), added in #71 to cap injected recall context, truncates by UTF-16 code unit using `line.length` and `line.slice`. When `recall.maxCharsPerMemory` or `recall.maxTotalRecallChars` is set and the cut lands between the two halves of a surrogate pair (any non-BMP character: emoji, CJK Extension B, etc.), the truncated line keeps a lone surrogate. When that line is UTF-8 encoded for the model request it becomes U+FFFD, so the recalled memory is corrupted in the injected `<relevant-memories>` block.

Same class as #30 (fixed in #31, "preserve non-BMP characters in sanitizeText"); the new truncation path reintroduced it.

### To Reproduce | 复现步骤
1. Enable a recall cap (`recall.maxCharsPerMemory`); both caps default to 0 (off).
2. Recall a memory whose text has a non-BMP character at the cut boundary, e.g. `"Client prefers email 😀"` with a 22-character cap.
3. The injected line ends in a lone high surrogate, which encodes to `�`.

Node check of the function's logic (both code branches), buggy vs fixed:

    BUGGY  corrupt(U+FFFD)? true
    FIXED  corrupt(U+FFFD)? false   (non-BMP char kept whole)

### Expected behavior | 预期行为
Truncation must not split a surrogate pair: a non-BMP character is kept whole or dropped whole, never corrupted to U+FFFD - the same guarantee #31 established for `sanitizeText`.

### Additional context | 补充信息
Both call sites (per-memory and total-budget) route through this one function, so the fix is local to it. Caps are opt-in (default 0), so this affects users who enabled recall budgeting and store non-BMP content.

akhilesharora:
## Description | 描述
`truncateRecallLine` (`src/core/hooks/auto-recall.ts`) truncated recalled-memory lines by UTF-16 code unit (`line.length` / `line.slice`). A cap landing between the halves of a surrogate pair left a lone surrogate, which becomes U+FFFD when the line is UTF-8 encoded for the request, corrupting any non-BMP character (emoji, CJK Ext-B) in the injected context. Switched the count and slice to code points via `Array.from`. Both recall call sites (per-memory and total-budget) route through this one function. Same defect class as #30/#31 ("preserve non-BMP characters in sanitizeText"), reintroduced by the #71 budget path.

## Related Issue | 关联 Issue
Fix #108

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明
Verified with Node on both branches of the function: a cap landing mid-pair produces U+FFFD before the change and an intact character after; ASCII/BMP input is byte-for-byte unchanged. CI has no test step and the repo currently ships no test files, so no test was added.

### DIFF
--- a/src/core/hooks/auto-recall.ts
+++ b/src/core/hooks/auto-recall.ts
@@ -782,11 +782,15 @@ function normalizeBudgetLimit(value: number | undefined): number | undefined {
 }
 
 function truncateRecallLine(line: string, maxChars: number): string {
-  if (line.length <= maxChars) return line;
+  // Count and slice by code point, not UTF-16 code unit, so a cut never lands
+  // between the halves of a surrogate pair (which would corrupt a non-BMP
+  // character to U+FFFD when the line is UTF-8 encoded for the request).
+  const cps = Array.from(line);
+  if (cps.length <= maxChars) return line;
   if (maxChars <= RECALL_TRUNCATION_SUFFIX.length) {
-    return line.slice(0, maxChars);
+    return cps.slice(0, maxChars).join("");
   }
-  return `${line.slice(0, maxChars - RECALL_TRUNCATION_SUFFIX.length).trimEnd()}${RECALL_TRUNCATION_SUFFIX}`;
+  return `${cps.slice(0, maxChars - RECALL_TRUNCATION_SUFFIX.length).join("").trimEnd()}${RECALL_TRUNCATION_SUFFIX}`;
 }
 
 /**


Maxwell-Code07:
Thank you for your contribution. We will review it and provide timely feedback. Thank you for your attention to our project!

Xuruida:
LGTM — thanks @akhilesharora for the fix! 🙏

Same class of bug as #31 (sanitizeText surrogate split), correctly addressed with `Array.from()` code-point-level slicing. Minimal change, ASCII/BMP behavior unchanged.

Closes #108.

PR #112, opened 2026-05-29
fix(core): reliable L3 persona writes for weak tool-calling models + warn on silent embedding misconfig

Lorenzocolucci:
Two small, independent core fixes I hit while running the standalone gateway with an OpenAI-compatible model (Moonshot/Kimi) for extraction.

## 1. L3 persona is silently dropped with weak tool-calling models

The persona prompt instructed the model to use the `write`/`edit` tools (OpenClaw names), but the standalone runner exposes `write_to_file`/`replace_in_file`. With a model that has weaker tool-calling, it returned the persona as plain text and called no tool, so the result was discarded and `persona.md` was never written.

This PR:
- makes the persona prompt **tool-dialect aware** (`openclaw`: write/edit — unchanged default; `standalone`: write_to_file/replace_in_file);
- optionally **forces the write tool** (`toolChoice: "required"` + stop on `write_to_file`) so reliable tool-calling is guaranteed when needed;
- adds a **text fallback**: if the model still returns the persona as text, it's persisted instead of lost.

The non-forced branch is unchanged; I preserved main's conditional `tools` spread in the merge.

## 2. Silent degradation when an embedding provider is misconfigured

When a remote embedding provider is configured but the service fails to initialize (e.g. an empty API key), search silently fell back to keyword-only with no signal. This adds a **WARN at startup** so the misconfiguration is visible instead of silent.

## Testing

Built against current `main` (build green). Verified the persona path writes `persona.md` via the tool (fallback count 0) and the WARN fires when the embedding key is missing.

Happy to adjust naming or split these into two PRs if you'd prefer.

### DIFF
--- a/src/adapters/standalone/llm-runner.ts
+++ b/src/adapters/standalone/llm-runner.ts
@@ -18,7 +18,7 @@
 
 import fsPromises from "node:fs/promises";
 import path from "node:path";
-import { generateText, tool, stepCountIs, jsonSchema } from "ai";
+import { generateText, tool, stepCountIs, hasToolCall, jsonSchema } from "ai";
 import { createOpenAI } from "@ai-sdk/openai";
 import { report } from "../../core/report/reporter.js";
 import type {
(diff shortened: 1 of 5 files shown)


Maxwell-Code07:
Thanks for the PR! Both fixes address real pain points — the L3 persona silent discard on weaker models and the invisible embedding misconfig. We'll review internally. Appreciate the thorough testing notes!

YOMXXX:
Scope note:

Both fixes described here are valid problem areas, but this PR bundles two independent behaviors:

1. L3 persona write reliability for weak tool-calling / standalone tool dialects.
2. Startup warning when a configured embedding provider fails and search silently degrades to keyword-only.

Those should be split so each can be reviewed, tested, and merged independently. I would not treat this as a duplicate of the session-isolation work (#354/#29), but it should not stay as a combined core-fixes PR either. A focused persona-write PR plus a focused embedding-misconfig diagnostics PR would be much easier to validate.

PR #128, opened 2026-06-02
fix: pass disableTools:true when enableTools=false in CleanContextRunner

withRiver:
## Summary

When `enableTools=false` (the default for L1 extraction / L1 dedup tasks), `CleanContextRunner` now passes `disableTools: true` to `runEmbeddedPiAgent`, ensuring no tool definitions are sent to the LLM API.

## Motivation

Previously, `disableTools` was hardcoded to `false` as a workaround for a specific provider (qwencode) that rejects an empty `tools[]` array. As a result, even pure text extraction tasks would include a `read_file` tool definition in the API request.

This caused two problems:

1. **Wasted tokens** — the tool schema occupies context space on every LLM call that never intends to use tools.
2. **Output quality degradation** — some models, upon seeing an available tool, may attempt tool calls instead of producing the expected direct text output.

## Changes

| File | Change |
|------|--------|
| `src/utils/clean-context-runner.ts` | 1. `disableTools` changed from `false` to `!this.options.enableTools`<br>2. `tools.allow` changed from `["read"]` to `[]` when `enableTools=false`<br>3. Updated file header comment to reflect new behavior |

## Migration Note

Users on providers that reject requests without a `tools` field or with an empty `tools[]` (e.g. qwencode) should switch to `StandaloneLLMRunner` via LLM configuration. `StandaloneLLMRunner` (powered by Vercel AI SDK) natively handles the `enableTools=false` case by simply omitting the `tools` parameter from the request.

### DIFF
--- a/src/utils/clean-context-runner.ts
+++ b/src/utils/clean-context-runner.ts
@@ -5,7 +5,7 @@
  * Guarantees:
  * 1. Blank conversation history (temporary session file)
  * 2. Independent system prompt (only the task prompt)
- * 3. No tool calls (tools restricted to minimal read-only set to avoid empty tools[] rejection by some providers)
+ * 3. No tool calls when enableTools=false (disableTools:true — no tool definitions sent to API)
  * 4. No contamination from the main agent's context
  */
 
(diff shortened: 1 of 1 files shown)


Xuruida:
LGTM — clean fix, well explained. Thanks @withRiver for catching this! 🙏

`disableTools: !this.options.enableTools` is the correct derivation. L1 extraction / L1 dedup will no longer leak tool schemas into the API request, saving tokens and avoiding unintended tool-call attempts from models that get "distracted" by available tools.

Migration note for qwencode users is clear and helpful.

Issue #118, opened 2026-05-31
after_tool_call hook: event.messages undefined, L3 compression broken in OpenClaw 5.27


zhao932:
## 环境
- OpenClaw: 5.27
- memory-tencentdb: 0.3.6 (npm latest)
- Node.js: 24.14.0
- OS: Linux 6.18.18-trim (x64)

## 问题描述
L3 上下文压缩功能完全失效。每次 after_tool_call 调用后，日志显示：

[context-offload] after_tool_call patch check: NOT EFFECTIVE (status=missing_field). event.messages is undefined. L3 compression will be skipped this turn.

`after_tool_call` hook 中 `event.messages` 为 undefined，导致 L3 压缩每次被跳过。

## 已验证
- L1 离线存储：正常（pending≥4时正常触发 flush）
- L2 向量存储：正常（vectors.db 活跃）
- offload 服务（http://192.168.232.62:8090）：在线，返回 200
- OpenClaw 5.27 release notes 提到 channel/session identity 安全加固，可能影响 hook event 结构

## 复现频率
每次 tool call 都触发，无例外。

## 期望
after_tool_call 能拿到 event.messages，L3 压缩正常工作。

Maxwell-Code07:
感谢详细的报告！可能确实是OpenClaw在5.27版本的 hook 接口变更导致，我们会内部排查，感谢您清晰的日志说明！

honchow:
有试过执行patch脚本吗，scripts/openclaw-after-tool-call-messages.patch.sh
这个脚本就是用来解决after_tool_call hook 中 event.messages 为 undefined的问题。

YOMXXX:
Triage update: #184 is the canonical fix for this issue and has been approved.

Local verification for #184:
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/offload/hooks/after-tool-call.test.ts` (4/4 passed)
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm test` (71/71 passed)
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm build`
- `git diff --check`

Duplicate cleanup: closed #249 as superseded by #184.

Note: local Node is v22.14.0, below the package engine `>=22.16.0`; commands completed successfully with the expected engine warning.

PR #131, opened 2026-06-03
fix(openclaw-patch): clean no-op backups, skip 5.22 inline-spread, harden filters

rocke2020:
## Summary

Three improvements to `scripts/openclaw-after-tool-call-messages.patch.sh`, plus two prior local-only fixes that ride along (all three commits touch only that one file).

The motivating incident: after running the patch script against OpenClaw 2026.5.22, two `.pre-offload-patch.bak` files remained next to the live `dist/` files, byte-identical to their sources. This made it look like the script had modified those files and broken the gateway. Investigation showed the gateway breakage was unrelated (a wrapper-plist + secret-broker config issue), but the misleading backups were a real UX bug worth fixing.

## Platform support

The script is **verified on macOS** (and remains compatible with Linux). The portability commit `6067d6d` was made specifically to make the script run on the default macOS shell + utility set:

| macOS default | Behavior the script no longer relies on |
|---|---|
| `bash 3.2.57(1)-release` (arm64-apple-darwin25), `/bin/bash` | `mapfile -t` (bash 4+) — replaced with a portable `while IFS= read -r` loop |
| BSD `grep 2.6.0-FreeBSD` | `grep -oP` (GNU PCRE) for version parse — replaced with a `node -e` JSON read |
| BSD `grep` | `grep -qP` (PCRE) inside strategies — replaced with `grep -Eq` (POSIX ERE) and `perl` |
| BSD `sed -i` | `sed -i -E` inline edit (BSD `sed -i` requires a backup-suffix arg) — replaced with `perl -i -pe` |

My new changes (commit `547874f`) intentionally use only macOS-portable primitives: `find ... -name`, `cmp -s`, `grep -qE`, `grep -q`, and `perl -0777 -ne` — all available on a vanilla macOS install with the system shell.

Test environment for this PR:

```
ProductName:    macOS
ProductVersion: 26.5
BuildVersion:   25F71
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)
grep (BSD grep, GNU compatible) 2.6.0-FreeBSD
OpenClaw 2026.5.22
```

### 1. Tighten candidate filter (commit 547874f)

Require `runAfterToolCall` (the actual hook call site) **in addition to** the literal string `after_tool_call`. Excludes schema/test-contract files such as `plugin-sdk/agent-runtime-test-contracts.js` and `command-registration-*.js`, which only mention `"after_tool_call"` as event-name data.

Before: 5 candidates → 3 with backups created speculatively.
After:  3 candidates → no speculative backups on non-runtime files.

### 2. Detect OpenClaw 5.22+ inline-spread shape and skip cleanly (commit 547874f)

`native-hook-relay-CKQJJkrR.js` in 5.22 was refactored to pass the hook event as an inline object to `runAfterToolCall(...)`, with `durationMs` wrapped in a conditional spread:

```js
await hookRunner.runAfterToolCall({
    toolName: params.toolName,
    ...
    ...params.startedAt != null ? { durationMs: Date.now() - params.startedAt } : {}
}, { /* second arg */ });
```

Two structural reasons this site cannot be patched by the current script:

| Structural issue | Consequence |
|---|---|
| No `hookEvent` literal identifier | Strategies 1, 3, 4 cannot anchor |
| No `ctx` in function scope (only `params`) | The injection text `ctx.params.session?.messages` would `ReferenceError` at runtime |
| `params` has no `session` field either | A `params.session?.messages` rewrite would always be `undefined` |

Patching this site needs an **upstream OpenClaw change** to thread `session.messages` into the function signature. Until then, the new detection emits an explicit INFO line:

```
[INFO] native-hook-relay-CKQJJkrR.js — OpenClaw 5.22+ 内联展开形态（调用点无 ctx 作用域），跳过；该位置需 OpenClaw 上游修改才能注入 messages
```

This replaces the silent debug-level skip + leftover backup, so future readers see *why* and *what's needed*.

The main runtime path `selection-hR-AeOeU.js:2490` still has `ctx` in scope and still gets patched successfully — context-offload's L1/L1.5 pipeline keeps receiving `messages` from that site.

### 3. Global no-op backup cleanup pass at end of run (commit 547874f)

Walk `$DIST_DIR` and drop any `.pre-offload-patch.bak` byte-identical to its source. Catches:

- regex matched nothing → file unchanged but backup left behind
- strategy 4 restored from backup → backup now identical to file
- legacy backups from prior runs that the new candidate filter no longer considers

The script's effect on disk now matches its summary: `.bak exists` ⇔ `script actually modified that file`. The summary line surfaces a `清理空备份` counter so the cleanup is visible without `DEBUG=1`.

### 4. macOS / portability fixes (commit 6067d6d, pre-existing on fork)

Detailed in the **Platform support** table above. In short: removes every GNU-only / bash-4+ dependency the script used to have, so it runs on a vanilla macOS install. Also moves `backup_file` *inside* each strategy so backups are only created when a strategy actually attempts an edit, and adds `edit_attempted` tracking so "no strategy matched" becomes `SKIPPED++` instead of `FAILED++`.

### 5. Empty candidate list guard (commit de98071, pre-existing on fork)

Guards the per-file loop with `if [[ ${#CANDIDATE_FILES[@]} -gt 0 ]]` so an empty `dist/` produces a clean summary instead of iterating empty.

## Test plan

- [x] `bash -n scripts/openclaw-after-tool-call-messages.patch.sh` — syntax OK on macOS `bash 3.2.57`.
- [x] First run against live OpenClaw 2026.5.22 install with `DEBUG=1`:
  - 3 candidates (down from 5; test-contract + command-registration excluded by filter)
  - `selection-hR-AeOeU.js` → already-patched, skip
  - `native-hook-relay-CKQJJkrR.js` → 5.22 inline-spread skip with explicit INFO
  - `hook-runner-global-BkXXy1ub.js` → no `durationMs`, skip
  - 2 no-op backups cleaned up
  - Final summary: `成功: 0  跳过: 3  失败: 0  清理空备份: 2`
- [x] Second run for idempotence: same skip counts, `清理空备份: 0`, no `.pre-offload-patch.bak` left in `dist/`.
- [x] `selection-hR-AeOeU.js:2490` confirmed to still carry `messages: ctx.params.session?.messages` after the cleanup pass.
- [x] Live end-to-end with a gateway tool call: `[context-offload] [local-llm] L1 <<< 4506ms, output=317 chars` and `[context-offload] [local-llm] L1.5 <<< 2765ms, output=132 chars` — context-offload still fully operational.

### DIFF
--- a/scripts/openclaw-after-tool-call-messages.patch.sh
+++ b/scripts/openclaw-after-tool-call-messages.patch.sh
@@ -176,6 +176,7 @@ SKIPPED=0
 FAILED=0
 
 # ─── 对每个候选文件尝试多种策略 ──────────────────────────────────
+if [[ ${#CANDIDATE_FILES[@]} -gt 0 ]]; then
 for f in "${CANDIDATE_FILES[@]}"; do
     fname="$(basename "$f")"
     relpath="${f#$DIST_DIR/}"
@@ -300,6 +301,7 @@ for f in "${CANDIDATE_FILES[@]}"; do
         fi
     fi
 done
+fi
 
 # ─── 结果报告 ────────────────────────────────────────────────────
 echo ""
--- a/scripts/openclaw-after-tool-call-messages.patch.sh
+++ b/scripts/openclaw-after-tool-call-messages.patch.sh
@@ -159,10 +159,17 @@ backup_file() {
 }
 
 # ─── 查找所有候选文件 ────────────────────────────────────────────
-# 收集所有包含 after_tool_call 的 JS 文件（不限子目录深度）
+# 收集所有包含 after_tool_call 的 JS 文件（不限子目录深度）。
+# 再要求出现 runAfterToolCall 调用点：去除仅在 schema/test-contract 中
+# 以字面量形式出现 "after_tool_call" 的非运行时文件，避免后续策略对其
+# 留下空操作备份。
 CANDIDATE_FILES=()
 while IFS= read -r candidate_file; do
-    CANDIDATE_FILES+=("$candidate_file")
+    if grep -q 'runAfterToolCall' "$candidate_file" 2>/dev/null; then
+        CANDIDATE_FILES+=("$candidate_file")
+    else
+        debug "${candidate_file#$DIST_DIR/} — 无 runAfterToolCall 调用点（仅含字面量 \"after_tool_call\"），不作为候选"
+    fi
 done < <(grep -rl 'after_tool_call' "$DIST_DIR" --include='*.js' 2>/dev/null || true)
 
 if [[ ${#CANDIDATE_FILES[@]} -eq 0 ]]; then
@@ -174,6 +181,7 @@ info "找到 ${#CANDIDATE_FILES[@]} 个候选文件"
 PATCHED=0
 SKIPPED=0
 FAILED=0
+CLEANED=0
 
 # ─── 对每个候选文件尝试多种策略 ──────────────────────────────────
 if [[ ${#CANDIDATE_FILES[@]} -gt 0 ]]; then
@@ -198,6 +206,20 @@ for f in "${CANDIDATE_FILES[@]}"; do
     # 确认 durationMs 附近有 after_tool_call 上下文（避免误匹配 before_compaction 等）
     if ! perl -0777 -ne 'exit(0) if /after_tool_call[\s\S]{0,2000}durationMs/; exit(1)' "$f" 2>/dev/null; then
         debug "$relpath — durationMs 不在 after_tool_call 上下文中，跳过"
+        ((SKIPPED++)) || true
+        continue
+    fi
+
+    # OpenClaw 5.22+ 内联展开形态检测：durationMs 仅出现在条件展开
+    # (...cond ? { durationMs: ... } : {}) 中，调用点没有 hookEvent 字面量、
+    # 函数作用域也没有 ctx（只有 params）。当前注入文本 ctx.params.session?.messages
+    # 在该作用域下会触发 ReferenceError，且 params 上也没有 session 字段，
+    # patch script 无法在此位置完成注入 —— 需要 OpenClaw 上游把 session.messages
+    # 透传进 runAgentHarnessAfterToolCallHook 的函数签名。提前显式跳过，
+    # 避免后续策略空跑并留下误导性的 .pre-offload-patch.bak。
+    if perl -0777 -ne 'exit(0) if /\.\.\.\s*\w+\s*\.\s*startedAt\s*!=\s*null\s*\?\s*\{\s*durationMs\s*:/; exit(1)' "$f" 2>/dev/null \
+       && ! grep -qE '(^|\W)(hookEvent|hook_event)(\W|$)' "$f" 2>/dev/null; then
+        info "$relpath — OpenClaw 5.22+ 内联展开形态（调用点无 ctx 作用域），跳过；该位置需 OpenClaw 上游修改才能注入 messages"
         ((SKIPPED++)) || true
         continue
     fi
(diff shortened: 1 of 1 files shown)


rocke2020:
## macOS support — details

Calling this out separately because it was the original motivation for commit `6067d6d` and applies to anyone running OpenClaw on a Mac with the system shell. Tested on `macOS 26.5 (25F71) arm64` with `/bin/bash` (`3.2.57(1)-release`) and BSD `grep` (`2.6.0-FreeBSD`).

### Why the original script didn't run on macOS

The pre-`6067d6d` script used four idioms that fail on a vanilla macOS install:

| # | Idiom in original script | Why it fails on macOS default | Concrete failure |
|---|---|---|---|
| 1 | `mapfile -t CANDIDATE_FILES < <(...)` | `mapfile` is bash 4+; macOS ships `/bin/bash` 3.2.57 | `mapfile: command not found` → empty `CANDIDATE_FILES`, script reports "0 candidates" |
| 2 | `VERSION=$(grep -oP '"version"\s*:\s*"\K[^"]+' package.json)` | BSD `grep` has no `-P` (PCRE), and no `\K` look-behind reset | `grep: invalid option -- P` → `VERSION=unknown` |
| 3 | `grep -qP '^\s+durationMs\s*$'` (strategy 2 pre-check) | Same BSD `grep` no-PCRE issue | Strategy 2 never fires on dispatch-*.js |
| 4 | `sed -i -E 's/.../.../' file` (strategy 2 edit) | BSD `sed -i` requires an explicit empty backup-suffix arg (`sed -i ''`); GNU accepts bare `-i` | `sed: 1: "...": invalid command code` → script aborts mid-strategy |

### What `6067d6d` swapped in (still in this PR)

| # | Replaced with | Why portable |
|---|---|---|
| 1 | `while IFS= read -r line; do arr+=("$line"); done < <(...)` | POSIX-shell idiom, works on bash 3.2+ |
| 2 | `node -e "const fs=require('fs'); const pkg=JSON.parse(fs.readFileSync(process.argv[1], 'utf8')); console.log(pkg.version || 'unknown'); " "$pkg"` | Node is already a hard dependency of OpenClaw; no awk/jq added |
| 3 | `grep -Eq '^[[:space:]]+durationMs[[:space:]]*$'` | POSIX ERE + POSIX character classes (`[[:space:]]` works on both BSD and GNU) |
| 4 | `perl -i -pe 's/.../.../' file` | Perl ships with macOS (`/usr/bin/perl`); `-i` works the same on BSD and GNU |

### What my new commit `547874f` adds (also macOS-safe)

Only these primitives are introduced, all available on a vanilla macOS install:

```bash
grep -q 'runAfterToolCall' "$f"         # POSIX, BSD-safe
grep -qE '(^|\W)(hookEvent|hook_event)(\W|$)' "$f"   # POSIX ERE
perl -0777 -ne '...'                    # /usr/bin/perl, BSD-safe
find "$DIST_DIR" -name '*.pre-offload-patch.bak'   # POSIX find
cmp -s "$src" "$bak"                    # POSIX
${bak%.pre-offload-patch.bak}           # POSIX parameter expansion
```

No `mapfile`, no `grep -P`, no `sed -i` without backup-suffix, no `readarray`, no GNU-only flags. The script as a whole now runs end-to-end on macOS `bash 3.2.57` without any Homebrew packages.

### Verified runtime evidence on macOS 26.5

```
$ bash --version | head -1
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)

$ grep --version | head -1
grep (BSD grep, GNU compatible) 2.6.0-FreeBSD

$ bash scripts/openclaw-after-tool-call-messages.patch.sh
[INFO]  OpenClaw 目录: /opt/homebrew/lib/node_modules/openclaw
[INFO]  检测到 OpenClaw 版本: 2026.5.22
[INFO]  找到 3 个候选文件
[INFO]  native-hook-relay-CKQJJkrR.js — OpenClaw 5.22+ 内联展开形态（调用点无 ctx 作用域），跳过；该位置需 OpenClaw 上游修改才能注入 messages
[WARN]  selection-hR-AeOeU.js — 已经 patch 过，跳过

═══════════════════════════════════════════════════════
  Patch 完成  (OpenClaw 2026.5.22)
═══════════════════════════════════════════════════════

  成功: 0  跳过: 3  失败: 0  清理空备份: 2
```

No `command not found`, no `invalid option`, no `invalid command code`. macOS users get the same behavior as Linux users.

Maxwell-Code07:
Thanks for the contribution! We've received your PR improving the OpenClaw patch script and will review it internally. We'll get back to you once we have results. Appreciate the detailed troubleshooting!

rocke2020:
Cross-linking #118 as related context.

Same symptom (`event.messages undefined` in `after_tool_call`) but on OpenClaw 5.27, which is a stricter superset of the 5.22+ inline-spread shape this PR detects. To be clear, this PR does **not** close #118 — the new skip just makes the failure mode honest (explicit INFO + no misleading `.pre-offload-patch.bak`) instead of silently dropping `messages` and leaving a backup that looks like the script broke something.

A real fix for #118 needs an OpenClaw upstream change. The inline-spread call site in `native-hook-relay-*.js` has no `ctx` in scope and `params` has no `session` field, so no in-place regex injection from this script can work there — `runAfterToolCall` would have to take `session.messages` in its signature.

The main runtime path on `selection-hR-AeOeU.js:2490` still has `ctx` and gets patched fine, which is why context-offload L1/L1.5 keep firing end-to-end in my run.

PR #134, opened 2026-06-03
fix(recall): scope L1 search by session_key for agent/user isolation

### LINKED ISSUE #111: searchMemories has no agent/user-level isolation — multi-agent memory cross-contamination
## Environment
- Plugin: `@tencentdb-agent-memory/memory-tencentdb` v0.3.6
- OpenClaw: 2026.5.19

## Why this matters

OpenClaw is a multi-agent system. Different agents serve completely different domains:

| Agent | Role | Memory domain |
|-------|------|---------------|
| `main` | daily chat, system admin, web search | personal preferences, configs, weather |
| `coder` | embedded driver development | POLSDK-9454, AR9481, HAL, I2C register configs |
| `omoc_coder` | omoc plugin development | TypeScript, Node.js, plugin architecture |

Without agent-level isolation, when a user asks `main` "what should I deal with today?", the recall injects coder memories about driver register configs and I2C timing bugs. Conversely, asking `coder` to analyze a driver bug pulls in daily chat trivia. Memory pollution wastes context window tokens and degrades recall precision.

The same issue affects multi-user isolation — `session_key` contains the user ID (e.g. `ou_8e408d15a585f2a33abcee02f911fdf5`) but is never used in recall filtering.

## Root cause

**`searchMemories` performs full-table scans with no scoping.**

### 1. `sessionKey` is lost in the call chain
```
handleBeforeRecall(userText, sessionKey)        // ✅ received
  → performAutoRecall({ sessionKey, ... })       // ✅ passed in
    → performAutoRecallInner(params)             // params.sessionKey exists
      → searchMemories(userText, ...)            // ❌ sessionKey not passed
```

### 2. FTS search — no agent filter
```sql
SELECT record_id, content_original AS content, type, priority, scene_name,
       session_key, session_id, ..., bm25(l1_fts) AS rank
FROM l1_fts
WHERE l1_fts MATCH ?
ORDER BY rank ASC
-- ❌ no session_key / agent filter
```

Note: `l1_fts.session_key` is `UNINDEXED`, so filtering must be done client-side after the query, not in the SQL WHERE.

### 3. Vector search — no agent filter
`searchL1Vector` retrieves `session_key` from `stmtGetMeta` but only puts it in the return object — no filtering.

### 4. Storage layer already has isolation capability
```sql
CREATE TABLE l1_records (
  record_id TEXT PRIMARY KEY,
  session_key TEXT DEFAULT ,  -- "agent:main:feishu:direct:ou_xxx" format
  ...
);
CREATE INDEX idx_l1_session_key ON l1_records(session_key);
```

`sesssion_key` encodes both agent ID and user ID. Records are written correctly per-session. The problem is only on the recall side.

## Impact

| Scenario | Effect |
|----------|--------|
| main agent daily chat | injects coder driver dev memories (POLSDK-9454, AR9481 registers, etc.) |
| coder agent coding | injects main daily trivia and config memories |
| multiple project agents | cross-contamination across project domains |
| multi-user (future) | user A sees user B memories |

Currently masked by another bug (L1 extraction pipeline failure → #110) since the L1 records table is empty. Once that is fixed, this will surface immediately.

YOMXXX:
## Summary

Fixes #111 — `searchMemories` was performing global L1 searches across every captured `session_key`, so in a multi-agent or multi-user setup the auto-recall hook pulled the wrong agent's / wrong user's memories into the prompt. The OP filed it as a serious cross-contamination bug with a clean root-cause analysis; this PR implements their proposed fix.

### Root cause (reproduced from the issue)

`handleBeforeRecall(userText, sessionKey)` → `performAutoRecall({...sessionKey})` → `performAutoRecallInner(params)` already carried `sessionKey`, but the next call `searchMemories(userText, …)` dropped it. From that point on:

- `searchByKeyword` ran `vectorStore.searchL1Fts(ftsQuery, maxResults * 2)` with no scoping.
- `searchByEmbedding` ran `vectorStore.searchL1Vector(queryEmbedding, maxResults * 2)` with no scoping.
- `searchHybrid` did the same in parallel and RRF-merged.
- The tool entry `executeMemorySearch` had no `sessionKey` parameter at all, while the sibling `executeConversationSearch` already had one.

`l1_fts.session_key` is `UNINDEXED` so server-side `WHERE` is not an option — client-side filter on the raw `L1SearchResult` / `L1FtsResult` (both already carry `session_key`) is the only safe path.

### Change

- **`src/core/hooks/auto-recall.ts`** — thread `sessionKey` from `performAutoRecallInner` into `searchMemories` and the four strategy branches (`searchByKeyword`, `searchByEmbedding`, `searchHybrid`, native-hybrid). Add `applySessionFilter` helper that drops raw FTS / vector results whose `session_key !== sessionKey`. Each strategy now over-retrieves (×4 single, ×5 hybrid RRF) when scoped so the post-filter pool can still fill `maxResults`.
- **`src/core/tools/memory-search.ts`** — add `sessionKey?` to `executeMemorySearch` params; add `session_key` to `MemorySearchResultItem` so RRF-merged items carry the key. Apply the session filter alongside the existing `type` / `scene` filters; over-retrieve when scoped.
- **`src/core/types.ts`** — add `sessionKey?: string` to `MemorySearchParams`.
- **`src/core/tdai-core.ts`** — forward `params.sessionKey` into `executeMemorySearch`.
- **`index.ts`** — expose `session_key` as an optional parameter on `tdai_memory_search`, mirroring `tdai_conversation_search`.

### Semantics

- **Auto-recall hook**: strictly scoped to the active `sessionKey`. The hook is already short-circuited upstream when `sessionKey` is empty (see `resolveSessionKey` in `index.ts`), so non-empty + strict-match is the right default.
- **`tdai_memory_search` tool**: `session_key` is **optional**. Agents that want cross-session recall (\"summarize everything you know about the user\") can still omit it; agents that want isolation pass it through. Same shape as `tdai_conversation_search` to keep the tool surface consistent.

### What's intentionally NOT in this PR

- No SQL schema change (no migration). `l1_fts.session_key` stays `UNINDEXED`; client-side filter is fine for the recall budget (`maxResults` is single digits) and the over-retrieve multiplier compensates for the post-filter drop. If profiling later shows the multiplier is hurting BM25 quality, we can add an indexed mirror or a composite FTS index in a follow-up.
- No backfill / migration for existing records — their `session_key` is already correctly stamped at write time.
- No automated isolation test — the repo currently ships no tracked `.test.ts` files (verified via `git ls-files`), so a one-off test for this PR wouldn't run anywhere. Verification is by hand below.

## Test plan

- [x] `npm run build` passes locally (tsdown clean).
- [ ] Two sessions on the same backend (e.g. `main` agent and `coder` agent under the same user, or two users under the same host):
  - capture distinct L1 memories from each session
  - trigger auto-recall from session A — confirm only A's records appear in `<relevant-memories>` (and the `[session-filter:*]` debug log shows the drop count)
  - same for session B
- [ ] `tdai_memory_search` called without `session_key` still returns the global pool (no regression for the cross-session use case).
- [ ] `tdai_memory_search` called with `session_key=<current>` filters correctly.
- [ ] Native-hybrid path (TCVDB backend with `nativeHybridSearch=true`) still works — verified that we now over-retrieve and post-filter rather than re-issue the request.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -372,6 +372,13 @@ export default function register(api: OpenClawPluginApi) {
             type: "string",
             description: "Optional filter by scene name",
           },
+          session_key: {
+            type: "string",
+            description:
+              "Optional: restrict results to memories captured under this session_key. " +
+              "Use this to keep agent/user memories isolated — pass the current session's " +
+              "key to avoid recalling other agents' or users' records.",
+          },
         },
         required: ["query"],
       },
@@ -381,15 +388,16 @@ export default function register(api: OpenClawPluginApi) {
         const limit = Math.min(Math.max(Number(params.limit) || 5, 1), 20);
         const typeFilter = typeof params.type === "string" ? params.type : undefined;
         const sceneFilter = typeof params.scene === "string" ? params.scene : undefined;
+        const sessionKeyFilter = typeof params.session_key === "string" ? params.session_key : undefined;
 
         api.logger.debug?.(
           `${TAG} [tool] tdai_memory_search called: ` +
           `query="${query.length > 80 ? query.slice(0, 80) + "…" : query}", ` +
-          `limit=${limit}, type=${typeFilter ?? "(all)"}, scene=${sceneFilter ?? "(all)"}`,
+          `limit=${limit}, type=${typeFilter ?? "(all)"}, scene=${sceneFilter ?? "(all)"}, session_key=${sessionKeyFilter ?? "(all)"}`,
         );
 
         try {
-          const result = await core.searchMemories({ query, limit, type: typeFilter, scene: sceneFilter });
+          const result = await core.searchMemories({ query, limit, type: typeFilter, scene: sceneFilter, sessionKey: sessionKeyFilter });
 
           const elapsedMs = Date.now() - startMs;
           api.logger.debug?.(
--- a/src/core/hooks/auto-recall.ts
+++ b/src/core/hooks/auto-recall.ts
@@ -121,7 +121,16 @@ async function performAutoRecallInner(params: {
     logger?.debug?.(`${TAG} User text empty/undefined, skipping memory search (persona/scene still injected)`);
   } else {
     effectiveStrategy = cfg.recall.strategy ?? "hybrid";
-    const searchResult = await searchMemories(userText, pluginDataDir, cfg, logger, effectiveStrategy as "keyword" | "embedding" | "hybrid", vectorStore, embeddingService);
+    const searchResult = await searchMemories(
+      userText,
+      pluginDataDir,
+      cfg,
+      logger,
+      effectiveStrategy as "keyword" | "embedding" | "hybrid",
+      vectorStore,
+      embeddingService,
+      params.sessionKey,
+    );
     memoryLines = searchResult.lines;
     searchTiming = searchResult.timing;
     memoryLines = applyRecallBudget(memoryLines, cfg.recall, logger);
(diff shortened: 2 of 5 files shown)


Maxwell-Code07:
Thanks for the contribution! We've received your PR fixing the cross-session memory pollution bug and will review it internally. We'll get back to you once we have results. Appreciate your help in improving multi-agent isolation!

YOMXXX:
Closing this as superseded by #354.

#354 fixes the same #111 session-isolation bug, is based on the same scoped-search approach, and adds focused regression coverage for the FTS-only, unscoped global search, and hybrid FTS/vector paths. I verified #354 locally with:

- `COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/core/tools/memory-search.test.ts`
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm test`
- `git diff --check`

Keeping #354 as the canonical PR should avoid maintaining two near-identical client-side session filtering implementations.

Issue #69, opened 2026-05-21
[good first issue]🎯 [Feature] 希望有一篇如何集成到hermes agent的文档


ahkimkoo:
### Is this feature related to a problem? | 该功能需求是否与某个问题相关？请描述

如何集成到已有的hermes agent， 一般情况下都是先装了hermes再装插件，不可能是为了使用插件而安装hermes，目前的文档写的是把hermes和插件粗暴的揉在一起的docker镜像。

### Describe the solution you'd like | 描述你期望的解决方案

希望有相关文档。谢谢。

Maxwell-Code07:
您好，非常感谢您提出这个问题，目前确实是有这个问题，当时主要是考虑到Hermes场景安装比较复杂，因此直接提供了Docker镜像。
后面我们会尽快补齐在已安装的Hermes上安装该插件的文档，感谢关注！

nors1993:
Hermes 不是自带memory吗？

csjlengxiang:
跟 hermes 聊，让hermes自己出方案。。我是弄的 hermes在本地，docker把 tdai-memory-gateway 给弄起来了

➜  ~ docker ps
CONTAINER ID   IMAGE                                   COMMAND                   CREATED        STATUS                  PORTS                                         NAMES
4bed5cc8093a   tdai-memory-gateway                     "docker-entrypoint.s…"   18 hours ago   Up 18 hours (healthy)   127.0.0.1:8420->8420/tcp                      tdai-memory-gateway
d12f17ddee98   ghcr.io/jo-inc/camofox-browser:latest   "docker-entrypoint.s…"   19 hours ago   Up 19 hours             0.0.0.0:9377->9377/tcp, [::]:9377->9377/tcp   camofox-browser

nextwa:
希望可以单独成为已安装hermes的memory.provider，走Hermes memory.provider原生路由，接管on_pre_compress 压缩前抢救
on_memory_write 镜像 MEMORY.md / USER.md
完整 Context Engine 接管
完整工具日志黑匣子归档

Maxwell-Code07:
We have selected this Issue as a 🎯 **Good First Issue** — perfect for new contributors to get started!

## Task Details

| Field | Content |
|---|---|
| Type | Feature |
| Difficulty | Low |
| Estimated Effort | 4 days |
[...]

PR #137, opened 2026-06-03
[good first issue] feat(embedding): support ZeroEntropy native embed API

### LINKED ISSUE #68: [good first issue]🎯 [Feature] Support ZeroEntropy native embedding API
### Is this feature related to a problem? | 该功能需求是否与某个问题相关？请描述

# Support ZeroEntropy native embedding API in `memory-tencentdb`

## Summary

Please add first-class support for ZeroEntropy as an embedding provider in `@tencentdb-agent-memory/memory-tencentdb`.

Currently, `memory-tencentdb` treats non-local embedding providers as OpenAI-compatible and calls:

ts
${this.baseUrl}/embeddings

This works for OpenAI-compatible providers, but ZeroEntropy’s working embedding endpoint is native/non-OpenAI-shaped:

POST https://api.zeroentropy.dev/v1/models/embed

with model:
zembed-1

and request body shape:

json
{
  "input": ["text to embed"],
  "input_type": "query",
  "model": "zembed-1"
}

The response shape is also different from OpenAI:

json
{
  "results": [
    {
      "embedding": [...]
    }
  ]
}

rather than:

json
{
  "data": [
    {
      "index": 0,
      "embedding": [...]
    }
  ]
}

## Environment

- Package: `@tencentdb-agent-memory/memory-tencentdb`
- Observed version: `0.3.5`
- Runtime: Hermes sidecar gateway
- Gateway health reports:
  - `vectorStore: true`
  - `embeddingService: true`
- Embedding config:
  - `provider: openai`
  - `baseUrl: https://api.zeroentropy.dev`
  - `model: zembed-1`
  - `dimensions: 2560`

## Current behavior

With ZeroEntropy configured, `memory-tencentdb` calls:

POST https://api.zeroentropy.dev/embeddings

That returns:

HTTP 404

If `/v1` is included in the base URL, the OpenAI-style endpoint still fails:

POST https://api.zeroentropy.dev/v1/embeddings
HTTP 404 {"detail":"Not Found"}

As a result, the embedding service object exists, but no vectors are persisted.

Example local DB state after multiple captured conversations:

l0_conversations: 123
l1_records: 3
l0_vec_rowids: 0
l1_vec_rowids: 0

Memory search falls back to FTS for exact keyword matches, while semantic/vector search returns no results.

## Expected behavior

`memory-tencentdb` should support ZeroEntropy embeddings, either via:

1. A dedicated provider, e.g.

yaml
embedding:
  enabled: true
  provider: zeroentropy
  baseUrl: https://api.zeroentropy.dev
  model: zembed-1
  dimensions: 2560

or

2. A configurable embedding endpoint/response adapter that can support non-OpenAI-compatible APIs.

For ZeroEntropy, the provider should:

- call:

POST {baseUrl}/v1/models/embed

- send:

json
{
  "input": ["..."],
  "input_type": "query",
  "model": "zembed-1"
}

- parse:

results[].embedding

- return normalized `Float32Array` embeddings as the existing embedding service does.

## Validation

Direct ZeroEntropy native call succeeds:

POST https://api.zeroentropy.dev/v1/models/embed
model: zembed-1

Result:

HTTP 200
embedding dimensions: 2560

But the current OpenAI-compatible path used by `memory-tencentdb` fails:

POST https://api.zeroentropy.dev/embeddings
HTTP 404

## Suggested implementation

Add a `ZeroEntropyEmbeddingService` or branch inside the existing remote embedding service:

ts
if (providerName === "zeroentropy") {
  const fetchUrl = ${baseUrl.replace(/\/+$/, "")}/v1/models/embed;
  const body = {
    input: texts,
    input_type: "query",
    model: this.model,
  };

  const json = await resp.json();
  if (!json.results || !Array.isArray(json.results)) {
    throw new Error("ZeroEntropy embedding API returned unexpected format: missing 'results' array");
  }

  return json.results.map((r) => sanitizeAndNormalize(r.embedding));
}

Optionally allow `input_type` to be configured (`query` vs `document`) if users need both retrieval modes.

## Why this matters

ZeroEntropy provides 2560-dimensional embeddings through `zembed-1`, and works well as a hosted embedding provider. Supporting it directly would let Hermes/TencentDB users use ZeroEntropy for semantic memory search without needing an OpenAI-compatible proxy.

_No response_

### Additional context | 补充说明

_No response_

YOMXXX:
## Summary

Fixes #68 — adds first-class support for ZeroEntropy as an embedding provider. Today, any non-`local` provider is routed through the OpenAI-compatible path (`POST {baseUrl}/embeddings` with `{input, model, dimensions?}` and a `{data: [{index, embedding}]}` response). ZeroEntropy's API is close but not the same shape, so the integration fails on the wire.

### Wire differences

| Aspect | OpenAI-compatible | ZeroEntropy |
|---|---|---|
| Endpoint | `POST {baseUrl}/embeddings` | `POST {baseUrl}/models/embed` |
| Required body field | — | `input_type: \"query\" \| \"document\"` |
| `dimensions` in body | optional | **not accepted** (fixed-dim model) |
| Response envelope | `{data: [{index, embedding}]}` | `{results: [{embedding}]}` |
| Result ordering | sort by `index` | preserved by array position |

### Change

Branched on `provider === \"zeroentropy\"` inside `OpenAIEmbeddingService._callApi` rather than introducing a new service class — the timeout, retry, batching, char-cap truncation, sanitize+normalize, and `qclaw` proxy logic are all identical, only the wire format diverges.

- `src/core/store/embedding.ts`:
  - New `ZeroEntropyEmbeddingResponse` type.
  - In `_callApi`: detect `providerName === \"zeroentropy\"`; switch path to `/models/embed`, set `input_type: \"query\"`, drop the `dimensions` field, parse `results[]` preserving array order. The qclaw `Remote-URL` header uses the same path so users can still front ZeroEntropy through the proxy.
- `openclaw.plugin.json`: `embedding.provider` description now lists `zeroentropy` alongside `openai` and `deepseek`.

### Why `input_type: \"query\"` as a default

The recall hot path is the only one that calls `embed()` for a Float32Array; capture-side batches eventually feed the same vector store. ZeroEntropy's `query` and `document` variants are symmetric enough that one type works for both directions, and bringing `input_type` up to the public API would require thread changes that aren't needed to make the bug-report use case work. If a future user reports recall quality regressions on cross-modal corpora, we can expose it via config.

## Test plan

- [x] `npm run build` (tsdown) clean.
- [ ] Configure `embedding.provider: \"zeroentropy\"`, `baseUrl: \"https://api.zeroentropy.dev/v1\"`, `model: \"zembed-1\"`, valid `apiKey`. Recall a remembered fact → confirm the gateway logs an embedding call against `/v1/models/embed` (not `/embeddings`) and the L1 vector hit comes back.
- [ ] Same config with an invalid model → confirm the 4xx is surfaced through the existing `EmbeddingApiError` path and the request is **not** retried.
- [ ] Existing OpenAI/deepseek/qclaw setups untouched — the branch only activates when `provider === \"zeroentropy\"`.

### DIFF
--- a/src/core/store/embedding.ts
+++ b/src/core/store/embedding.ts
@@ -399,6 +399,17 @@ interface OpenAIEmbeddingResponse {
   };
 }
 
+/**
+ * ZeroEntropy's `/v1/models/embed` returns input order via `results[i]`
+ * (no `index` field) and omits the OpenAI `data` envelope. See:
+ *   https://docs.zeroentropy.dev/api-reference/models/embed
+ */
+interface ZeroEntropyEmbeddingResponse {
+  results: Array<{
+    embedding: number[];
+  }>;
+}
+
 export class OpenAIEmbeddingService implements EmbeddingService {
   private readonly baseUrl: string;
   private readonly apiKey: string;
@@ -494,23 +505,43 @@ export class OpenAIEmbeddingService implements EmbeddingService {
   }
 
   private async _callApi(texts: string[], timeoutOverride?: number): Promise<Float32Array[]> {
+    // ZeroEntropy speaks a slightly different shape: native endpoint
+    // `/v1/models/embed`, request body carries `input_type` (must be
+    // "query" or "document"), response is `{results: [{embedding}]}`
+    // instead of OpenAI's `{data: [{index, embedding}]}`. The rest of
+    // the call surface (timeout, retry, batching) is identical, so we
+    // branch only on URL + body + parse rather than introducing a new
+    // service class. See issue #68.
+    const isZeroEntropy = this.providerName === "zeroentropy";
+
     const body: Record<string, unknown> = {
       input: texts,
       model: this.model,
     };
-    if (this.sendDimensions) {
+    if (isZeroEntropy) {
+      // ZeroEntropy rejects requests without input_type. Default to "query"
+      // because the recall hot path is the only one that calls embed() with
+      // a Float32Array return — capture-side batches eventually feed the
+      // same vector store, and ZeroEntropy's symmetry between "query" and
+      // "document" makes a single type safe.
+      body.input_type = "query";
+    } else if (this.sendDimensions) {
+      // ZeroEntropy's zembed-1 has a fixed dimension and rejects an
+      // explicit `dimensions` parameter — only set it on OpenAI-style
+      // providers (and only when the caller opted in).
       body.dimensions = this.dims;
     }
 
     // Determine fetch URL and headers based on proxy mode
     const useProxy = this.providerName === "qclaw" && !!this.proxyUrl;
-    const fetchUrl = useProxy ? this.proxyUrl! : `${this.baseUrl}/embeddings`;
+    const embedPath = isZeroEntropy ? "/models/embed" : "/embeddings";
+    const fetchUrl = useProxy ? this.proxyUrl! : `${this.baseUrl}${embedPath}`;
     const headers: Record<string, string> = {
       "Content-Type": "application/json",
       Authorization: `Bearer ${this.apiKey}`,
     };
     if (useProxy) {
-      headers["Remote-URL"] = `${this.baseUrl}/embeddings`;
+      headers["Remote-URL"] = `${this.baseUrl}${embedPath}`;
       this.logger?.debug?.(
         `${TAG} [qclaw-proxy] Forwarding embedding request via proxy: ${fetchUrl}, Remote-URL: ${headers["Remote-URL"]}`,
       );
@@ -545,14 +576,26 @@ export class OpenAIEmbeddingService implements EmbeddingService {
             continue;
           }
 
-          const json = (await resp.json()) as OpenAIEmbeddingResponse;
+          const json = await resp.json() as
+            | OpenAIEmbeddingResponse
+            | ZeroEntropyEmbeddingResponse;
 
-          if (!json.data || !Array.isArray(json.data)) {
+          if (isZeroEntropy) {
+            const zeJson = json as ZeroEntropyEmbeddingResponse;
+            if (!zeJson.results || !Array.isArray(zeJson.results)) {
+              throw new Error("ZeroEntropy embedding API returned unexpected format: missing 'results' array");
+            }
+            // ZeroEntropy preserves input order via array position (no `index`).
+            return zeJson.results.map((r) => sanitizeAndNormalize(r.embedding));
+          }
+
+          const oaJson = json as OpenAIEmbeddingResponse;
+          if (!oaJson.data || !Array.isArray(oaJson.data)) {
             throw new Error("Embedding API returned unexpected format: missing 'data' array");
           }
 
           // Sort by index to ensure correct order, then sanitize+normalize for consistency with local provider
-          const sorted = [...json.data].sort((a, b) => a.index - b.index);
+          const sorted = [...oaJson.data].sort((a, b) => a.index - b.index);
           return sorted.map((d) => sanitizeAndNormalize(d.embedding));
         } finally {
           clearTimeout(timeoutId);
--- a/src/core/store/embedding.ts
+++ b/src/core/store/embedding.ts
@@ -636,13 +636,16 @@ export class OpenAIEmbeddingService implements EmbeddingService {
  * ZeroEntropy native embedding adapter.
  *
  * Reuses {@link OpenAIEmbeddingConfig} for the wire-config shape (baseUrl /
- * apiKey / model / dimensions are identical), but the wire format diverges
- * in three places, so we keep this provider on its own class instead of
- * branching {@link OpenAIEmbeddingService}:
+ * apiKey / model / dimensions / sendDimensions are identical), but the wire
+ * format diverges in three places, so we keep this provider on its own class
+ * instead of branching {@link OpenAIEmbeddingService}:
  *
  * 1. Endpoint is `${baseUrl}/models/embed` (not `/embeddings`).
- * 2. Request body requires `input_type` (`"query"` or `"document"`) and
- *    rejects an explicit `dimensions` field (zembed-1 is fixed-dim).
+ * 2. Request body requires `input_type` (`"query"` or `"document"`).
+ *    `dimensions` is optional — for `zembed-1` the accepted values are the
+ *    Matryoshka set [2560, 1280, 640, 320, 160, 80, 40]; any other value is
+ *    rejected by the server. The config's `sendDimensions` flag (default
+ *    true) controls whether it is forwarded, matching the OpenAI path.
  * 3. Response envelope is `{ results: [{ embedding }] }` and preserves
  *    input order via array position rather than an `index` field.
  *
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Thanks for participating in our Good First Issue program! 🎯

We're keeping these 2 PRs for review:

The PR for #69 — documentation for non-Docker Hermes integration
The PR for #68 — support for ZeroEntropy native embedding API
We'll review them internally and get back to you soon. Really appreciate your contributions and looking forward to merging these! 🚀

src/core/store/embedding.ts
```diff
@@ -494,23 +505,43 @@ export class OpenAIEmbeddingService implements EmbeddingService {
...
 
   private async _callApi(texts: string[], timeoutOverride?: number): Promise<Float32Array[]> {
+    // ZeroEntropy speaks a slightly different shape: native endpoint
+    // `/v1/models/embed`, request body carries `input_type` (must be
+    // "query" or "document"), response is `{results: [{embedding}]}`
+    // instead of OpenAI's `{data: [{index, embedding}]}`. The rest of
+    // the call surface (timeout, retry, batching) is identical, so we
+    // branch only on URL + body + parse rather than introducing a new
+    // service class. See issue #68.
+    const isZeroEntropy = this.providerName === "zeroentropy";
```
Xuruida: Suggestion: since ZeroEntropy diverges on wire format (/models/embed, input_type, results[]vs data[]), I’d prefer we extract a small EmbeddingService and introduce a dedicated ZeroEntropyEmbeddingService in this PR, rather than adding provider-specific if/else into _callApi.

This keeps OpenAIEmbeddingService focused on OpenAI-compatible behavior and avoids accumulating protocol differences over time.

The refactoring is small and low-risk—appreciate it if we can address it here 🚀

Xuruida (changes requested):

YOMXXX:
Thanks for the review @Xuruida — agreed, branching a single service on `providerName` was the wrong call. Pushed [`401f813`](https://github.com/YOMXXX/TencentDB-Agent-Memory/commit/401f813) which addresses it:

**Split into two parallel `EmbeddingService` implementations sharing only the transport layer.**

- New module-level helpers (provider-agnostic):
  - `truncateEmbeddingInputs(texts, maxInputChars, logger)` — was previously a private method on the OpenAI class.
  - `postEmbeddingRequest({fetchUrl, headers, body, timeoutMs})` — owns fetch + `AbortController` timeout + exponential-backoff retry + the `EmbeddingApiError` non-retry rule for 4xx (except 429). Returns the parsed JSON; callers own the response shape.
- `OpenAIEmbeddingService._callApi` is back to OpenAI-only — no more `isZeroEntropy` reads, qclaw proxy branch unchanged, the `dimensions`-in-body path is exactly what it was before this PR was opened.
- New `ZeroEntropyEmbeddingService implements EmbeddingService` encodes the ZeroEntropy wire format independently: posts to `${baseUrl}/models/embed`, sends `input_type: "query"`, omits `dimensions` (`zembed-1` is fixed-dim), parses `{results: [{embedding}]}` preserving array order. The class doesn't carry the OpenAI-only knobs (`sendDimensions`, `proxyUrl`, qclaw), so its surface stays small.
- `createEmbeddingService` factory: new `provider === "zeroentropy"` branch placed before the OpenAI-compatible catch-all so it wins the dispatch.

**Behaviour preserved exactly:**
- All existing OpenAI / DeepSeek / Azure / qclaw deployments hit the same wire calls as before — diff isolated to the OpenAI `_callApi` body now delegating to `postEmbeddingRequest`.
- ZeroEntropy wire format (URL + body + response) matches the original PR's branched implementation byte for byte.
- Timeout, retry, error mapping, batch splitting, char-cap truncation, sanitize+normalize are shared through the module helpers so the two services can't drift by accident.

One judgment call I'd appreciate your read on: I left `ZeroEntropyEmbeddingService`'s constructor on `OpenAIEmbeddingConfig` because the wire-config shape (baseUrl/apiKey/model/dimensions/maxInputChars/timeoutMs) is genuinely identical. Adding a `ZeroEntropyEmbeddingConfig` alias would just be a re-export. Happy to introduce one if you'd rather keep the config types one-to-one with the service classes.

CI was already green pre-refactor; pushing this should keep it that way (build is clean locally). Let me know if you'd like anything tightened further.

src/core/store/embedding.ts
```diff
@@ -516,63 +611,150 @@ export class OpenAIEmbeddingService implements EmbeddingService {
...
+ * ZeroEntropy native embedding adapter.
+ *
+ * Reuses {@link OpenAIEmbeddingConfig} for the wire-config shape (baseUrl /
+ * apiKey / model / dimensions are identical), but the wire format diverges
+ * in three places, so we keep this provider on its own class instead of
+ * branching {@link OpenAIEmbeddingService}:
+ *
+ * 1. Endpoint is `${baseUrl}/models/embed` (not `/embeddings`).
+ * 2. Request body requires `input_type` (`"query"` or `"document"`) and
+ *    rejects an explicit `dimensions` field (zembed-1 is fixed-dim).
```
Xuruida: One more thing: although ZeroEntropy treats dimensions as optional, we should still allow users to configure it explicitly.

Right now, dimensions from EmbeddingOptions is not forwarded to the ZeroEntropy /models/embed request, so users have no way to control embedding dimensionality end-to-end.

Please pass dimensions through if sendDimensions is set so this isn’t silently ignored.

Happy to help adjust the mapping if you want 👍

ps: https://docs.zeroentropy.dev/api-reference/models/embed
Image

Xuruida (changes requested):

YOMXXX:
You're right @Xuruida — verified against [the docs](https://docs.zeroentropy.dev/api-reference/models/embed): `dimensions` is an optional integer; for zembed-1 the accepted set is `[2560, 1280, 640, 320, 160, 80, 40]` (Matryoshka). My earlier "fixed-dim, rejects overrides" comment in the code was just wrong, sorry about that.

Pushed [`af244e9`](https://github.com/YOMXXX/TencentDB-Agent-Memory/commit/af244e9):

- `ZeroEntropyEmbeddingService` now mirrors `OpenAIEmbeddingService`: reads `config.sendDimensions ?? true` into a `sendDimensions` field and emits `dimensions: this.dims` in the request body when set.
- Class docstring corrected — the misleading "rejects explicit overrides" line is replaced with the documented Matryoshka set.
- Forwarding is verbatim, not clamped. Server still rejects out-of-set values, but that's a config choice the user owns; silent clamping would surprise users who deliberately picked a smaller dim for storage savings.

This also fixes a latent invariant violation I missed in the first round: previously the request omitted `dimensions` while `getDimensions()` kept returning `config.dimensions`, so the runtime vector size and the configured one could disagree.

Let me know if there are any other knobs you'd like me to surface (e.g. `encoding_format` / `latency` are also documented but felt out of scope for this PR — happy to add them in a follow-up).

Xuruida:
LGTM​.

Tests pass and outputs match the ZeroEntropy spec.

Clipboard_Screenshot_1780561675
Clipboard_Screenshot_1780561688

Merging now — thanks!

PR #144, opened 2026-06-04
[good first issue] docs: recommend exact version install for OpenClaw

### LINKED ISSUE #107: [good first issue]🎯 package.json uses caret range (^) which causes plugin to be disabled when installed via OpenClaw
When `@tencentdb-agent-memory/memory-tencentdb` is installed via `npm install` (or via OpenClaw's plugin system), the `package.json` dependency is written with a caret range:

```json
"@tencentdb-agent-memory/memory-tencentdb": "^0.3.6"
```

OpenClaw's plugin manager does **not** support npm semver ranges — it requires exact versions or dist-tags. On the next OpenClaw update, the plugin is silently disabled with:

```
Disabled "memory-tencentdb" after plugin update failure.
Failed to update memory-tencentdb: unsupported npm spec: use an exact version or dist-tag (ranges are not allowed)
```

**Suggested fix:** Either:
1. Pin the dependency to an exact version in the install docs (`npm install --save-exact`), or
2. Note in the README that OpenClaw requires exact versions, and suggest users run `npm install @tencentdb-agent-memory/memory-tencentdb@0.3.6 --save-exact` instead of `@latest`.

yangjj-iso:
﻿## Description | 描述
Updates the OpenClaw installation docs to recommend installing `@tencentdb-agent-memory/memory-tencentdb` with an exact version. This avoids npm semver ranges such as `^0.3.6`, which OpenClaw does not support during plugin updates.

## Related Issue | 关联 Issue
Fix #107

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明
Validation performed:
- `npm install --ignore-scripts --no-package-lock`
- `npm test`
- `npm pack --dry-run`
- `git diff --check`
- `openclaw --profile codex-issue-107 plugins install @tencentdb-agent-memory/memory-tencentdb@0.3.6`
- `openclaw --profile codex-issue-107 --log-level debug plugins install @tencentdb-agent-memory/memory-tencentdb@0.3.6 --pin`
- `npm pack @tencentdb-agent-memory/memory-tencentdb@0.3.6`
- `tar -xzf tencentdb-agent-memory-memory-tencentdb-0.3.6.tgz`

Notes:
- Plain `npm install` failed on Windows because the existing postinstall command uses `bash ... || true`, and `true` is not available in the current Windows cmd environment. CI uses `npm install --ignore-scripts`, so this docs-only PR follows the CI install path.
- The OpenClaw exact-version install command was accepted and reached the download/extract step, but full installation was blocked on Windows by an existing OpenClaw extraction error: `failed to extract archive: Error: EINVAL: invalid argument, open '.../package/LICENSE'`. The npm tarball itself downloads and extracts successfully with system `tar`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,13 +135,14 @@ graph LR
 ### 1.1 Install the plugin
 
 ```bash
-openclaw plugins install @tencentdb-agent-memory/memory-tencentdb@0.3.6
+openclaw plugins install @tencentdb-agent-memory/memory-tencentdb
 openclaw gateway restart
 ```
 
-For OpenClaw installs, use an exact package version. OpenClaw does not support
-npm semver ranges such as `^0.3.6`; if you install with npm directly, run
-`npm install --save-exact @tencentdb-agent-memory/memory-tencentdb@0.3.6`.
+> Please use the native OpenClaw command to upgrade the plugin. This approach prevents the plugin from being disabled caused by semantic version ranges.
+> ```bash
+> openclaw plugins update @tencentdb-agent-memory/memory-tencentdb
+> ```
 
 ### 1.2 Zero-config to enable
 
--- a/README.md
+++ b/README.md
@@ -135,9 +135,13 @@ graph LR
 ### 1.1 Install the plugin
 
 ```bash
-openclaw plugins install @tencentdb-agent-memory/memory-tencentdb
+openclaw plugins install @tencentdb-agent-memory/memory-tencentdb@0.3.6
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Thank you for participating in our good first issue event, we will complete the review internally as soon as possible!

PR #181, opened 2026-06-10
fix(record): harden L1 parsing and embedding timeout

### LINKED ISSUE #110: L1 JSON parse failure + background embedding timeout abort — memory pipeline stalls
## Environment
- Plugin: `@tencentdb-agent-memory/memory-tencentdb` v0.3.6
- OpenClaw: 2026.5.19
- L1 extraction model: deepseek-v4-flash
- Embedding: ollama nomic-embed-text (CPU)

## Symptoms

Every session starts cold — recall always returns 0 hits, no memories ever land in the store.

```
[memory-tdai] [recall] hybrid search: hits=0, fts=0hits, vec=0hits
[memory-tdai] [pipeline-factory] [l1] L1 complete: extracted=0, stored=0 (1 group(s))
[memory-tdai] [pipeline] L2 cold-start skip: not updating l2LastRunTime
```

## Root Cause 1: L1 extraction JSON parse failure (primary)

deepseek-v4-flash outputs non-standard JSON — unescaped Chinese characters and trailing text after the JSON object. The current `cleaned` logic in L1 extraction cannot recover from these outputs:

```
[memory-tdai][l1-extractor] Failed to parse extraction result:
Unexpected token 姓, "[姓名]）在 [时间]"... is not valid JSON

[memory-tdai][l1-extractor] Failed to parse extraction result:
Unexpected non-whitespace character after JSON at position 1174
```

This causes `extracted=0, stored=0` for all L1 groups, leaving the L1 records table empty.

## Root Cause 2: background embedding aborted by timeout (secondary)

The background embedding task uses `AbortController` with a default timeout of 10,000ms. ollama `nomic-embed-text` on CPU routinely exceeds this (>12s per batch), causing the embed request to be aborted before completion:

```
[memory-tdai] [capture] [L0-vec-index-bg] Background embedding failed
(12938ms, non-fatal): This operation was aborted
```

The configured `captureTimeoutMs: 15000` does not reach the embedding API layer — it governs only the capture flow, not the underlying `fetch` call.

L0 metadata + FTS is written successfully, but `vectors.db` never receives vector data.

## Combined effect

With no L1 records and no vectors, the entire store→recall pipeline is blocked:
- `tdai_memory_search` returns nothing
- auto-recall always 0 hits
- L2 scene/persona never triggers

RerankerGuo:
## Description | 描述
Harden the L1 extraction path against non-strict LLM JSON output and ensure capture-path embedding timeout configuration reaches vector dual-write calls.

- Parse the first valid scene payload instead of greedily matching from the first `[` to the last `]`, so explanatory text like `[姓名]）在 [时间]` no longer poisons extraction.
- Accept object-wrapped scene arrays such as `{ "scenes": [...] }` while still ignoring unrelated JSON before the actual scene payload.
- Forward `embedding.captureTimeoutMs` / capture embedding timeout through `writeMemory()` so slow CPU embedding providers are not limited by the default embedding timeout during L1 vector writes.

## Related Issue | 关联 Issue
Fix #110

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

### DIFF
--- a/src/core/record/l1-extractor.test.ts
+++ b/src/core/record/l1-extractor.test.ts
@@ -1,0 +1,50 @@
+import { describe, expect, it } from "vitest";
+import { parseExtractionResult } from "./l1-extractor.js";
+
+describe("parseExtractionResult", () => {
+  it("ignores non-JSON bracketed text before the extraction array", () => {
+    const raw = `
+[姓名]）在 [时间] 这类说明不是 JSON。
+[
+  {
+    "scene_name": "排障",
+    "message_ids": ["m1"],
+    "memories": [
+      {
+        "content": "用户使用 ollama nomic-embed-text 作为 embedding 模型",
+        "type": "episodic",
+        "priority": 60,
+        "source_message_ids": ["m1"],
+        "metadata": {}
+      }
+    ]
+  }
+]
+补充说明：以上为提取结果。
+`;
+
+    const scenes = parseExtractionResult(raw);
+
+    expect(scenes).toHaveLength(1);
+    expect(scenes[0]?.scene_name).toBe("排障");
+    expect(scenes[0]?.memories[0]?.content).toBe("用户使用 ollama nomic-embed-text 作为 embedding 模型");
+  });
+
+  it("accepts object-wrapped scene arrays with trailing text", () => {
+    const raw = `{"scenes":[{"scene_name":"配置","message_ids":["m1"],"memories":[]}]} trailing text`;
+
+    const scenes = parseExtractionResult(raw);
+
+    expect(scenes).toHaveLength(1);
+    expect(scenes[0]?.scene_name).toBe("配置");
+  });
+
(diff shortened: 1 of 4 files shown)


Maxwell-Code07:
Two solid improvements in one PR — better JSON parsing tolerance and proper embedding timeout forwarding. The scope-limited scene array matching is a smart approach. We'll review it.

YOMXXX (approved):
Approved after local verification. This is the canonical #110 fix for hardening L1 scene JSON parsing and forwarding capture-path embedding timeout into L1 vector dual-write calls.

Verified locally:
- COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/core/record/l1-extractor.test.ts src/core/record/l1-writer.test.ts
- git diff --check

PR #186, opened 2026-06-10
feat(gateway): expose session diagnostics

### LINKED ISSUE #64: Suggestion: improve observability for session and L1 pipeline status
During local daily use, one difficulty I repeatedly ran into is observability: when a memory item does not appear as expected, it is hard to quickly tell where the problem is.

Typical questions are:

- did this session enter L0 successfully?
- has L1 extraction already run for this session?
- if L1 is missing, is it because:
  - the trigger threshold was not reached yet,
  - the pipeline is still pending,
  - extraction produced no storable result,
  - dedup/merge removed it,
  - or an actual error occurred?
- if a session is behind, is it temporary lag or a real failure?

At the moment, this can often be inferred by checking SQLite data, checkpoint files, and gateway logs together, but in practice that is still quite difficult during long-running use.

A few improvements that would help a lot:

- clearer per-session status
- structured L0/L1 progress visibility
- grouped failure reasons instead of only raw logs
- optional read-only diagnostics/debug endpoints

To better understand these cases in my own setup, I have been experimenting with a small read-only observability UI as a reference prototype:

https://github.com/sirenexcelsior/tdai-memory-observatory

If you could provide any ideas or guidance, I would be very honored.

Thank you again.

RerankerGuo:
## Description | 描述
Add a read-only Gateway diagnostics endpoint for session and L1 pipeline observability.

`GET /diagnostics/sessions` returns:

- store availability plus L0/L1 record counts when available
- scheduler enabled/started state and queue status
- checkpoint aggregate counters
- per-session in-memory scheduler state, buffered message count, checkpoint runner state, and checkpoint pipeline state

The endpoint is routed through the existing Gateway auth gate, unlike `/health`, because it can expose session keys and operational state.

## Related Issue | 关联 Issue
Fix #64

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [x] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

### DIFF
--- a/src/core/tdai-core.ts
+++ b/src/core/tdai-core.ts
@@ -68,6 +68,37 @@ export interface TdaiCoreOptions {
   instanceId?: string;
 }
 
+export interface TdaiDiagnostics {
+  dataDir: string;
+  store: {
+    vectorStore: boolean;
+    embeddingService: boolean;
+    l0Count: number | null;
+    l1Count: number | null;
+    countError?: string;
+  };
+  scheduler: {
+    enabled: boolean;
+    started: boolean;
+    queues: ReturnType<MemoryPipelineManager["getQueueSizes"]> | null;
+  };
+  checkpoint: {
+    totalProcessed: number;
+    l0ConversationsCount: number;
+    totalMemoriesExtracted: number;
+    memoriesSinceLastPersona: number;
+    scenesProcessed: number;
+  };
+  sessions: Array<{
+    sessionKey: string;
+    trackedInMemory: boolean;
+    bufferedMessages: number;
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:
A diagnostics endpoint is a welcome addition for observability. The read-only design with auth gating is the right approach. We'll review it internally.

YOMXXX (approved):
Approved as the canonical session/L1 observability endpoint for #64.

Local verification:
- `git diff --check`

This is a better fit for #64 than a generic status endpoint because it exposes per-session scheduler/checkpoint/pipeline state and keeps the diagnostics route behind the existing Gateway auth gate.

PR #188, opened 2026-06-10
feat(recall): add injected context visibility option

### LINKED ISSUE #114: feat: Add recall transparency option to show injected memories to users
## Problem

When `autoRecall` is enabled, the plugin automatically injects relevant memories into the user prompt via `<relevant-memories>` tags before the agent processes the message. However, users have **no visibility** into what memories are being injected.

This creates a trust and accuracy risk:
- Old/inaccurate memories may silently influence the agent's behavior
- Users cannot verify or correct injected content
- The `<relevant-memories>` tag is stripped from persisted messages by the `before_message_write` hook, making it invisible in conversation history

## Proposed Solution

Add a config option `recall.showInjected` (boolean, default `false`):

```jsonc
{
  "recall": {
    "showInjected": true
  }
}
```

When enabled, the `before_message_write` hook should skip stripping `<relevant-memories>` tags, preserving them in the conversation history so users can see what was injected.

## Current Workaround

I patched `dist/index.mjs` to add this check:

```javascript
const showInjected = api.pluginConfig?.recall?.showInjected ?? false;
api.on("before_message_write", (event) => {
    if (showInjected) return; // preserve <relevant-memories>
    // ... existing strip logic
});
```

This works but is lost on plugin upgrades.

## Use Case

For long-term collaborative agents, memory accuracy degrades over time. Users need transparency into what the agent "remembers" to catch and correct stale or incorrect memories before they influence responses.

RerankerGuo:
## Description | 描述
Add `recall.showInjected` so users can opt into preserving auto-recall context in conversation history for transparency.

- Parses `recall.showInjected` with a backward-compatible default of `false`.
- Keeps existing transcript cleanup behavior by default, but skips stripping injected recall context when enabled.
- Moves recall stripping into a small tested helper and documents the option in plugin schema and README files.

## Related Issue | 关联 Issue
Fix #114
Related to #120

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [x] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明
Verified with `npx vitest run src/config.test.ts src/utils/recall-injection.test.ts`, `npm test`, and `npm run build` using Node v24.15.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,7 @@ If `MEMORY_TENCENTDB_GATEWAY_API_KEY` is unset, the plugin also looks at `TDAI_G
 | `recall.maxResults` | `5` | Number of items returned per recall |
 | `recall.maxCharsPerMemory` | `0` | Max characters injected for one recalled L1 memory; `0` disables this guard |
 | `recall.maxTotalRecallChars` | `0` | Total character budget for auto-recalled L1 memories; `0` disables this guard |
+| `recall.showInjected` | `false` | Preserve injected recall context in conversation history so users can inspect what memory was used |
 | `pipeline.everyNConversations` | `5` | Trigger an L1 memory extraction every N turns |
 | `extraction.maxMemoriesPerSession` | `20` | Max memories extracted per L1 pass |
 | `persona.triggerEveryN` | `50` | Generate the user persona every N new memories |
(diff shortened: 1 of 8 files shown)


Maxwell-Code07:
Adding user control over injected context visibility is a thoughtful enhancement. The backward-compatible default keeps existing behavior unchanged. We'll review it and get back to you.

YOMXXX (approved):
Approved after local verification. This keeps the default transcript cleanup behavior and makes preserving injected recall context an explicit opt-in via recall.showInjected, which is a low-risk related improvement for #120 and directly addresses #114.

Verified locally:
- COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/config.test.ts src/utils/recall-injection.test.ts
- git diff --check

PR #190, opened 2026-06-10
fix(record): retry unparsable L1 extraction output

### LINKED ISSUE #115: bug: seed command fails silently — L1 extraction JSON parse errors, L2/L3 timer conflicts, no retry
## Description

The `openclaw memory-tdai seed` command has multiple reliability issues that cause it to silently lose data during bulk import.

### 1. L1 extraction JSON parse failures (silent data loss)

The L1 extractor calls the LLM to extract structured memories from conversation pairs. When the LLM returns malformed JSON (common with non-OpenAI models), `parseExtractionResult()` fails and returns an empty array. The extraction count shows `extracted=0, stored=0` with no retry.

```
[l1-extractor] Failed to parse extraction result: Expected ',' or '}' after property value in JSON at position 212
[l1] L1 complete: extracted=0, stored=0 (1 group(s))
```

The aggressive JSON sanitization (`aggressiveSanitizeJson`) handles trailing commas and control chars, but cannot fix structural JSON errors (missing quotes, unescaped special chars in string values, etc).

**Root cause**: Non-OpenAI LLMs (DeepSeek, MiMo, etc) frequently produce JSON with syntax errors in the `content` field when the extracted memory contains markdown, quotes, or special characters.

### 2. L2/L3 timer conflicts with seed main loop

The seed pipeline starts L2/L3 runners with timers. When L2 fires during the seed loop, it can cause:
- `L2 enqueue conflict on queue "L2": task already queued/running`
- The L3 persona generation (`Starting persona generation: 首次冷启动`) blocks the pipeline

The seed loop only waits for L1 idle, but L2/L3 operations can interfere with the pipeline state.

### 3. No retry mechanism

When L1 extraction fails for a batch:
- The failed messages are NOT retried (cursor not advanced: `NOT advancing cursor — messages will be retried`)
- But in practice, the retry never happens because the seed loop moves on
- The checkpoint records `extracted=0` and the messages are lost

### 4. Process exits prematurely

In multiple runs, the seed process exits with code 0 after processing only 10-60 out of 155 rounds. The exit is clean (no error), suggesting the pipeline destroy or L3 completion signal triggers an early exit.

## Expected Behavior

1. L1 extraction should retry on JSON parse failure (at least 1 retry with a simplified prompt)
2. Failed messages should be reliably retried, not silently skipped
3. L2/L3 timers should not interfere with the seed main loop
4. Seed should process ALL input rounds before exiting

## Environment

- OpenClaw 2026.5.22
- TencentDB Agent Memory (latest npm)
- LLM: DeepSeek V4 Pro (for L1 extraction)
- Input: 155 conversation rounds (pseudo-Q&A pairs from memory files)
- Result: Only 10 L1 memories extracted from 310 L0 messages

RerankerGuo:
## Description | 描述
Retry L1 extraction once when the model returns output that cannot be parsed as the required JSON array.

- Keeps valid empty extraction results (`[]`) as-is and does not retry them.
- Reuses the same extraction prompt with an additional retry instruction that asks for JSON only.
- Adds focused coverage for parse-failure retry success and for not retrying a valid empty array.

## Related Issue | 关联 Issue
Fix #115

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明
Verified with `npx vitest run src/core/record/l1-extractor.test.ts`, `npm test`, and `npm run build` using Node v24.15.0.

### DIFF
--- a/src/core/record/l1-extractor.test.ts
+++ b/src/core/record/l1-extractor.test.ts
@@ -1,0 +1,99 @@
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+import { extractL1Memories } from "./l1-extractor.js";
+import type { ConversationMessage } from "../conversation/l0-recorder.js";
+import type { LLMRunner } from "../types.js";
+
+const tempDirs: string[] = [];
+
+describe("extractL1Memories retry", () => {
+  afterEach(async () => {
+    await Promise.all(tempDirs.splice(0).map((dir) => fs.rm(dir, { recursive: true, force: true })));
+  });
+
+  it("retries once when the first extraction response is unparsable", async () => {
+    const dataDir = await fs.mkdtemp(path.join(os.tmpdir(), "tdai-l1-retry-"));
+    tempDirs.push(dataDir);
+    const calls: Array<{ taskId?: string; systemPrompt?: string }> = [];
+    const llmRunner: LLMRunner = {
+      async run(params) {
+        calls.push({ taskId: params.taskId, systemPrompt: params.systemPrompt });
+        if (calls.length === 1) {
+          return '[{"scene_name":"work","message_ids":["msg_1"],"memories":[{"content":"broken quote}]}]';
+        }
+        return JSON.stringify([
+          {
+            scene_name: "work",
+            message_ids: ["msg_1", "msg_2"],
+            memories: [
+              {
+                content: "User prefers concise engineering status updates.",
+                type: "instruction",
+                priority: 80,
+                source_message_ids: ["msg_1"],
+                metadata: {},
+              },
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Retrying on parse failure is a pragmatic improvement for L1 extraction reliability. We appreciate the care taken to not retry valid empty results. We'll review it internally.

YOMXXX (approved):
Approved after local verification. This is the canonical #115 fix for retrying unparsable L1 extraction output once while preserving valid empty [] responses.

Verified locally:
- COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/core/record/l1-extractor.test.ts
- git diff --check

PR #193, opened 2026-06-11
fix(pipeline): stop idle L2 cold-start polling

### LINKED ISSUE #155: Main thread CPU pinning at 100%: aggressive SQLite polling in memory-tdai pipeline
## Description

When running as the context engine in OpenClaw v2026.6.1, the memory-tencentdb plugin (v0.3.6) causes the main Node.js thread to be pinned at 99.9% CPU usage permanently. Total CPU consumption across worker threads reaches ~150% on a single core.

## Environment

- **OS**: Oracle Linux Server 10.1
- **Node.js**: v24.14.1
- **OpenClaw**: 2026.6.1
- **Plugin**: TencentDB-Agent-Memory v0.3.6
- **Memory**: ~1.1 GB RSS consumed by the process

## Symptoms

1. Main thread is permanently at 99.9% CPU (state `R - running`)
2. System load average stays above 2.0 even during idle periods
3. Memory consumption stays at ~1.1GB RSS even when no user interaction is happening

## Log Analysis

The memory-tdai pipeline performs frequent SQLite L1-queries across multiple commitment sessions. Each cycle shows:

1. `[memory-tdai][sqlite] [L1-query]` — runs for each commitment session key every ~10-15 minutes
2. `[memory-tdai] [pipeline] [L2 cold-start skip]` — consistently runs and skips (wasted I/O)
3. `[context-offload] [local-llm]` — makes repeated calls to the configured model
4. `[memory-tdai] [recall] Strategy 'hybrid' requested but EmbeddingService not available, falling back to keyword` — embedding service is not configured

## Key observations

- SQLite L1-queries run even when there are no new messages from any user
- Multiple commitment sessions are being polled constantly (10+ different UUIDs observed)
- The L2 pipeline cold-start runs every cycle even though it always skips because minInterval hasn't passed
- No embedding service available causes fallback to keyword search on every recall

## Expected behavior

The plugin should not consume CPU during idle periods. SQLite polling should have a configurable interval (or be event-driven rather than poll-driven). The L1 and L2 pipeline should not run when there are no new interactions to process.

## Suggested investigation areas

1. **Polling interval**: L1-query appears to run aggressively for each commitment session even when idle. Consider adding a configurable polling interval or switching to a trigger-based approach.
2. **Event loop blocking**: Verify that all SQLite operations use async/promises and do not block the Node.js event loop.
3. **Cold-start cycles**: L2 pipeline cold-start runs every cycle and always skips. Consider checking minInterval before triggering the pipeline query.
4. **Embedding service**: When embedding service is unavailable, cache the strategy decision rather than falling back on every recall.

RerankerGuo:
## Description | 描述
Reduce idle L2 cold-start polling when a session has no L1 cursor yet.

When the first L2 run skips because no records are available, the scheduler still keeps one `maxInterval` retry so delayed L1 writes can be picked up. If that follow-up retry also skips and no cursor has been established, the scheduler stops re-arming the periodic L2 timer for that idle session. The next L1 completion re-arms L2 normally.

This avoids repeated empty SQLite scans for idle sessions while preserving the delayed retry path after new L1 activity.

## Related Issue | 关联 Issue
Fix #155

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

### DIFF
--- a/src/utils/pipeline-manager.test.ts
+++ b/src/utils/pipeline-manager.test.ts
@@ -1,0 +1,104 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { MemoryPipelineManager, type CapturedMessage, type PipelineConfig } from "./pipeline-manager.js";
+
+const config: PipelineConfig = {
+  everyNConversations: 1,
+  enableWarmup: false,
+  l1: {
+    idleTimeoutSeconds: 60,
+  },
+  l2: {
+    delayAfterL1Seconds: 1,
+    minIntervalSeconds: 5,
+    maxIntervalSeconds: 10,
+    sessionActiveWindowHours: 24,
+  },
+};
+
+const message = (content: string): CapturedMessage => ({
+  role: "user",
+  content,
+  timestamp: new Date().toISOString(),
+});
+
+const flushMicrotasks = async (): Promise<void> => {
+  await Promise.resolve();
+  await Promise.resolve();
+};
+
+describe("MemoryPipelineManager", () => {
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("stops idle L2 maxInterval polling after the cold-start retry also skips", async () => {
+    vi.useFakeTimers();
+    vi.setSystemTime(new Date("2026-06-11T00:00:00.000Z"));
+
+    const manager = new MemoryPipelineManager(config);
+    const l1Runner = vi.fn().mockResolvedValue(undefined);
+    const l2Runner = vi.fn().mockResolvedValue({ skipped: true });
+    const persister = vi.fn().mockResolvedValue(undefined);
+
+    manager.setL1Runner(l1Runner);
+    manager.setL2Runner(l2Runner);
+    manager.setPersister(persister);
+
+    await manager.notifyConversation("session-a", [message("first")]);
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Reducing idle SQLite polling is a meaningful performance improvement. The delayed retry path after L1 activity is well thought out. We'll take a look and provide feedback.

YOMXXX (approved):
Reviewed and verified locally. This is the stronger #155 fix because it preserves the first cold-start retry, then stops re-arming maxInterval polling for sessions that still have no L2 cursor. That directly addresses the idle empty SQLite scan loop while keeping the established-cursor path intact.

Verification:
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/utils/pipeline-manager.test.ts`

RerankerGuo:
Maintenance update: rebased onto current `main` (`98d6878`) as single signed commit `e319700`. Revalidated locally: focused regression tests 2/2 passed, full suite 69/69 passed, `pnpm build` passed, and `git diff --check` is clean. Refreshed GitHub CI is 4/4 green.

PR #201, opened 2026-06-12
feat(hermes): pass full messages (incl. tool calls) to Gateway capture endpoint

qsgy-edge:
## Problem

The Hermes sync_turn integration only sends user_content and assistant_content as plain text. Tool call names, arguments, results, and success/failure status are stripped.

This means L0 has no tool call records, L1 extraction operates on incomplete context, and the offload pipeline cannot function.

## Root Cause

Hermes MemoryManager.sync_all() already supports passing the full messages list (including tool_call, tool_result roles) to providers via inspect.signature detection. But memory-tencentdb providers sync_turn signature only declares user_content, assistant_content, and session_id.

## Fix

Add messages: list | None = None to sync_turn signature and forward to client.capture(). The Gateway /capture endpoint already accepts and processes messages (server.ts:932).

No Hermes core changes needed.

## Changes

- __init__.py: Add messages to sync_turn signature, pass to client.capture()
- client.py: Add messages to capture() signature, include in request body

### DIFF
--- a/hermes-plugin/memory/memory_tencentdb/__init__.py
+++ b/hermes-plugin/memory/memory_tencentdb/__init__.py
@@ -862,7 +862,7 @@ class MemoryTencentdbProvider(MemoryProvider):
         """No-op — recall is done synchronously in prefetch()."""
         pass
 
-    def sync_turn(self, user_content: str, assistant_content: str, *, session_id: str = "") -> None:
+    def sync_turn(self, user_content: str, assistant_content: str, *, session_id: str = "", messages: list | None = None) -> None:
         """Send the turn to Gateway for capture (non-blocking).
 
         Threading model:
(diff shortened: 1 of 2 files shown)


qsgy-edge:
Closing: the L0 recorder's `extractUserAssistantMessages()` filters out tool role messages before they reach any downstream pipeline. Passing `messages` to /capture has no effect on tool data — the information is discarded at the recorder level.

The correct path for tool call data is the offload pipeline's `after-tool-call` hook, which is a separate mechanism independent of /capture. This requires a Hermes-side hook (e.g. on_tool_call) that doesn't exist yet.

Will re-open if/when Hermes adds a tool call hook that can feed the offload pipeline.

Maxwell-Code07:
Got it, thanks for the detailed explanation. We understand the current data flow and filtering rules, and know that tool call data relies on hooks on the Hermes side. We will follow up on this requirement once the relevant hooks are available.

Issue #205, opened 2026-06-13
[Bug] Hermes system_prompt.] _block 返回 静态文本，agent 看不到L3 persona


pgjh:
## 问题

TencentDB 的 L3 persona pipeline 已经在正常工作（4 层深度扫描：anchors → interests → interaction → cognitive core），`/recall` 端点也返回了 `recalledL3Persona` 字段。

但 Hermes 插件的 `system_prompt_block()` 返回的是**静态工具使用说明**，没有调用 `/recall` 把已生成的 persona 注入到 agent 的 system prompt 里。画像已经算出来了，但 agent 从来看不到。

## 影响

- 用户花 token 跑了 L3 pipeline，但 agent 完全没用到这些数据
- 对比 OpenClaw：contextEngine slot 能用上全链路能力；Hermes 只能手动调 memory_search 工具

## 建议修复

`system_prompt_block()` 应调用 `/recall` 获取 `recalledL3Persona`，注入到 system prompt。改动 ~15 行。

## 附加问题

1. `on_memory_write` 是 `pass` — 内置 memory 写入不同步到 TencentDB
2. 只有 2 个只读工具 — 没有写入/画像工具，Hermes 侧只能搜不能写

Maxwell-Code07:
Thanks for the bug report! We'll try to reproduce this issue. Could you please provide more information: Hermes version, plugin version used?

pgjh:
> Thanks for the bug report! We'll try to reproduce this issue. Could you please provide more information: Hermes version, plugin version used?

Hermes Agent v0.16.0 (commit 6f43ff55)
Memory plugin v0.3.6

The issue is in the Hermes plugin's `system_prompt_block()` — it returns a static
tool-usage guide block instead of calling the `/recall` endpoint and injecting
`recalledL3Persona` into the system prompt. The L3 pipeline runs fine, but the
agent never sees the computed persona.

YOMXXX:
Triage update for the Hermes persona/system-prompt PRs:

- #195 is **not** a duplicate of this issue; it is the canonical approved fix for #101 and splits Gateway recall output into `system_context` and `prepend_context`.
- #239 has been closed as superseded. It injected the legacy combined `/recall.context` into `system_prompt_block()`, which conflicts with the #195 direction because dynamic L1 recall can end up in the stable system prompt path.
[...]

PR #206, opened 2026-06-13
fix(hermes): inject L3 persona into system_prompt_block

ferminquant:
## Description

The Hermes provider's `system_prompt_block()` returned a static string and ignored the L3 persona that the Gateway had already generated and exposed via the core's auto-recall pipeline. L3 compute was paid but the agent never saw the result.

This wires the existing data path end-to-end: the Gateway's `/recall` response now carries `recalledL3_persona`, and the Hermes provider fetches it (with a short TTL cache) and appends it to the system prompt as a `## User Persona` block when present.

Refs #205 (partial — fixes the main bullet only; sub-bullets 1 and 2 are tracked separately and intentionally out of scope for this PR)

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Code optimization

## Root Cause

Two missing links in the L3 delivery chain:

1. **Gateway side** — `src/gateway/server.ts:handleRecall` built the `RecallResponse` from `appendSystemContext` and the L1 count, but never copied `result.recalledL3Persona` (which the core had already populated in `src/core/hooks/auto-recall.ts:238`) into the wire response. The field existed in the core's `RecallResult` type and was being thrown away at the HTTP boundary.
2. **Provider side** — `hermes-plugin/memory/memory_tencentdb/__init__.py:system_prompt_block` returned a hard-coded string describing the four layers. Even if the Gateway had returned the persona, the provider would have ignored it. Verified by reading the function and by adding the negative-control test (`TestNegativeControl::test_pre_fix_static_block_does_not_contain_persona`), which would have failed on `main` before this fix.

## Changes

- `src/gateway/types.ts` — add optional `recalledL3_persona?: string | null` to `RecallResponse`. Documented as `undefined`-tolerant so older clients keep working.
- `src/gateway/server.ts` — populate the new field in `handleRecall` from `result.recalledL3Persona ?? null`. One-line change.
- `hermes-plugin/memory/memory_tencentdb/__init__.py` — `system_prompt_block()` now fetches `/recall` (empty query, user_id-keyed), caches the result for 60s, and appends a `## User Persona` section when non-empty. Cache key is implicit on the provider instance, which Hermes re-creates per session, so a session switch cannot serve stale persona to the wrong scope.
  - `__init__` gains `self._persona_cache: Dict[str, Any]`.
  - New helpers `_get_cached_persona()` and `_fetch_persona()`. Both swallow exceptions so `system_prompt_block()` remains a pure function.
- `hermes-plugin/memory/memory_tencentdb/tests/test_l3_persona_injection.py` — 9 new tests:
  - 4 for the core behavior (static fallback, persona injected, empty-string treated as absent, null treated as absent)
  - 3 for the cache (TTL hits, TTL expires, no-cache-on-failure for empty cache)
  - 1 for "Gateway died mid-session" (fall back to cached value when a later fetch fails)
  - 1 negative control (the regression test that would have failed on `main`)

## Cache Semantics

| Scenario | Behavior |
|---|---|
| First call, Gateway up, persona present | Fetch, cache for 60s, return persona |
| First call, Gateway up, no persona yet | Fetch, cache empty for 60s, return static block |
| First call, Gateway down | Return static block, **do not** advance `ts` — next call retries |
| Subsequent call within TTL, Gateway up | Return cached value, no network call |
| Subsequent call within TTL, Gateway down | Return cached value (no network) |
| Subsequent call after TTL, Gateway down | Return previous cached value, **do not** advance `ts` |

The asymmetry on failure (don't advance `ts` on the first failure, but do fall back to a stale value on later failures) is intentional: a fresh-empty cache with a sick Gateway should not be stuck for 60 seconds before retrying; a warm cache with a sick Gateway should not retry on every turn.

## Self-test Checklist

- [x] Verified locally

```
hermes-plugin/memory/memory_tencentdb/tests/test_l3_persona_injection.py .......  9 passed
hermes-plugin/memory/memory_tencentdb/tests/test_memory_tencentdb_recovery.py ......  17 passed
hermes-plugin/memory/memory_tencentdb/tests/test_gateway_shutdown_leak.py ...  2 failed, 16 passed (pre-existing on main, not caused by this PR)
```

Pre-existing failures: `test_gateway_shutdown_leak.py` fails on `main` with `'NoneType' object has no attribute 'client'` from `__init__.py:785` (now line 793 after this PR). Confirmed by stashing this branch's changes and running the test on `main` — same failure, same line. Not in scope for #205.

## Additional Notes

**On the field name `recalledL3_persona`:** I deliberately used snake_case (Python-style) on the wire, matching the existing Python client's reading style and the rest of the `RecallResponse` keys (`memory_count`, `session_key`). The core's internal field is camelCase (`recalledL3Persona`) — `server.ts` is the boundary that translates.

**On `query=""` in `_fetch_persona`:** The persona is keyed on `user_id` and the L3 pipeline runs on its own schedule, so the query field is irrelevant for the persona portion of the response. The Gateway's `/recall` handler accepts an empty `query` and will return whatever persona content the L3 pipeline has produced. If you'd rather have a no-op-only endpoint, the cleanest follow-up is a dedicated `GET /persona` route — happy to do that as a follow-up PR if you want, but it's not required for this fix.

**On the 60s TTL:** Arbitrary; the L3 pipeline regenerates every 50 new L1 memories by default, so 60s is a comfortable lower bound. Configurable via `self._PERSONA_CACHE_TTL_SECS` if you want a different default — left as a class attribute rather than a config-schema change to keep this PR narrow.

### DIFF
--- a/hermes-plugin/memory/memory_tencentdb/__init__.py
+++ b/hermes-plugin/memory/memory_tencentdb/__init__.py
@@ -375,6 +375,14 @@ class MemoryTencentdbProvider(MemoryProvider):
         self._user_id = ""
         self._gateway_available = False
         self._initialized = False  # Track if initialize() has been called
+        # L3 persona cache: {"ts": float, "value": str}. Read by
+        # _get_cached_persona() on every system_prompt_block() call;
+        # refreshed at most every _PERSONA_CACHE_TTL_SECS to keep the
+        # prompt builder cheap. Keyed implicitly on (session_id,
+        # user_id) because the provider is constructed per-session in
+        # Hermes — when session_id changes, the provider instance is
+        # re-created and the cache is naturally cleared.
+        self._persona_cache: Dict[str, Any] = {"ts": 0.0, "value": ""}
 
         # Background sync threads.
         # We allow at most _MAX_INFLIGHT_SYNCS in-flight sync threads at any
@@ -816,7 +824,7 @@ class MemoryTencentdbProvider(MemoryProvider):
     def system_prompt_block(self) -> str:
         if not self._gateway_available:
             return ""
-        return (
+        base = (
             "# memory-tencentdb Memory\n"
             f"Active. User: {self._user_id}.\n"
             "Four-layer memory system (L0→L1→L2→L3) with automatic conversation "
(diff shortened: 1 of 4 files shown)


Maxwell-Code07:
Thanks for the contribution! We'll review the L3 persona injection approach.

YOMXXX:
Blocking verification issue found while triaging #205.

TypeScript-side checks passed locally:
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm test` (38/38 passed on this branch)
- `git diff --check`

However, the new Python regression test currently fails during pytest collection before any test body runs:

```text
ModuleNotFoundError: No module named 'agent'
```

Repro command I used in a temporary venv:

```bash
python3 -m venv /tmp/tdai-pr206-pytest
/tmp/tdai-pr206-pytest/bin/python -m pip install -q pytest
/tmp/tdai-pr206-pytest/bin/python -m pytest hermes-plugin/memory/memory_tencentdb/tests/test_l3_persona_injection.py
```

Root cause: because the test file lives under `hermes-plugin/memory/memory_tencentdb/tests`, pytest imports the parent `memory_tencentdb` package during collection, which executes `__init__.py` and imports `agent.memory_provider` before the test module can run its skip/path setup. The existing test needs the same kind of pre-import stubbing/path strategy to happen before package import, or the test layout/import path needs adjusting.

I am not approving this yet; the implementation direction looks relevant for #205, but the regression test must be runnable in this repo checkout.

PR #209, opened 2026-06-13
fix(core): restore LocalStorageBackend auto-wire in TdaiCore constructor (CR-2)

### LINKED ISSUE #208: CR-2 guard fires on every L1 write in OpenClaw standalone mode (auto-wire missing)
### What I see

In the OpenClaw host-adapter path (the in-process plugin install, not the standalone HTTP gateway), every L1 extraction emits this warning 4 times per session (once per memory written):

```
[memory-tdai][l1-writer] [CR-2 guard] writeMemory called without storage adapter; falling back to local fs at <dataDir>/records/2026-06-13.jsonl. In service mode this means JSONL is written to ephemeral pod fs and will be lost on restart. Caller must pass 'storage' to writeMemory.
```

It is reproducible on a fresh single-host install of `memory-tencentdb` v1.0.0 in OpenClaw. The warnings appear on every successful L1 batch, not on failures.

### What I think is happening

The CR-2 guard at `src/core/record/l1-writer.ts:202-217` was added in the 2026-05-19 fix to make a silent data-loss bug visible. The guard's own comment says standalone mode is supposed to be benign because "the gateway auto-wires a `LocalStorageBackend` at startup (`server.ts:199-203`)".

When I trace the OpenClaw path, I see the auto-wire is missing:

1. `src/gateway/server.ts:261` constructs `new TdaiCore({ hostAdapter, config, sessionFilter })` — no `storage`.
2. The OpenClaw plugin entry at the bundled `dist/index.mjs:21372` (the `OpenClawHostAdapter` host) does the same — `new TdaiCore({ hostAdapter, config: cfg, sessionFilter })`.
3. `TdaiCore.constructor` at `src/core/tdai-core.ts:139` sets `this.storage = opts.storage` and never falls back, so `this.storage` is `undefined`.
4. The L1/L2/L3 runners are constructed with `storage: this.storage` (`src/core/tdai-core.ts:525, 541, 556`) and forward `undefined` to `extractL1Memories`, which forwards it to `writeMemory`.
5. `writeMemory` enters the `else` branch and emits the CR-2 warning.

So the warning is correct, but the auto-wire it depends on is not happening for the OpenClaw host-adapter entry — the comment in `l1-writer.ts` documents an intent that is not in the current code.

The standalone HTTP gateway at `server.ts:261` has the same shape, but the same `server.ts` file does construct a `LocalStorageBackend` at lines 362, 558, and 1443, just not at the constructor at 261. So the auto-wire is half-implemented.

### Impact

- **Standalone mode (single host):** noisy logs, but data still lands at `<dataDir>/records/<date>.jsonl` because of the `else` branch's fs fallback. No actual data loss today, but the warning is misleading — it warns about a data-loss risk that is not real in this configuration.
- **Service mode (pod with a `CosStorageBackend`):** the caller is expected to pass `storage` to `TdaiCore`. If they forget, the warning fires and data goes to ephemeral pod fs and is lost on restart. This is the real data-loss path the CR-2 fix was meant to catch, and it is still real.

The current code conflates the two cases in its warning text, which makes the warning un-actionable for the standalone operator.

### Expected

The CR-2 guard should not fire in standalone mode. Either:

1. Restore the auto-wire the guard comment promises: in `TdaiCore.constructor`, default `this.storage` to `new LocalStorageBackend({ rootDir: this.dataDir, logger: this.logger })` when `opts.storage` is undefined. Service-mode callers that pass a `CosStorageBackend` still win because the fallback only fires when undefined. This is a one-line change in the constructor and matches the documented intent.
2. Or, change the warning text so it no longer claims the auto-wire is happening when it is not. Lower-quality fix, since the CR-2 root cause is still latent for service-mode callers who forget to pass `storage`.

I prefer (1). It is closer to the original design intent, makes the guard actually fire only for the real bug (service mode with missing `storage`), and removes the noisy logs in standalone mode.

### Reproduction

Fresh OpenClaw install, `memory-tencentdb` v1.0.0 (npm `@tencentdb-agent-memory/memory-tencentdb` latest). Open a session, send a few turns, wait for an L1 extraction round. Warnings appear in `journalctl --user -u openclaw-gateway.service`. The L1 records themselves do land at `~/.openclaw/memory-tdai/records/<date>.jsonl`, so the data is safe; only the warning text is wrong.

### Proposed fix

In `src/core/tdai-core.ts`:

```ts
import { LocalStorageBackend } from "./storage/local-backend.js";
// ...
constructor(opts: TdaiCoreOptions) {
  // ...
  // CR-2 (2026-05-19): restore the auto-wire for standalone mode that the
  // l1-writer guard comment refers to. The guard expects every writeMemory
  // call to be backed by a StorageAdapter. Service-mode callers pass one
  // explicitly (e.g. CosStorageBackend). Standalone installs omit it and
  // rely on the gateway to fall back to local fs. We restore that
  // auto-wire here. Custom-storage callers still win because the fallback
  // only fires when opts.storage is undefined.
  this.storage = opts.storage ?? new LocalStorageBackend({ rootDir: this.dataDir, logger: this.logger });
}
```

I am happy to send a PR with this change if maintainers agree. I have already tested it locally on v1.0.0 / OpenClaw — CR-2 warnings stop, L1 writes still land at the same JSONL path (the `LocalStorageBackend` wraps the same fs append with `O_APPEND`, so the on-disk format is identical).

### Environment

- `memory-tencentdb` v1.0.0 (npm)
- OpenClaw v2026.5.20
- Node 22.22.0
- Host: WSL2 Ubuntu, single-host install, OpenClaw host-adapter path

ferminquant:
## Summary

Restore the `LocalStorageBackend` auto-wire in `TdaiCore.constructor` that the `l1-writer` CR-2 guard comment refers to but is missing in this build. The auto-wire was documented (`l1-writer.ts:202-217` says "the gateway auto-wires a `LocalStorageBackend` at startup (`server.ts:199-203`)") but neither `server.ts:261` nor the OpenClaw host-adapter entry actually constructs one before constructing `TdaiCore`, so `this.storage` is `undefined` in standalone installs.

This makes the CR-2 guard fire on every L1 write in standalone mode (no real data loss, but the warning is misleading) and leaves the original data-loss path open in service mode for any caller that forgets to pass `storage` explicitly.

## Changes

- `src/core/tdai-core.ts` — in the constructor, default `this.storage` to `new LocalStorageBackend({ rootDir: this.dataDir, logger: this.logger })` when `opts.storage` is undefined. One import line, one logic line, plus a comment explaining the intent.

## Why this shape

- **Why default in the constructor and not at the call site:** there are at least three call sites (`server.ts:134`, `server.ts:261`, the bundled OpenClaw host-adapter) and the gateway may add more. Putting the fallback in one place is harder to forget than a chain of call-site changes.
- **Why `??` and not a config flag:** the documented intent is that this is always safe in standalone mode (the `LocalStorageBackend` writes to the same `dataDir` the existing fs fallback would write to, with the same JSONL format). Adding a flag would imply a case where the operator wants the warning to keep firing, which is the opposite of what the guard comment says.
- **Why not change the warning text instead:** the CR-2 root cause is still latent for service-mode callers who forget to pass `storage`. Lowering the warning would hide a real data-loss path. The constructor auto-wire makes the warning fire only when the operator actually has a problem.

## Testing

I have this running on my local OpenClaw install (`@tencentdb-agent-memory/memory-tencentdb` v1.0.0 + OpenClaw v2026.5.20, single host, WSL2). The CR-2 warnings stop appearing in `journalctl --user -u openclaw-gateway.service` after restart with the patch. L1 records continue to land at `~/.openclaw/memory-tdai/records/<date>.jsonl` with the same on-disk format (the `LocalStorageBackend` uses `O_APPEND` atomic appends, which is the same primitive the previous `fs.appendFile` path used).

I have not run the upstream test suite — my fork is a clean checkout. If maintainers want CI green before merge, I can rebase and run it locally if there's a Makefile or test script I should use.

## Risk

- **Service mode:** none. `opts.storage ?? LocalStorageBackend` only falls back when `opts.storage` is undefined. Callers that pass a `CosStorageBackend` (or any other custom storage) keep using it.
- **Standalone mode:** behavior changes from "warn and write to local fs via `fs.appendFile`" to "write to local fs via `LocalStorageBackend`." The on-disk format is identical (both use `fs.appendFile` under the hood, the storage wrapper just adds the `O_APPEND` flag which is the default for `appendFile`). Records already on disk from the old path remain readable.
- **No new dependencies.** `LocalStorageBackend` is already used in `server.ts:362, 558, 1443`, so the import is already in the dep graph at the package level.

## Out of scope

- The `PipelineFactoryOptions` / `createPipeline` interface also lacks a `storage?` field, which is a related but distinct issue. I left it alone in this PR to keep the diff small. Happy to send a follow-up PR if the maintainers want the data flow to be explicit at every layer.
- The `seed-runtime` path also calls `createPipeline` and may have the same issue. Same — follow-up PR if needed.

## Linked

Closes #208

### DIFF
--- a/src/core/tdai-core.ts
+++ b/src/core/tdai-core.ts
@@ -33,6 +33,7 @@ import type { MemoryTdaiConfig } from "../config.js";
 import type { IMemoryStore } from "./store/types.js";
 import type { EmbeddingService } from "./store/embedding.js";
 import type { StorageAdapter } from "./storage/adapter.js";
+import { LocalStorageBackend } from "./storage/local-backend.js";
 import { performAutoRecall } from "./hooks/auto-recall.js";
 import { reportRecallMetrics } from "./report/metric-tracking-recall.js";
 import { performAutoCapture } from "./hooks/auto-capture.js";
@@ -136,7 +137,16 @@ export class TdaiCore {
     this.runnerFactory = opts.hostAdapter.getLLMRunnerFactory();
     this.sessionFilter = opts.sessionFilter ?? new SessionFilter([]);
     this.instanceId = opts.instanceId;
-    this.storage = opts.storage;
+    // CR-2 (2026-05-19): the l1-writer CR-2 guard expects every writeMemory call
+    // to be backed by a StorageAdapter. In service mode the caller passes one
+    // (e.g. CosStorageBackend). In standalone mode (the typical install on a
+    // single host) the caller usually omits it, expecting the gateway to
+    // auto-wire a LocalStorageBackend. That auto-wire was documented in
+    // l1-writer.ts:202-217 ("server.ts:199-203") but is missing in this build
+    // for the OpenClaw host-adapter entry. We restore the auto-wire here so
+    // standalone installs stop emitting the CR-2 guard warning, while
+    // service-mode callers that pass a custom storage still win.
+    this.storage = opts.storage ?? new LocalStorageBackend({ rootDir: this.dataDir, logger: this.logger });
   }
 
   // ============================
--- a/src/core/tdai-core.ts
+++ b/src/core/tdai-core.ts
@@ -32,7 +32,7 @@ import type {
 import type { MemoryTdaiConfig } from "../config.js";
 import type { IMemoryStore } from "./store/types.js";
 import type { EmbeddingService } from "./store/embedding.js";
-import type { StorageAdapter } from "./storage/adapter.js";
+import { StorageAdapter, type StorageAdapter } from "./storage/adapter.js";
 import { LocalStorageBackend } from "./storage/local-backend.js";
 import { performAutoRecall } from "./hooks/auto-recall.js";
 import { reportRecallMetrics } from "./report/metric-tracking-recall.js";
@@ -146,7 +146,7 @@ export class TdaiCore {
     // for the OpenClaw host-adapter entry. We restore the auto-wire here so
     // standalone installs stop emitting the CR-2 guard warning, while
     // service-mode callers that pass a custom storage still win.
-    this.storage = opts.storage ?? new LocalStorageBackend({ rootDir: this.dataDir, logger: this.logger });
+    this.storage = opts.storage ?? new StorageAdapter(new LocalStorageBackend({ rootDir: this.dataDir, logger: this.logger }));
   }
 
   // ============================


Maxwell-Code07:
Thanks for your attention. We will review it and get back to you soon.

YOMXXX:
Triage note: this PR appears stale against current `main`.

The linked #208 discussion already notes that the current upstream code no longer has the exact CR-2 guard / `TdaiCoreOptions.storage` / `StorageAdapter` path described in the report. That means this one-line constructor fallback no longer maps cleanly onto the current codebase and should not be merged as-is.

Recommendation: close this PR unless maintainers confirm #208 should target an older maintenance branch. If there is still a storage auto-wire problem on current `main`, it should be re-opened with a fresh reproduction and a new focused PR against the current storage flow.

PR #218, opened 2026-06-14
fix(offload): emit render-safe MMD metadata

rocke2020:
## Summary
- Change L2 Mermaid generation guidance to emit metadata as a Mermaid comment (`%% mmd-meta: {...}`) instead of an arbitrary directive (`%%{...}%%`).
- Add shared MMD metadata parsing that supports the new render-safe comment format while preserving legacy `%%{...}%%` parsing.
- Reuse the shared parser in active/history MMD injection paths.
- Add regression coverage for both new and legacy metadata formats.

## Test plan
- [x] `mmdc` renders an MMD file containing `%% mmd-meta: {...}` metadata
- [x] `npx vitest run src/offload/mmd-meta.test.ts`
- [x] `npm test`
- [x] `npm run build`


rocke2020:
currently 'mmdc -i 001-fix-unauthorized-access.mmd -o 001.svg' will raise error; after this fix, we can render safe mmd file from td agent memory.

Maxwell-Code07:
Thanks for the PR! We'll take a look at the MMD metadata format change.

rocke2020:
Follow-up commit `4dfd2df`: fixes one site the format switch missed.

`isEmptyShellMmd` in `before-agent-start.ts` guarded on the literal substring `%%{`. The new render-safe metadata line `%% mmd-meta: {...}` has a space after `%%`, so it does **not** contain `%%{`. A freshly L2-generated minimal MMD is exactly 3 non-empty lines (metadata + `flowchart TD` + one node), so it slipped past the guard and `lines.length <= 3` classified it as an empty shell → `deleteMmd()` on task transition. 

Fix: added a centralized `hasMmdHeaderMeta()` presence check (matches both the comment and legacy directive forms, regardless of which fields are populated — keying on `taskGoal` alone would under-protect empty-taskGoal / timestamp-only metadata) and reused it in the guard, consistent with this PR's parser-centralization. Added regression tests.

Benefit recap (also noted in the L2 prompt): the comment form keeps the `.mmd` a valid, renderable graph, so a human can open it and read the real flowchart — the old `%%{...}%%` directive broke rendering.

rocke2020:
Heads up on a data-loss bug this change introduced (and how I fixed it).

While testing I noticed the format switch broke the empty-shell cleanup. That guard decides whether an MMD is just a throwaway skeleton, and if so it hard-deletes the file (`deleteMmd` is a plain `unlink`). The old check looked for `%%{` to mean "this has metadata, it's a real graph, keep it". The new `%% mmd-meta:` line has a space after `%%`, so it no longer matched — and a freshly generated 3-line graph (metadata + `flowchart TD` + one node) would get deleted on the next task transition. Silent, irreversible.

Fixed in two steps:
- `hasMmdHeaderMeta()` so we recognize metadata in both the new comment form and the old directive form, no matter which fields are filled in (4dfd2df).
- then tightened the delete check itself so it only removes a file that's genuinely empty — no metadata and no `status:`/`summary:`/`Timestamp:` on any node (0254f7d).

Net effect: the cleanup can now only err on the safe side. Worst case it keeps a skeleton it could've removed; it can never delete a graph that has real content. `deleteMmd` only has this one caller, so there's no other path to worry about.

Added tests for both pieces and the build's green.

PR #223, opened 2026-06-15
fix: support linux config paths with unified home resolution

### LINKED ISSUE #103: [good first issue]🎯 Plugin reads platform-specific paths that do not exist on Linux
## Description
The plugin attempts to read several filesystem paths that only exist on specific platforms or configurations, causing repeated ENOENT errors on Linux servers.

## Errors observed (Linux, running as root)

1. `ENOENT: /Users/fh2497/Library/Application Support/Claude/claude_desktop_config.json` — macOS-only Claude Desktop config path
2. `ENOENT: /home/user/agent_memory/memory.json` — hardcoded /home/user/ path
3. `ENOENT: /home/user/data/agent_memory/config/user_profile.yaml` — same issue

## Impact
- Logs are spammed with error entries on every check
- Not a crash, but noisy

## Expected behavior
The plugin should gracefully skip paths that do not exist (log at debug level, not error), and use `os.homedir()` with a configurable fallback.

## Related
- Issue #89 (resolveStateDir crash) — same class of path-resolution bugs

## Environment
- OS: Linux (Enterprise Linux 10.1)
- Plugin version: v0.3.5

jiang1997:
## Description | 描述

- checks the per-user app config directory before falling back to the data dir
  - Linux: `$XDG_CONFIG_HOME/memory-tencentdb` or `~/.config/memory-tencentdb`
  - macOS/Windows paths are supported by the same helper
- uses one shared home fallback chain for config lookup and default data dir:
  `HOME -> USERPROFILE -> os.homedir() -> userInfo().homedir -> /tmp`
- keeps missing optional config candidates silent, avoiding repeated missing-path noise
- adds regression tests for Linux config paths and home-resolution consistency

## Related Issue | 关联 Issue
Fix #103

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

### DIFF
--- a/src/gateway/config.ts
+++ b/src/gateway/config.ts
@@ -12,6 +12,7 @@ import fs from "node:fs";
 import path from "node:path";
 import YAML from "yaml";
 import { getEnv } from "../utils/env.js";
+import { candidateAppConfigFiles, optionalConfigFileExists, resolveHomeDir } from "../utils/config-paths.js";
 import { parseConfig as parseMemoryConfig } from "../config.js";
 import type { MemoryTdaiConfig } from "../config.js";
 import { normalizeDisableThinking } from "../utils/no-think-fetch.js";
@@ -78,8 +79,9 @@ export interface GatewayConfig {
  * Resolution order for config file:
  * 1. `TDAI_GATEWAY_CONFIG` env var (explicit path)
  * 2. `./tdai-gateway.yaml` or `./tdai-gateway.json` in CWD
- * 3. `<dataDir>/tdai-gateway.yaml` or `<dataDir>/tdai-gateway.json`
- * 4. Pure environment-variable config (no file)
+ * 3. Linux/macOS/Windows app config dir (`~/.config/memory-tencentdb/` on Linux)
+ * 4. `<dataDir>/tdai-gateway.yaml` or `<dataDir>/tdai-gateway.json`
+ * 5. Pure environment-variable config (no file)
  */
 export function loadGatewayConfig(overrides?: Partial<GatewayConfig>): GatewayConfig {
   let fileConfig: Record<string, unknown> = {};
@@ -122,7 +124,7 @@ export function loadGatewayConfig(overrides?: Partial<GatewayConfig>): GatewayCo
   // Data config (expand leading ~ to $HOME so Node.js fs/path can resolve it)
   const dataConfig = obj(fileConfig, "data");
   const rawBaseDir = env("TDAI_DATA_DIR") ?? str(dataConfig, "baseDir") ?? resolveDefaultDataDir();
-  const home = getEnv("HOME") ?? getEnv("USERPROFILE") ?? "/tmp";
+  const home = resolveHomeDir();
   const baseDir = rawBaseDir.startsWith("~/") ? path.join(home, rawBaseDir.slice(2)) : rawBaseDir;
 
   // LLM config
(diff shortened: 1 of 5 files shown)


Maxwell-Code07:
Thanks for the contribution! We'll review the Linux config path support.

Xuruida:
Thanks for the contribution! The gateway config path resolution part looks good — clean code, proper XDG support, and solid test coverage.

However, this PR only partially fixes #103. There are two other files with the same fragile home-directory resolution pattern that should be updated in this PR for a complete fix:

1. src/utils/ensure-hook-policy.ts — resolveConfigPath()

Line 155 still uses the old inline pattern without a /tmp fallback:

```ts
const home = getEnv("HOME") ?? getEnv("USERPROFILE") ?? "";
```

This should use resolveHomeDir() instead. The three fs.existsSync calls in this function (lines 145, 151, 158) should also be replaced with optionalConfigFileExists for consistency.

2. src/utils/openclaw-state-dir.ts — resolveOpenClawStateDir()

Line 33 uses bare homedir() with no fallback chain:
```ts
path.join(homedir(), ".openclaw")
```
This should use resolveHomeDir() so all three call sites share the same HOME → USERPROFILE → os.homedir() → userInfo().homedir → /tmp resolution chain.

Both are the same class of bug described in #103. Could you apply resolveHomeDir() to these two files as well?

YOMXXX:
Triage note: #443 is now the canonical verified fix for #103.

I previously could not select #223 because local focused verification still resolved the default port instead of the test fixture config path. #443 covers the same Linux/XDG config-path problem with a shared helper and regression tests for Gateway config, OpenClaw config, and OpenClaw state-dir fallback.

Verified #443 locally with:
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/utils/config-paths.test.ts`
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm test`
- `git diff --check`

Recommendation: close this as superseded by #443.

jiang1997:
> Thanks for the contribution! The gateway config path resolution part looks good — clean code, proper XDG support, and solid test coverage.
> 
> However, this PR only partially fixes #103. There are two other files with the same fragile home-directory resolution pattern that should be updated in this PR for a complete fix:
> 
> 1. src/utils/ensure-hook-policy.ts — resolveConfigPath()
> 
> Line 155 still uses the old inline pattern without a /tmp fallback:
> 
> ```ts
> const home = getEnv("HOME") ?? getEnv("USERPROFILE") ?? "";
> ```
> 
> This should use resolveHomeDir() instead. The three fs.existsSync calls in this function (lines 145, 151, 158) should also be replaced with optionalConfigFileExists for consistency.
> 
> 2. src/utils/openclaw-state-dir.ts — resolveOpenClawStateDir()
> 
> Line 33 uses bare homedir() with no fallback chain:
> 
> ```ts
> path.join(homedir(), ".openclaw")
> ```
> 
> This should use resolveHomeDir() so all three call sites share the same HOME → USERPROFILE → os.homedir() → userInfo().homedir → /tmp resolution chain.
> 
> Both are the same class of bug described in #103. Could you apply resolveHomeDir() to these two files as well?

Sorry for the long delay. I’ve addressed the requested changes and added regression coverage. Could you please take another look when you have time? Thanks!

aleronwang:
Thanks for tackling the Linux config path resolution, @jiang1997 — the XDG
support and regression tests were solid, and thanks for iterating on the review
feedback.

Per @yomxxx's triage (Jul 9), #443 is now the canonical validated fix for #103:
it covers the same Linux/XDG config path problem via a shared helper with
regression tests, and passed focused verification. To avoid two parallel fixes
for the same issue, we're closing this as superseded.

Closing as superseded by #443 — thanks again for the careful work.

PR #232, opened 2026-06-20
fix(tcvdb): escape session filter literals

YOMXXX:
## Summary
- Escape backslashes and double quotes before interpolating `session_key` / `session_id` into TCVDB filter expressions.
- Add a regression test that captures generated L1/L0 query filters for injection-shaped session values.

## Verification
- `npm test -- --run`
- `npm run build`

### DIFF
--- a/src/core/store/tcvdb.test.ts
+++ b/src/core/store/tcvdb.test.ts
@@ -1,0 +1,46 @@
+import { describe, expect, it } from "vitest";
+import { TcvdbMemoryStore } from "./tcvdb.js";
+
+function createStoreWithCapturedQueryFilters(): {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Thanks for this fix! Session filter injection was indeed a security concern — properly escaping backslashes and double quotes, plus adding a regression test, very thorough work 👍

YOMXXX:

This is a narrow injection-boundary fix for TCVDB filter expressions. It escapes backslashes and double quotes before interpolating `session_key` / `session_id`, and includes a regression test for injection-shaped session values.

Current status:
- mergeable

This should be low-risk and security-positive because it only changes filter literal construction and preserves the existing query shape.

PR #234, opened 2026-06-23
fix(offload): invalidate token cache after stripping deleted tool_use blocks

### LINKED ISSUE #233: [Bug] Stale per-message token cache after in-place tool_use strip over-deletes messages

N/A - logic bug in the token cache, independent of OpenClaw version. Reproduced against repo HEAD ee76536.

### Plugin Version | 插件版本

0.3.6

### Operating System | 操作系统

Linux (OS-independent; pure token-accounting logic).

N/A

### Describe the bug | 问题描述

`context-token-tracker.ts` caches per-message token counts in a WeakMap keyed on the message object plus its `_offloaded` flag (`cachedMessageTokens`). An entry is only dropped when `_offloaded` flips or `invalidateTokenCache(msg)` is called.

Two offload hooks strip deleted `tool_use` blocks out of mixed assistant messages in place with `content.splice(...)`, but neither invalidates the cache nor sets `_offloaded`:

- `before-prompt-build.ts` (splice at line 109)
- `llm-input-l3.ts` (splice at line 1377)

So the next `buildTiktokenContextSnapshot` returns the stale pre-strip count for that message. The strip runs in the same pass that precedes the guard and emergency snapshots, so an inflated total can cross the aggressive/emergency thresholds and delete real messages that should have been kept. Every other in-place mutation already calls `invalidateTokenCache` (`l3-helpers.ts:219/257/302`, `llm-input-l3.ts:1019`); these two strip sites are the only ones that skip it.

### To Reproduce | 复现步骤

Seed the cache with a snapshot, strip a deleted `tool_use` block in place the way the hooks do, then snapshot again. Against the real `context-token-tracker.ts`:

1. Build a mixed assistant message (a small text block plus a large `tool_use` block).
2. `buildTiktokenContextSnapshot(...)` once to populate the cache.
3. `content.splice(...)` to drop the `tool_use` block, with no `invalidateTokenCache` call (matching the hooks).
4. `buildTiktokenContextSnapshot(...)` again.

    seed (cached)         : 537
    after strip, no inval : 537   <- stale
    ground truth (fresh)  : 18
    drift                 : 519 tokens

Calling `invalidateTokenCache(msg)` after the splice drops the count to 18 (drift 0).

### Expected behavior | 预期行为

After a `tool_use` block is stripped in place, the next snapshot should reflect the smaller message, the same as it does after `replaceWithSummary` and the other in-place mutations, so the offload guards count the real post-strip size.

N/A

### Additional context | 补充信息

One line per site: call `invalidateTokenCache(msg)` after each strip loop, and add the import in `before-prompt-build.ts` (`llm-input-l3.ts` already imports it).

akhilesharora:
## Description | 描述

`context-token-tracker.ts` caches per-message token counts in a WeakMap keyed on message identity plus `_offloaded`. Two offload hooks strip deleted `tool_use` blocks out of mixed assistant messages in place with `content.splice(...)` but never call `invalidateTokenCache(msg)` and never set `_offloaded`:

- `before-prompt-build.ts` (splice at line 109)
- `llm-input-l3.ts` (splice at line 1377)

So the next `buildTiktokenContextSnapshot` returns the stale pre-strip count, and since the strip runs just before the guard and emergency snapshots, the inflated total can cross the thresholds and over-delete real messages. Every other in-place mutation already invalidates the cache (`l3-helpers.ts:219/257/302`, `llm-input-l3.ts:1019`); these two were the only ones that skipped it. This calls `invalidateTokenCache(msg)` after each strip loop, only when a block was actually removed, and adds the import to `before-prompt-build.ts`.

## Related Issue | 关联 Issue

Closes #233

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Reproduced against the real `context-token-tracker.ts`: seed the cache, strip a deleted `tool_use` block in place, snapshot again.

Before:

    after strip, no inval : 537
    ground truth (fresh)  : 18
    drift                 : 519

After (invalidate following the splice): 18, drift 0.

`npm test` -> 47 passed. No new test added: the strip loops live in non-exported hook closures, so a unit test cannot drive them without a mocking harness the repo does not have. The runtime repro above is the evidence.

### DIFF
--- a/src/offload/hooks/before-prompt-build.ts
+++ b/src/offload/hooks/before-prompt-build.ts
@@ -8,7 +8,7 @@
  */
 import { PLUGIN_DEFAULTS } from "../types.js";
 import { readOffloadEntries, markOffloadStatus } from "../storage.js";
-import { buildTiktokenContextSnapshot } from "../context-token-tracker.js";
+import { buildTiktokenContextSnapshot, invalidateTokenCache } from "../context-token-tracker.js";
 import { traceOffloadDecision } from "../opik-tracer.js";
 import { injectMmdIntoMessages, findHistoryMmdInsertionPoint } from "../mmd-injector.js";
 import { createL3TokenCounter } from "../l3-token-counter.js";
@@ -101,15 +101,18 @@ export function createBeforePromptBuildHandler(
         if (hasDeleted && isAssistantMessageWithToolUse(msg) && !isOnlyToolUseAssistant(msg)) {
           const content = msg.type === "message" ? msg.message?.content : msg.content;
           if (Array.isArray(content)) {
+            let stripped = false;
             for (let j = content.length - 1; j >= 0; j--) {
               const block = content[j] as any;
               if ((block.type === "tool_use" || block.type === "toolCall") && block.id) {
                 const blockIdNorm = normalizeToolCallIdForLookup(block.id);
                 if (stateManager.deletedOffloadIds.has(block.id) || stateManager.deletedOffloadIds.has(blockIdNorm)) {
                   content.splice(j, 1);
+                  stripped = true;
                 }
               }
             }
+            if (stripped) invalidateTokenCache(msg);
           }
         }
         if (msg._offloaded) continue;
--- a/src/offload/hooks/llm-input-l3.ts
+++ b/src/offload/hooks/llm-input-l3.ts
@@ -1369,15 +1369,18 @@ async function fastPathReApply(messages: any[], stateManager: OffloadStateManage
     if (hasDeleted && isAssistantMessageWithToolUse(msg) && !isOnlyToolUseAssistant(msg)) {
       const content = msg.type === "message" ? msg.message?.content : msg.content;
       if (Array.isArray(content)) {
+        let stripped = false;
         for (let j = content.length - 1; j >= 0; j--) {
           const block = content[j] as any;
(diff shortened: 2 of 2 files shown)


Maxwell-Code07:
@akhilesharora Clean fix! Adding invalidateTokenCache() at both splice sites is the right approach — targeted and thorough with regression tests. Nice work 👍

xyaohubery:
已认领本任务

YOMXXX:
Triage note: #347 has been locally verified and approved as the preferred #233 fix. It covers the same cache-invalidation bug, adds an exported shared helper with regression coverage, and also updates the equivalent `src/offload/index.ts` path. I recommend treating this PR as superseded by #347.

aleronwang:
Thanks for the precise root-cause analysis, @akhilesharora — the stale-cache /
over-delete repro (drift 519 → 0) was exactly the kind of evidence that makes
this reviewable.

Per @yomxxx's triage (Jul 9), #347 has been locally verified and approved as the
preferred fix for #233: it covers the same cache-invalidation bug, adds an
exported shared helper with regression coverage, and also updates the equivalent
`src/offload/index.ts` path. To consolidate on one implementation, we're closing
this as superseded.

Closing as superseded by #347 — appreciate the thorough repro.

PR #237, opened 2026-06-24
fix(gateway): reject invalid and oversized JSON bodies

YOMXXX:
## Summary
- return HTTP 400 for malformed JSON request bodies instead of treating them as server errors
- cap gateway JSON request bodies at 8 MiB using both Content-Length and streamed byte counting
- add gateway regression tests for malformed and oversized request bodies

## Test plan
- npx vitest run src/gateway
- npm test
- npm run build

## Notes
- npm run test:cc-plugin -- --run currently fails because package.json has no test:cc-plugin script.

### DIFF
--- a/src/gateway/server.test.ts
+++ b/src/gateway/server.test.ts
@@ -1,0 +1,136 @@
+import { mkdtemp, rm } from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { Readable } from "node:stream";
+import { afterEach, describe, expect, it } from "vitest";
+import { parseConfig } from "../config.js";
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

YOMXXX:

This is a bounded Gateway input-hardening fix:
- malformed JSON now returns HTTP 400 instead of being treated as a server error
- JSON request bodies are capped at 8 MiB using both `Content-Length` and streamed byte counting
- regression tests cover malformed and oversized request bodies

Current status:
- mergeable

This pairs well with the other Gateway hardening work because it reduces malformed/oversized request risk without changing normal request handling.

PR #67, opened 2026-05-21
fix: support Codex integration requirements

YOMXXX:
## Summary
- Add host capability metadata and a standalone `memory-tdai` CLI entry for non-OpenClaw hosts such as Codex.
- Harden offload/standalone LLM integration: secure backend TLS defaults, custom CA support, providerOptions passthrough, multi-slash provider/model parsing, and Unicode-safe storage sanitization.
- Add L1 memory scope (`global|project|session`) plus independent L2/L3 stage switches, schema updates, migration compatibility, and regression tests.
- Fix gateway multi-user isolation by forwarding a stable `userId` to the offload backend and exposing process/data-dir diagnostics in `/health`.
- Add ZeroEntropy native embedding support (`POST <baseUrl>/v1/models/embed`) and preserve qclaw `proxyUrl` / embedding `timeoutMs` through store creation.
- Document installation into an existing Hermes Agent and add ZeroEntropy embedding examples.

## Test Plan
- `npm test`
- `npm run build`
- `npx tsc --noEmit --module NodeNext --moduleResolution NodeNext --target ES2022 --allowImportingTsExtensions --types node --skipLibCheck index.ts src/**/*.ts`
- `npm pack --dry-run --json`
- `git diff --check`

## Issue coverage
- Addresses #68 and #69.
- Improves diagnostics for #64.
- Avoids implementing #66 direct writes in this PR because maintainers noted the bypass risk for L1 dedup/conflict detection.
- Leaves #38 prompt localization for a dedicated prompt/template refactor, matching maintainer feedback that it should land in a next version.

PR #242, opened 2026-06-24
fix(offload): keep backend TLS verification enabled

YOMXXX:
## Summary

- Removes the hardcoded `rejectUnauthorized: false` from HTTPS backend requests so Node keeps TLS certificate verification enabled by default.
- Makes the `BackendClient` constructor timeout parameter actually apply to L1/L1.5/L2/L4 backend calls instead of always using 120s.
- Keeps `/offload/v1/store` on its existing short 10s timeout so reporting remains non-blocking.
- Adds a focused regression test that mocks `https.request` and verifies TLS is not disabled and the configured timeout is used.

This is a small standalone extraction from the larger, conflicting Codex integration PR #67, limited to backend HTTP safety and timeout behavior.

## Verification

- `npx vitest run src/offload/backend-client.test.ts`
- `npm test` (4 files, 48 tests)
- `npm run build`
- `git diff --check`

### DIFF
--- a/src/offload/backend-client.test.ts
+++ b/src/offload/backend-client.test.ts
@@ -1,0 +1,69 @@
+import { EventEmitter } from "node:events";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import { BackendClient } from "./backend-client.js";
+
+const mocks = vi.hoisted(() => ({
+  capturedOptions: [] as any[],
+  httpsRequest: vi.fn(),
+}));
+
+vi.mock("node:https", () => ({
+  request: mocks.httpsRequest,
+}));
+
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

YOMXXX:

This is a small backend HTTP safety fix extracted from a larger integration branch. It keeps TLS certificate verification enabled by default and makes the configured backend timeout apply to L1/L1.5/L2/L4 backend calls, while preserving the short non-blocking timeout for `/offload/v1/store`.

Current status:
- mergeable

I would treat this as a high-priority low-risk safety PR because it removes an unsafe HTTPS default and fixes timeout propagation without changing the backend API contract.

PR #244, opened 2026-06-24
fix(conversation): scope L0 search candidates by session

YOMXXX:
## Summary
- pass the optional `sessionKey` from `tdai_conversation_search` down into L0 store searches before candidate limiting
- add L0 search filter support to the store interface, SQLite FTS/vector paths, and TCVDB L0 hybrid search
- keep the existing tool-level post-filter as a defensive final guard

## Root cause
`executeConversationSearch` accepted `sessionKey`, but searched global L0 candidates first and only filtered after FTS/vector results were already limited. If other sessions occupied the top candidates, the requested session could return no messages even when matching L0 rows existed.

## Verification
- `npx vitest run src/core/tools/conversation-search.test.ts` (RED before fix, green after)
- `npm test`
- `npm run build`
- `git diff --check`

### DIFF
--- a/src/core/store/sqlite.ts
+++ b/src/core/store/sqlite.ts
@@ -32,6 +32,7 @@ import type {
   L1FtsResult,
   L0SearchResult,
   L0FtsResult,
+  L0SearchFilter,
 } from "./types.js";
 import type { Logger } from "../types.js";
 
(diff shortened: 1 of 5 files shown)


Maxwell-Code07:

YOMXXX:

This is a data-isolation fix for `tdai_conversation_search`: it pushes the optional `sessionKey` filter down into L0 candidate retrieval before candidate limiting, while keeping the existing tool-level post-filter as a defensive guard.

Current status:
- mergeable

This should reduce cross-session candidate leakage risk and improve result quality because filtering happens before the top-k/candidate limit is applied.

PR #247, opened 2026-06-24
fix(pipeline): stop idle L2 cold-start polling

### LINKED ISSUE #155: Main thread CPU pinning at 100%: aggressive SQLite polling in memory-tdai pipeline

YOMXXX:
## Summary
- pass the L2 trigger source into `runL2` so timer-originated cold-start retries can be distinguished from fresh L1 activity
- stop re-arming maxInterval polling when the cold-start retry still skips and no L2 cursor exists
- keep maxInterval polling unchanged after L2 establishes a cursor, and re-arm normally on the next L1 event
- add regression coverage for the idle skip loop and the established-cursor path

## Verification
- npm test
- npm run build

Fixes #155
Related: #193

### DIFF
--- a/src/utils/pipeline-manager.test.ts
+++ b/src/utils/pipeline-manager.test.ts
@@ -1,0 +1,105 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { MemoryPipelineManager, type CapturedMessage, type PipelineConfig } from "./pipeline-manager.js";
+
+const config: PipelineConfig = {
+  everyNConversations: 1,
+  enableWarmup: false,
+  l1: {
+    idleTimeoutSeconds: 60,
+  },
+  l2: {
+    delayAfterL1Seconds: 1,
+    minIntervalSeconds: 5,
+    maxIntervalSeconds: 10,
+    sessionActiveWindowHours: 24,
+  },
+};
+
+function message(content: string): CapturedMessage {
+  return {
+    role: "user",
+    content,
+    timestamp: new Date().toISOString(),
+  };
+}
+
+async function flushMicrotasks(): Promise<void> {
+  await Promise.resolve();
+  await Promise.resolve();
+}
+
+describe("MemoryPipelineManager", () => {
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("stops idle L2 maxInterval polling after the cold-start retry also skips", async () => {
+    vi.useFakeTimers();
+    vi.setSystemTime(new Date("2026-06-11T00:00:00.000Z"));
+
+    const manager = new MemoryPipelineManager(config);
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

YOMXXX:
Closing as a duplicate of #193. The implementation is functionally the same idle L2 cold-start polling fix, and #193 is already verified and approved, so keeping one PR avoids duplicate #155 cleanup branches.

PR #182, opened 2026-06-10
fix(openclaw): tolerate missing state dir resolver

RerankerGuo:
## Description | 描述
Harden OpenClaw state directory resolution during plugin registration.

When OpenClaw loads the plugin before `runtime.state` is fully initialized, state directory lookup should not make registration fail. This change keeps the existing priority order but makes it more defensive:

- use `runtime.state.resolveStateDir()` when available,
- fall back to `OPENCLAW_STATE_DIR` when the runtime state is missing, empty, or throws,
- fall back to `~/.openclaw` when no host-provided state directory is available.

Added focused tests for runtime, environment, throwing resolver, and default fallback paths.

## Related Issue | 关联 Issue
Fix #89

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [ ] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

## Additional Notes | 其他说明
Verified with `npm test` and `npm run build` using Node v24.15.0.

PR #248, opened 2026-06-24
fix(openclaw): tolerate uninitialized state dir resolver

### LINKED ISSUE #89: TypeError: Cannot read properties of undefined (reading 'resolveStateDir') on plugin register
## Description

When loading `memory-tencentdb` as a non-bundled OpenClaw plugin (auto-discovered after install), the following error occurs during plugin registration:

```
[plugins] memory-tencentdb failed during register from .../dist/index.mjs:
```

The error originates at line 16969 of the compiled dist:
```js
```

`api.runtime.state` is `undefined` when the plugin registers before the runtime state is fully initialized.

## Environment

- OpenClaw version: 2026.5.22
- Plugin: @tencentdb-agent-memory/memory-tencentdb (latest)
- Install method: npm global install + auto-discovery via plugins.allow empty

### 1. L1 extraction JSON parser greedy regex

The `parseExtractionResult` function uses a greedy regex `/\[[\s\S]*\]/` that captures trailing text when the LLM adds content after the JSON array. This causes `JSON.parse` failures like:
```
Failed to parse extraction result: Unexpected non-whitespace character after JSON at position 564
```

**Local fix**: replaced with bracket-depth tracking (counting `[`/`{` and `]`/`}`) to find the first complete JSON array and ignore trailing text.

### 2. EmbeddingService not available

With default config (`embedding.provider: "none"`), the recall falls back to keyword-only:
```
Strategy "hybrid" requested but EmbeddingService not available, falling back to keyword
```

Would be helpful to document which embedding providers/models are supported and how to configure them (e.g. baseUrl, apiKey, model, dimensions requirements).

## Expected Behavior

Plugin should either:
- Gracefully handle `resolveStateDir()` being unavailable at registration time (e.g. by using a fallback path)
- Or defer registration until runtime state is available

YOMXXX:
## Summary
- make OpenClaw state-dir resolution tolerate an uninitialized or throwing runtime resolver during plugin registration
- trim runtime and environment fallback values while preserving the existing priority order
- add focused coverage for runtime, env, throwing resolver, and default fallback paths

## Verification
- npm test
- npm run build

Fixes #89
Related: #182

### DIFF
--- a/src/utils/openclaw-state-dir.test.ts
+++ b/src/utils/openclaw-state-dir.test.ts
@@ -1,0 +1,51 @@
+import { homedir } from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+
+import { resolveOpenClawStateDir } from "./openclaw-state-dir.js";
+
+describe("resolveOpenClawStateDir", () => {
+  const originalStateDir = process.env.OPENCLAW_STATE_DIR;
+
+  afterEach(() => {
+    if (originalStateDir === undefined) {
+      delete process.env.OPENCLAW_STATE_DIR;
+    } else {
+      process.env.OPENCLAW_STATE_DIR = originalStateDir;
+    }
+  });
+
+  it("uses the runtime state directory when the host provides it", () => {
+    process.env.OPENCLAW_STATE_DIR = "/env/state";
+
+    const stateDir = resolveOpenClawStateDir({
+      resolveStateDir: () => " /runtime/state ",
+    });
+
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

YOMXXX:
Closing as duplicate/superseded by #182.

#182 fixes the same #89 OpenClaw state-dir resolver fallback issue and has been locally verified/approved:
- full Vitest suite passed
- build passed
- `git diff --check` passed

Keeping #182 as canonical avoids maintaining two equivalent fallback implementations.

PR #251, opened 2026-06-25
fix(record): harden L1 dedup JSON parsing

YOMXXX:
## Description | 描述
Harden L1 conflict-detection result parsing so noisy LLM output does not force every extracted memory down the fallback `store` path.

The old parser used a greedy `/\[[\s\S]*\]/` match, so bracketed prose such as `[store/update/merge/skip]` before the JSON decision array could make parsing fail and silently disable dedup for the batch.

This PR:
- scans for balanced JSON arrays/objects instead of using a greedy regex
- accepts object-wrapped decision arrays such as `{ "decisions": [...] }`
- keeps fallback-to-store behavior when no valid decision payload exists
- adds focused regression coverage for noisy bracketed prose and object-wrapped payloads

## Related Issue | 关联 Issue
Related to #110 (same L1 JSON robustness failure mode, on the conflict-detection path).

## Verification | 验证
- RED: `npm test -- src/core/record/l1-dedup.test.ts` failed with fallback `store` decisions
- GREEN: `npm test -- src/core/record/l1-dedup.test.ts`
- `npm test`
- `npm run build`

### DIFF
--- a/src/core/record/l1-dedup.test.ts
+++ b/src/core/record/l1-dedup.test.ts
@@ -1,0 +1,67 @@
+import { describe, expect, it } from "vitest";
+
+import { parseBatchResult } from "./l1-dedup.js";
+import type { ExtractedMemory } from "./l1-writer.js";
+
+function memory(record_id: string): ExtractedMemory & { record_id: string } {
+  return {
+    record_id,
+    content: "用户在诊断场景要求 AI 只回复 OK",
+    type: "instruction",
+    priority: 75,
+    source_message_ids: ["m1"],
+    metadata: {},
+    scene_name: "诊断调试",
+  };
+}
+
+describe("parseBatchResult", () => {
+  it("ignores bracketed prose before and after the conflict decision JSON array", () => {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
@YOMXXX 4 PRs in one day tackling L1 extraction robustness — JSON parsing tolerance, dedup hardening, failure diagnostics persistence, and auto-retry. Great coverage! We'll review them one by one 🚀

YOMXXX:
Product/merge triage note: I rechecked this PR as part of the L1 behavior queue.

Assessment:
- This is a low product-risk parser robustness fix: it does not change the dedup policy itself, only recovers valid decision payloads from noisy LLM output instead of falling back to `store` for the whole batch.
- Focused verification passed locally: `COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/core/record/l1-dedup.test.ts` (1 file, 2 tests).
- `git diff --check upstream/main...HEAD` passed.

Merge-order note: this PR and #265 both add `src/core/record/l1-dedup.test.ts`, so if one lands first, the other will need a small test-file rebase. That is a mechanical test-file conflict, not a functional duplicate.

PR #177, opened 2026-06-10
fix(checkpoint): recalibrate counters after cleanup

RerankerGuo:
## Description | 描述
  Recalibrate checkpoint aggregate counters from the actual memory store counts so cleanup or manual pruning does not leave
  `l0_conversations_count` and `total_memories_extracted` permanently overstated. Also adjust `memories_since_last_persona` downward
  when L1 records are removed, and run recalibration after cleanup and on scheduler startup.

  ## Related Issue | 关联 Issue
  Fix #157

  ## Change Type | 修改类型
  - [x] Bug fix | Bug 修复
  - [ ] New feature | 新功能
  - [ ] Documentation update | 文档更新
  - [ ] Code optimization | 代码优化

  ## Self-test Checklist | 自测清单
  - [x] Verified locally | 本地验证通过
  - [x] No existing features affected | 无影响现有功能

  ## Additional Notes | 其他说明
  Verified with `npm test` and `npm run build` using Node v24.15.0.

PR #253, opened 2026-06-25
fix(checkpoint): recalibrate global counters against ground truth on startup

L2ncE:
## Description

Fixes checkpoint counter drift (issue #157). The four additive-only global counters in `CheckpointManager` — `total_memories_extracted`, `l0_conversations_count`, `total_processed`, `memories_since_last_persona` — only ever increment. When `memory-cleaner` deletes expired data or JSONL files are pruned manually, these counters stay permanently inflated above the actual data and never self-correct. The inflated `memories_since_last_persona` can also spuriously trigger persona generation.

Adds `CheckpointManager.recalibrate()`, invoked once on gateway startup (per the issue's suggested fix), which re-syncs the four counters against authoritative sources:

| Counter | Source of truth |
|---|---|
| `total_memories_extracted` | `records/*.jsonl` total line count (append-only, includes dedup stale rows — same semantics as the counter) |
| `l0_conversations_count` | distinct `recordedAt` in `conversations/*.jsonl` (= capture count) |
| `total_processed` | `store.countL0()` (falls back to JSONL line count when the store is degraded) |
| `memories_since_last_persona` | `records/*.jsonl` lines with `updatedAt > last_persona_time` |

Why JSONL for counters 1/2/4: they accumulate write events (each store/update/merge appends a JSONL line), so JSONL line count is same-sourced; the store dedups and would under-count. Counter 2 uses distinct `recordedAt` because the TCVDB adapter has no DISTINCT/group-by capability, so it must read the JSONL files. Counter 3 is a message count, same in store and JSONL, so the existing `countL0()` is reused.

Trigger timing: recalibrate runs inside the `coreReady` promise chain and is awaited, so the first `agent_end` (which does `await coreReady`) cannot overtake it. This keeps recalibration free of concurrency with the L2 repair path in `pipeline-factory` (which would otherwise `Math.max` the counters back to stale values).

Incremental extraction gates on per-session cursors (`last_l1_cursor` / `last_extraction_updated_time`), not on these global counters, so counter drift cannot cause records to be skipped — issue acceptance #3 is satisfied by the existing design.

## Related Issue

Fix #157

## Change Type

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist

## Additional Notes

- **Issue repro coverage**: the manual-JSONL-pruning repro from the issue is covered by a unit test (`issue repro: manual JSONL pruning`): seed 5 L1 lines → counter = 5 → prune to 2 lines → counter still 5 (drift) → recalibrate → counter = 2. A cleaner-style "delete expired shard file" case is also covered.
- **Scope note**: issue #157's repro also mentions deleting `pipeline_states` entries. `pipeline_states` holds per-session incremental cursors (positional semantics), not global counters — deleting them causes re-processing, not counter drift, and is a separate concern outside this PR. Only the JSONL pruning affects the four counters and is fixed here.
- **Known trade-off**: recalibrate runs only at startup (as the issue requests). Between a `memory-cleaner` run and the next restart the counters remain inflated; this is acceptable since drift is gradual and the next startup corrects it.
- **Test suite**: 84 tests pass (`npx vitest run`), including new unit tests for the count helpers (shard filtering, malformed-line handling, B2 bad-row invariant) and integration tests reproducing the drift and correction.
- **CHANGELOG**: updated under `[Unreleased]` → `🐛 Bug 修复`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ## [Unreleased]
 
+### 🐛 Bug 修复
+
+- **Checkpoint 计数器漂移** ([#157](https://github.com/Tencent/TencentDB-Agent-Memory/issues/157))：`total_memories_extracted`、`l0_conversations_count`、`total_processed`、`memories_since_last_persona` 四个全局计数器只增不减，`memory-cleaner` 清理或人工修剪 JSONL 后永久高估实际数据。新增 `CheckpointManager.recalibrate()`，在 Gateway 启动时用真实数据（L0/L1 JSONL 行数、store `countL0()`）重算这四个计数器，纠正漂移。其中 `memories_since_last_persona` 校准后不再虚高误触发 persona 生成。增量提取门控使用 per-session 游标，不读这些全局计数器，故不会因漂移跳过记录。
+
 ### ✨ 新功能
 
 - **时区可配置** ([#75](https://github.com/Tencent/TencentDB-Agent-Memory/issues/75) / [#87](https://github.com/Tencent/TencentDB-Agent-Memory/issues/87))：新增顶层 `timezone` 配置项，支持 IANA 时区名（`Asia/Shanghai`、`Europe/Berlin`）和 UTC 偏移串（`+08:00`、`-05:30`）。默认 `"system"`（跟随进程系统时区），升级零感。
--- a/index.ts
+++ b/index.ts
@@ -35,6 +35,9 @@ import { registerMemoryTdaiCli } from "./src/cli/index.js";
 import { initDataDirectories, resetStores } from "./src/utils/pipeline-factory.js";
 import { getOrCreateInstanceId, initReporter, report, resetReporter } from "./src/core/report/reporter.js";
 import { ensureL2L3Local } from "./src/core/profile/profile-sync.js";
+import { CheckpointManager } from "./src/utils/checkpoint.js";
+import { countL1JsonlLines, countL1JsonlLinesSince } from "./src/core/record/l1-reader.js";
+import { countL0JsonlStats } from "./src/core/conversation/l0-recorder.js";
 
 // Core abstractions (host-neutral)
 import { OpenClawHostAdapter } from "./src/adapters/openclaw/host-adapter.js";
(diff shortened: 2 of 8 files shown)


Maxwell-Code07:
@L2ncE Welcome as a first-time contributor! Checkpoint counter drift (#157) is a long-standing issue — the recalibrate-on-startup approach is clean and well-implemented. Thanks for the contribution! 👍

YOMXXX:
Triage note: #177 has been locally verified and approved as the canonical #157 fix. It recalibrates checkpoint counters against actual store counts, adjusts memories_since_last_persona safely, and wires recalibration after cleanup plus scheduler startup. This PR appears to cover the same checkpoint-counter drift scope, so I recommend treating it as superseded unless it intentionally adds behavior beyond #177.

chrispfwu:
Thanks @L2ncE for your work on #157.

We’ve reviewed your implementation in #253. Overall, the direction is solid: `recalibrate()` covers all four global counters, the JSONL vs. store semantics are clearly reasoned, and the test coverage is strong enough to reproduce the drift scenarios described in the issue. This part of the work meets our expectations.

However, the PR is **not ready to merge yet**, for two main reasons:

1. **Incomplete trigger coverage**  
   Recalibration currently runs only at Gateway startup. Per the acceptance criteria for #157, checkpoint counters should reflect actual data **after cleanup**, not only after a restart. With the current design, counters can remain inflated between a `memory-cleaner` run and the next restart. Please add **post-cleanup recalibration** (e.g. wire it into `memory-cleaner` after cleanup completes, similar to the approach in #177).

2. **Merge readiness**  
   The PR is currently `mergeable: unstable` and needs to be **rebased onto the latest `main`** with conflicts resolved. This issue also has multiple overlapping PRs; from a maintenance perspective, we prefer a smaller, landable fix. Please briefly explain what #253 adds beyond other approaches (especially #177), so we can avoid duplicate implementations.

### Next steps

Once you address the following, we can treat this as a completed delivery for #157:

- [ ] Trigger `recalibrate()` (or equivalent logic) after `memory-cleaner` finishes cleanup
- [ ] Rebase onto latest `main`, ensure CI passes, and confirm the PR is mergeable
- [ ] Update the PR description with a short note on why JSONL is used as the source of truth for some counters, and how this differs from #177

After that, we’ll prioritize review and provide a final merge/close decision.

Thanks again for the thoughtful contribution.

PR #254, opened 2026-06-25
fix(record): retry unparsable L1 extraction output

### LINKED ISSUE #115: bug: seed command fails silently — L1 extraction JSON parse errors, L2/L3 timer conflicts, no retry

YOMXXX:
## Description | 描述
Retry L1 extraction once when the model returns output that cannot be parsed as the required JSON scene array.

This addresses the L1 parse-failure part of seed data loss in #115. Non-OpenAI-compatible models can produce structurally invalid JSON; previously `parseExtractionResult()` returned `[]`, so the batch ended as `extracted=0, stored=0` with no second attempt.

This PR:
- distinguishes parse failures from a valid empty extraction array
- retries exactly once on parse failure with a stricter JSON-only system instruction
- keeps valid `[]` responses as-is and does not retry them
- reuses the same runner instance on the CleanContextRunner path
- adds focused regression coverage for retry success and valid-empty no-retry behavior

## Related Issue | 关联 Issue
Fixes #115

## Verification | 验证
- RED: `npm test -- src/core/record/l1-extractor-retry.test.ts` failed because only `l1-extraction` ran once
- GREEN: `npm test -- src/core/record/l1-extractor-retry.test.ts`
- `npm test`
- `npm run build`

### DIFF
--- a/src/core/record/l1-extractor-retry.test.ts
+++ b/src/core/record/l1-extractor-retry.test.ts
@@ -1,0 +1,103 @@
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+
+import type { ConversationMessage } from "../conversation/l0-recorder.js";
+import type { LLMRunner } from "../types.js";
+import { extractL1Memories } from "./l1-extractor.js";
+
+const tempDirs: string[] = [];
+
+describe("extractL1Memories retry", () => {
+  afterEach(async () => {
+    await Promise.all(tempDirs.splice(0).map((dir) => fs.rm(dir, { recursive: true, force: true })));
+  });
+
+  it("retries once when the first extraction response is unparsable", async () => {
+    const baseDir = await fs.mkdtemp(path.join(os.tmpdir(), "tdai-l1-retry-"));
+    tempDirs.push(baseDir);
+
+    const calls: Array<{ taskId: string; systemPrompt?: string }> = [];
+    const llmRunner: LLMRunner = {
+      async run(params) {
+        calls.push({ taskId: params.taskId, systemPrompt: params.systemPrompt });
+        if (calls.length === 1) {
+          return '[{"scene_name":"工程状态同步","message_ids":["msg-1"],"memories":[{"content":"broken quote}]}]';
+        }
+
+        return JSON.stringify([
+          {
+            scene_name: "工程状态同步",
+            message_ids: ["msg-1", "msg-2"],
+            memories: [
+              {
+                content: "用户偏好工程实现过程中的状态更新保持简洁。",
+                type: "instruction",
+                priority: 80,
+                source_message_ids: ["msg-1"],
+                metadata: {},
+              },
+            ],
+          },
+        ]);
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
@YOMXXX 4 PRs in one day tackling L1 extraction robustness — JSON parsing tolerance, dedup hardening, failure diagnostics persistence, and auto-retry. Great coverage! We'll review them one by one 🚀

YOMXXX:
Closing as duplicate/superseded by #190.

#190 covers the same #115 parse-failure retry behavior: retry unparsable L1 extraction output once while preserving valid empty [] responses. I verified #190 locally with focused tests, full test suite, build, and diff check, then approved it.

PR #262, opened 2026-06-26
feat(recall): add show injected visibility option

### LINKED ISSUE #114: feat: Add recall transparency option to show injected memories to users

YOMXXX:
## Summary
- add `recall.showInjected` config with a safe default of `false`
- preserve `<relevant-memories>` in visible user-message history when enabled
- extract the relevant-memory stripping logic into a tested helper for string and text-part content

Fixes #114
Refs #11

## Verification
- npm test -- src/utils/recall-visibility.test.ts
- npm test
- npm run build

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -31,6 +31,7 @@ import {
 } from "./src/utils/clean-context-runner.js";
 import { SessionFilter } from "./src/utils/session-filter.js";
 import { LocalMemoryCleaner } from "./src/utils/memory-cleaner.js";
+import { stripRelevantMemoriesFromContent } from "./src/utils/recall-visibility.js";
 import { registerMemoryTdaiCli } from "./src/cli/index.js";
 import { initDataDirectories, resetStores } from "./src/utils/pipeline-factory.js";
 import { getOrCreateInstanceId, initReporter, report, resetReporter } from "./src/core/report/reporter.js";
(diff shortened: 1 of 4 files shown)


Maxwell-Code07:

YOMXXX:
Closing as duplicate/superseded by #188.

#188 covers the same recall.showInjected feature with docs/schema updates, a tested helper, and backward-compatible default false. I verified and approved #188 earlier with focused tests, full test suite, build, and diff check.

Issue #82, opened 2026-05-24
L1 memory extraction pipeline: 4 improvements to reduce LLM dependency


yuanrengu:
## Summary

The L1 memory extraction pipeline (`l1-extractor.ts`) currently relies entirely on a single LLM call for scene segmentation, memory extraction, priority scoring, and type classification. This makes the quality of extracted memories brittle — LLM hallucination, JSON formatting errors, or low-quality output all go undetected.

This issue proposes 4 targeted improvements, implemented and tested on a fork.

---

## Problem 1: L1 quality gate is neutered

**File:** `src/utils/sanitize.ts` → `shouldExtractL1()`

Length filters and prompt-injection detection are commented out:

```typescript
// const isCJK = /[\u4e00-\u9fff...]/.test(text);
// if (isCJK && text.length < 2) return false;
// if (!isCJK && text.length < 2) return false;
// if (text.length > 5000) return false;
// ...
// if (looksLikePromptInjection(text)) return false;
```

This means short messages ("好的", "OK", "hi"), prompt-injection payloads, and excessively long pasted logs all enter the LLM pipeline, wasting tokens and risking injection into persistent memory.

**Fix:** Re-enable filters with appropriate thresholds (CJK ≥ 4 chars, alpha ≥ 10 chars), add conversational filler filter, and enable prompt-injection guard.

---

## Problem 2: No rule-based pre-extraction — everything goes to LLM

Even obviously structured patterns like "我是Python工程师" (identity) or "以后请用中文回复" (instruction) require a full LLM round-trip. This wastes tokens and introduces latency for clear-cut cases.

**Fix:** Add `src/core/record/pre-extractor.ts` — a new rule-based extraction layer that runs BEFORE the LLM call:
- 10 persona patterns: 我喜欢/我是/我的职业是/我擅长/我认为…
- 8 instruction patterns: 以后/记住/禁止/从现在开始/使用X语言回复…
- Date + action-verb episodic detection ("2025-03-15 部署了新版本")

HIGH-confidence matches bypass the LLM entirely and are directly merged into results. MEDIUM-confidence matches are passed as hints.

**Impact:** For common patterns, no LLM call is needed at all. For everything else, the LLM gets higher-quality input.

---

## Problem 3: JSON parse failures are silently discarded

[...]

PR #266, opened 2026-06-26
fix(l1): drop ungrounded extracted memories

YOMXXX:
## Summary

- Adds a conservative post-LLM validation pass for L1 extraction output.
- Drops extracted memories whose significant tokens have too little overlap with their source messages, preventing obvious hallucinated facts from reaching L1 storage.
- Keeps normal paraphrases/rewrites by using a low overlap threshold and falling back to the whole qualified batch when source IDs are missing.

## Why

Issue #82 notes that LLM extraction output currently goes directly to dedup/storage without quality validation. A model can emit plausible but unsupported persona/instruction facts, and those facts then become durable memory. This PR adds a narrow source-grounding gate before dedup/write.

## Tests

- `npm test -- src/core/record/l1-extractor.test.ts`
- `npm test`
- `npm run build`

Refs #82

PR #269, opened 2026-06-27
feat(l1): pre-extract direct memory rules

YOMXXX:
## Summary
- add a conservative high-confidence rule pre-extractor for explicit persona and instruction statements
- bypass the LLM only when every new user message is covered by direct rules
- keep complex or partially covered batches on the existing LLM extraction path

## Tests
- `npm test -- src/core/record/l1-extractor.test.ts`
- `npm test`
- `npm run build`

Refs #82

### DIFF
--- a/src/core/record/l1-extractor.test.ts
+++ b/src/core/record/l1-extractor.test.ts
@@ -1,0 +1,145 @@
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { extractL1Memories } from "./l1-extractor.js";
+import type { LLMRunner } from "../types.js";
+
+describe("extractL1Memories rule pre-extraction", () => {
+  let tmpDir: string;
+
+  beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

YOMXXX:
Product/merge triage note: I rechecked this PR as part of the L1 behavior queue.

Assessment:
- This is not a duplicate of #266. This PR adds a deterministic pre-extraction path for short, explicit, high-confidence user statements and only bypasses the LLM when every new user message is covered by direct rules.
- Product trade-off: this improves reliability/cost for obvious persona/instruction statements, but it also introduces hardcoded rule patterns and fixed phrasing/priority for those direct memories. Maintainers should decide whether that deterministic behavior belongs in L1 by default.
- Focused verification passed locally: `COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/core/record/l1-extractor.test.ts` (1 file, 3 tests).
- `git diff --check upstream/main...HEAD` passed.

Merge-order note: #266 and #269 both add `src/core/record/l1-extractor.test.ts`; merging both will require a small test-file rebase. The two approaches are complementary, not mutually exclusive: direct-rule pre-extraction happens before the normal LLM path, while post-LLM grounding validation protects the storage boundary.

PR #274, opened 2026-06-27
fix(offload): honor backend timeout config

YOMXXX:
## Summary
- store the `backendTimeoutMs` value passed into `BackendClient` instead of ignoring it
- use the configured timeout for L1/L1.5/L2/L4 backend calls and `/offload/v1/store` state uploads
- add regression tests proving the constructor timeout drives the actual HTTP request timer

## Why
`registerOffload()` already passes `offloadConfig.backendTimeoutMs` into `BackendClient`, but the constructor parameter was ignored. Backend-mode LLM calls always used a hardcoded 120s timeout, while state uploads always used a hardcoded 10s timeout, so user configuration could not tune slow or strict backend deployments.

## Tests
- `npm test -- src/offload/backend-client.test.ts`
- `npm test`
- `npm run build`

### DIFF
--- a/src/offload/backend-client.test.ts
+++ b/src/offload/backend-client.test.ts
@@ -1,0 +1,80 @@
+import { EventEmitter } from "node:events";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { BackendClient } from "./backend-client.js";
+import type { PluginLogger } from "./types.js";
+
+const httpMocks = vi.hoisted(() => ({
+  request: vi.fn(),
+}));
+
+vi.mock("node:http", () => ({
+  request: httpMocks.request,
+}));
+
+vi.mock("node:https", () => ({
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

YOMXXX:
Closing this PR to reduce duplicate/conflicting review surface with #242.

#242 already keeps BackendClient TLS verification enabled and makes the configured backend timeout apply to L1/L1.5/L2/L4 backend calls. The only behavioral difference here is making /offload/v1/store use the same configurable timeout, while #242 intentionally keeps store uploads on the short non-blocking timeout. That shorter upload timeout is the safer merge target for now.

If we later want configurable store-upload timeout, it should be a separate follow-up with explicit discussion of the shutdown/reporting tradeoff, not merged as a competing BackendClient timeout fix.

PR #275, opened 2026-06-28
fix(config): align offload timeout default

YOMXXX:
## Summary
- align `openclaw.plugin.json` default for `offload.backendTimeoutMs` with the runtime `parseConfig({})` default
- update the `OffloadConfig` comment to the same 120000ms default
- add a regression test that compares the manifest schema default with the parsed runtime default

## Why
`parseConfig()` defaults `offload.backendTimeoutMs` to 120000ms, but the manifest schema advertised 10000ms. Hosts that materialize schema defaults can therefore override the runtime default and make backend L1/L1.5/L2/L4 calls time out after 10 seconds instead of the intended 120 seconds.

## Tests
- `npm test -- src/config.test.ts`
- `npm test`
- `npm run build`

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -163,7 +163,7 @@
           "mmdMaxTokenRatio": { "type": "number", "default": 0.2, "description": "MMD 注入 token 预算比例" },
           "backendUrl": { "type": "string", "description": "后端服务 URL（如 https://offload-api.example.com），配置后 L1/L1.5/L2/L4 走后端" },
           "backendApiKey": { "type": "string", "description": "后端 API 认证 token" },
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

YOMXXX:
Merge-order triage note: I checked this with the config/timeout group.

This PR is not a duplicate of #278. #275 aligns the manifest default for `offload.backendTimeoutMs` with the runtime parser default; #278 exposes additional already-supported offload fields in the manifest.

Merge-order note: #275 and #278 both edit `openclaw.plugin.json`, so whichever lands second will need a small manifest rebase. The scopes are complementary.

PR #276, opened 2026-06-28
fix(offload): honor local L2 timeout config

YOMXXX:
## Summary
- remove the hardcoded 120s timeout override from local-mode L2 generation
- let L2 use the same configured `timeoutMs` stored on `LocalLlmClient` as L1/L1.5
- add a regression test for the effective L2 timeout

## Why
`registerOffload()` passes `offload.backendTimeoutMs` into `LocalLlmClient` for local mode. L1 and L1.5 already inherit that config via `callLlm`, but L2 overrode it with `120_000`, so users could not shorten or lengthen local L2 calls consistently.

## Tests
- `npm test -- src/offload/local-llm/index.test.ts`
- `npm test`
- `npm run build`

### DIFF
--- a/src/offload/local-llm/index.test.ts
+++ b/src/offload/local-llm/index.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, it, vi } from "vitest";
+import { LocalLlmClient } from "./index.js";
+
+const llmMocks = vi.hoisted(() => ({
+  callLlm: vi.fn(async () => JSON.stringify({
+    file_action: "write",
+    mmd_content: "flowchart TD\n  A-->B",
+    node_mapping: {},
+  })),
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

YOMXXX:
Ready triage note: I checked this with the timeout/config group.

This PR has independent scope: it removes the local-mode L2 hardcoded timeout override so L2 honors the configured `LocalLlmClient` timeout, matching L1/L1.5 behavior. I did not find duplicate scope with #277 or #279; those address different runner layers/call sites.

Merge-order note: the broader timeout/config group has some expected test/manifest overlap, but this PR itself is a narrow local L2 timeout fix.

PR #277, opened 2026-06-28
fix(l1): honor runner timeout config

YOMXXX:
## Summary
- stop passing hardcoded 180s timeouts to host-neutral L1 extraction and conflict-detection runners
- keep the existing 180s timeout only for the legacy CleanContextRunner fallback path
- add regression tests for extraction and dedup so configured runner timeouts are not overridden

## Why
Standalone/Gateway/OpenClaw standalone-LLM paths create `LLMRunner` instances with configured timeouts (`cfg.llm.timeoutMs`). L1 extraction and L1 dedup overrode those runner defaults with `180_000`, making shorter or longer user-configured LLM timeouts ineffective for those calls.

## Tests
- `npm test -- src/core/record/l1-extractor.test.ts src/core/record/l1-dedup.test.ts`
- `npm test`
- `npm run build`

PR #279, opened 2026-06-29
fix(pipeline): honor L2/L3 runner timeout config

YOMXXX:
## Summary
- stop L2 scene extraction from overriding injected LLM runner timeout with a hard-coded 300s default
- stop L3 persona generation from overriding injected LLM runner timeout with a hard-coded 180s default
- preserve the legacy CleanContextRunner fallback defaults when no runner is injected
- add regression tests for both L2 and L3 runner timeout forwarding

## Tests
- npm test -- src/core/scene/scene-extractor.test.ts src/core/persona/persona-generator.test.ts
- npm test
- npm run build

### DIFF
--- a/src/core/persona/persona-generator.test.ts
+++ b/src/core/persona/persona-generator.test.ts
@@ -1,0 +1,44 @@
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+import { PersonaGenerator } from "./persona-generator.js";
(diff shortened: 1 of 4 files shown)


Maxwell-Code07:

YOMXXX:
Ready triage note: I checked this with the timeout/config group.

This PR has independent scope: it stops L2 scene extraction and L3 persona generation from overriding injected runner timeouts. It is complementary to #276 and #277, which cover local L2 and L1 extraction/dedup respectively.

I did not find duplicate scope; this should be reviewed as part of the same timeout-consistency cleanup set.

PR #272, opened 2026-06-27
fix(search): honor recall embedding timeout

YOMXXX:
## Summary
- pass `embedding.recallTimeoutMs ?? embedding.timeoutMs` from `TdaiCore` into both agent search tools
- forward the timeout to query embedding calls in `tdai_memory_search` and `tdai_conversation_search`
- add regression coverage for recall override and timeout fallback

## Why
`recallTimeoutMs` is already documented as the user-facing recall-path embedding timeout and auto-recall honors it, but the agent-callable search tools still called `embed(query)` without per-call options. A slow embedding provider could therefore ignore the shorter recall timeout during interactive tool search.

## Tests
- `npm test -- src/core/tdai-core.test.ts`
- `npm test`
- `npm run build`

PR #280, opened 2026-06-29
fix(core): honor layer model config for runners

YOMXXX:
## Summary
- pass extraction.model into standalone/host-neutral L1 runner creation
- pass persona.model into standalone/host-neutral L2/L3 runner creation
- apply the same mapping in seed runtime so seeded pipelines respect layer-specific model config
- add a TdaiCore regression test that captures runner factory options

## Tests
- npm test -- src/core/tdai-core.test.ts
- npm test
- npm run build

### DIFF
--- a/src/core/seed/seed-runtime.ts
+++ b/src/core/seed/seed-runtime.ts
@@ -91,8 +91,14 @@ async function createSeedPipeline(opts: SeedRuntimeOptions): Promise<{ pipeline:
       },
       logger,
     });
-    l1LlmRunner = runnerFactory.createRunner({ enableTools: false });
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

YOMXXX:
Merge-order triage note: I checked this with the runner config group.

This PR is independent from the timeout fixes: it passes layer-specific model config into standalone/host-neutral runner creation and seed runtime.

Merge-order note: #280 and #272 both add coverage in `src/core/tdai-core.test.ts`, so whichever lands second will need a small test-file rebase. That is a mechanical test conflict, not duplicate functionality.

PR #283, opened 2026-06-30
fix(pipeline): drain shutdown transitions through L3

YOMXXX:
## Summary
- keep the pipeline in a closing state during destroy so shutdown flush can still advance in-flight L1 output into L2/L3
- stop new public work while closing, and avoid arming future L2 max-interval timers during shutdown
- wait for L2 to drain before waiting for L3, because L2 completion can enqueue L3 work

## Tests
- npm test
- npm run build

### DIFF
--- a/src/utils/pipeline-manager.test.ts
+++ b/src/utils/pipeline-manager.test.ts
@@ -1,0 +1,103 @@
+import { describe, expect, it } from "vitest";
+import { MemoryPipelineManager } from "./pipeline-manager.js";
+import type { CapturedMessage, PipelineConfig } from "./pipeline-manager.js";
+
+function makePipelineConfig(): PipelineConfig {
+  return {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

YOMXXX:
Ready triage note: I checked this PR separately because it is a pipeline lifecycle fix rather than part of a larger overlap group.

Assessment:
- Scope is independent: keep shutdown in a closing state long enough for L1 output to drain through L2/L3, while rejecting new public work.
- I did not find duplicate scope with the other ready PRs in my queue.

Focused verification passed locally:
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/utils/pipeline-manager.test.ts` — 1 file, 2 tests passed
- `git diff --check upstream/main...HEAD` — passed

PR #292, opened 2026-06-30
fix(checkpoint): recalibrate aggregate counters on startup to fix drift (#157)

SurvivorNo1:
## Summary

Fix the counter drift described in #157. `total_memories_extracted` and `l0_conversations_count` in `recall_checkpoint.json` previously only ever incremented inside `CheckpointManager`, so external cleanup (memory-cleaner, manual JSONL pruning, deletion of test pipeline_states) left them permanently overstated. The drift corrupted persona-generation thresholds (`cp.total_processed`, `cp.memories_since_last_persona`) and status reporting.

## Changes

- **`CheckpointManager.recalibrate()`** (`src/utils/checkpoint.ts`)
  - New method that overwrites the two aggregate counters with authoritative counts.
  - Atomic: reuses the existing per-file lock + atomic tmp+rename write.
  - Returns a `RecalibrationReport` describing before/after values and whether each counter changed.
  - Accepts optional `{ l1Memories?, l0Conversations? }` so callers can recalibrate whichever counts they can obtain (VectorStore counts may not always be available).

- **`TdaiGateway.start()`** (`src/gateway/server.ts`)
  - After `core.initialize()`, runs recalibrate once using `vectorStore.countL1()` and `vectorStore.countL0()` before serving traffic.
  - Non-fatal: any failure is logged and the gateway still starts — drift will self-heal on the next startup.

- **`mutate()` generalized** to forward the callback's return value (used by `recalibrate()` to return its report). All existing callers still work; `incrementScenesProcessed()` updated to capture the new value inside the locked section.

- **Tests** (`src/utils/checkpoint.test.ts`, 10 new tests, all passing):
  - Counter drift reproduction
  - Overwrite lower (cleanup) and higher (backup restore) values
  - Partial recalibration (only one counter)
  - Empty input no-op
  - Fresh-checkpoint creation when none exists
  - Persistence across `new CheckpointManager()` instances
  - Preservation of unrelated fields (persona state, per-session state)
  - Input validation (rejects negative / non-finite)
  - Atomicity under concurrent `mutate()` calls

Closes #157

### DIFF
--- a/src/gateway/server.ts
+++ b/src/gateway/server.ts
@@ -23,6 +23,7 @@ import { loadGatewayConfig } from "./config.js";
 import type { GatewayConfig } from "./config.js";
 import { initDataDirectories } from "../utils/pipeline-factory.js";
 import { SessionFilter } from "../utils/session-filter.js";
+import { CheckpointManager } from "../utils/checkpoint.js";
 import type {
   HealthResponse,
   RecallRequest,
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:
ok

PR #178, opened 2026-06-10
fix(search): sanitize fts query operators

RerankerGuo:
## Description | 描述
  Strip FTS5 boolean/proximity operators from raw search text before tokenization so user input cannot carry query syntax into the
  generated MATCH expression.

  ## Related Issue | 关联 Issue
  Fix #160

  ## Change Type | 修改类型
  - [x] Bug fix | Bug 修复
  - [ ] New feature | 新功能
  - [ ] Documentation update | 文档更新
  - [ ] Code optimization | 代码优化

  ## Self-test Checklist | 自测清单
  - [x] Verified locally | 本地验证通过
  - [x] No existing features affected | 无影响现有功能

  ## Additional Notes | 其他说明
  Verified with `npm test` and `npm run build` using Node v24.15.0.

PR #309, opened 2026-06-30
fix(store): strip FTS5 special chars from user query tokens

chiluncc:
## Description | 描述

This PR sanitizes user query tokens before constructing SQLite FTS5 queries.

- Strip FTS5 special characters (`*`, `"`, `'`, `(`, `)`) from query tokens before wrapping them in double quotes.
- Preserve existing query behavior by continuing to join sanitized tokens with `OR`.
- Add test cases covering:
  - normal query construction
  - FTS5 special characters
  - empty/pure-special-character input
  - keyword operators (`AND`, `OR`, `NOT`, `NEAR`) treated as literal terms

## Related Issue | 关联 Issue

Fix #<issue-number>

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

None.

### DIFF
--- a/src/core/store/sqlite.test.ts
+++ b/src/core/store/sqlite.test.ts
@@ -1,0 +1,91 @@
+import { describe, expect, it } from "vitest";
+import { buildFtsQuery } from "./sqlite.js";
+
+describe("buildFtsQuery", () => {
+  // ── Normal behavior (no regression) ──
+
+  it("splits English words and OR-joins them as quoted phrases", () => {
+    expect(buildFtsQuery("hello world")).toBe('"hello" OR "world"');
+  });
+
+  it("handles Unicode (Chinese) text via the fallback regex", () => {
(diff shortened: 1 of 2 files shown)


YOMXXX:
Triage note: this is related to, but not the same as, #178.

#178 removes FTS5 boolean/proximity operator words (`AND`, `OR`, `NOT`, `NEAR`) before tokenization. This PR instead strips FTS5 special characters such as `*`, quotes, and parentheses from already-tokenized terms. That makes it a useful follow-up safety hardening rather than a duplicate.

However, both PRs touch `src/core/store/sqlite.ts` and add `src/core/store/sqlite.test.ts` from the same base, so this will likely need to be rebased after #178 and its tests folded around the canonical operator-sanitization coverage. I would not merge this before #178 lands.

PR #331, opened 2026-07-01
fix(record): retry invalid l1 extraction output and trim pipeline debug logs

AustinZ29671:
Summary:
- retry L1 extraction once when the first response is empty or non-JSON
- add a regression test covering natural-language and empty-string failures
- replace oversized pipeline state JSON debug logs with a compact summary

Why:
- in real Hermes standalone use with low-cost extraction models, L1 sometimes returns empty text or free-form analysis instead of the required JSON array
- the existing parser then drops the extraction entirely even though a second stricter pass is often enough to recover
- pipeline debug logs can also grow excessively because they dump full persisted state JSON on every write

Validation:
- npm test -- --run src/core/record/l1-extractor.test.ts src/utils/no-think-fetch.test.ts
- npm test

### DIFF
--- a/src/core/record/l1-extractor.test.ts
+++ b/src/core/record/l1-extractor.test.ts
@@ -1,0 +1,130 @@
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+import { extractL1Memories } from "./l1-extractor.js";
+import type { ConversationMessage } from "../conversation/l0-recorder.js";
+import type { LLMRunner, LLMRunParams, Logger } from "../types.js";
+
+class SequenceRunner implements LLMRunner {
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

YOMXXX:
Triage note: the L1 parse/retry parts of this PR overlap with already selected narrower fixes.

Relevant canonical PRs:
- #181: robust L1 parsing and capture-path embedding timeout propagation for #110
- #190: retry unparsable L1 extraction output exactly once for #115
- #430: treat empty/non-JSON/provider `content_filter` L1 extraction output as failed and preserve the cursor for retry in #104

This PR also includes pipeline debug log trimming, which may still be useful, but it should be split into a small follow-up. Recommendation: re-scope to only the compact pipeline log summary if that behavior is still needed, and drop the L1 retry/parse changes that are already covered by #181/#190/#430.

PR #343, opened 2026-07-01
feat(utils): add optional cache-shape diagnostics (PrefixShape)

gugu23456789:
## Related Issue
fix #120

## Description

Optional diagnostic utility for reasoning about prompt-cache hit/miss across turns. Inspired by Reasonix `PrefixShape` (internal/agent/cache_shape.go, v1.13.1).

```typescript
const shape = captureShape(systemPrompt, toolSchemas);
const diag = compareShape(prevShape, shape);
if (diag.prefixChanged) {
  logger?.warn("cache miss:", diag.prefixChangeReasons.join(", "));
}
```

**Non-invasive**: zero runtime side effects, no integration required. Platforms import only if they want diagnostics; tree-shaking elides it otherwise.

**Files:**
- `src/utils/cache-diagnostics.ts` PrefixShape + captureShape/compareShape
- `src/utils/cache-diagnostics.test.ts` 8 tests

### DIFF
--- a/src/utils/cache-diagnostics.test.ts
+++ b/src/utils/cache-diagnostics.test.ts
@@ -1,0 +1,58 @@
+import { describe, it, expect } from "vitest";
+import { captureShape, compareShape, type PrefixShape } from "./cache-diagnostics";
+
+describe("captureShape", () => {
+  it("produces deterministic hashes for identical inputs", () => {
+    const a = captureShape("you are a helpful assistant", [{ name: "read_file" }]);
+    const b = captureShape("you are a helpful assistant", [{ name: "read_file" }]);
+    expect(a.systemHash).toBe(b.systemHash);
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

YOMXXX:
Duplicate cleanup / re-scope note for #120:

The canonical scoped runtime mitigation for #120 is now #375. It adds an explicit `recall.injectionMode` opt-in path, keeps backward-compatible behavior by default, preserves safe injected-history cleanup via `recall.showInjected=false`, and has targeted tests plus full local verification.

This PR is diagnostic-only and intentionally has no runtime effect, so it should not be treated as the fix for #120. If maintainers still want it, I recommend re-scoping it as an optional follow-up utility after #375 rather than closing #120 from this PR.

Maxwell-Code07:

PR #347, opened 2026-07-01
fix(offload): invalidate token cache after tool use stripping

### LINKED ISSUE #233: [Bug] Stale per-message token cache after in-place tool_use strip over-deletes messages

RerankerGuo:
## Description | 描述

Fix the stale token-cache path after deleted `tool_use` blocks are stripped out
of mixed assistant messages.

`buildTiktokenContextSnapshot()` caches token counts by message object identity
and `_offloaded` state. The offload fast paths mutate mixed assistant messages
in place with `content.splice(...)`, but the old code did not invalidate that
per-message cache. A later token snapshot could therefore keep counting the
removed `tool_use` payload and trigger unnecessary guard/emergency compression.

This PR keeps the runtime change narrow:
- adds `stripDeletedToolUseBlocks()` as the shared helper for deleted
  `tool_use` removal,
- invalidates the token cache only when a block is actually removed,
- reuses the helper in the before-prompt-build fast path, the llm-input L3 fast
  path, and the context-engine assemble fast path,
- adds a Vitest regression that seeds the cache, strips a deleted block, and
  verifies the next snapshot reflects the smaller message.

## Related Issue | 关联 Issue

Fix #233

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证

```bash
PATH=$HOME/.nvm/versions/node/v24.15.0/bin:$PATH npm_config_cache=/tmp/npm-cache-tam-233 npx vitest run src/offload/l3-helpers.test.ts
PATH=$HOME/.nvm/versions/node/v24.15.0/bin:$PATH npm_config_cache=/tmp/npm-cache-tam-233 npm test
PATH=$HOME/.nvm/versions/node/v24.15.0/bin:$PATH npm_config_cache=/tmp/npm-cache-tam-233 npm run build
git diff --check
```

Results:
- `src/offload/l3-helpers.test.ts`: 2 tests passed.
- `npm test`: 5 files / 69 tests passed.
- `npm run build`: plugin and scripts build passed.

### DIFF
--- a/src/offload/hooks/before-prompt-build.ts
+++ b/src/offload/hooks/before-prompt-build.ts
@@ -24,6 +24,7 @@ import {
   replaceWithSummary,
   replaceAssistantToolUseWithSummary,
   compressNonCurrentToolUseBlocks,
+  stripDeletedToolUseBlocks,
   getCurrentTaskNodeIds,
 } from "../l3-helpers.js";
 import {
@@ -99,18 +100,7 @@ export function createBeforePromptBuildHandler(
         // FIX: For mixed assistant messages (text + tool_use), strip deleted tool_use
         // blocks to prevent orphaned tool_use without matching tool_result (Anthropic 400).
         if (hasDeleted && isAssistantMessageWithToolUse(msg) && !isOnlyToolUseAssistant(msg)) {
-          const content = msg.type === "message" ? msg.message?.content : msg.content;
-          if (Array.isArray(content)) {
-            for (let j = content.length - 1; j >= 0; j--) {
-              const block = content[j] as any;
-              if ((block.type === "tool_use" || block.type === "toolCall") && block.id) {
-                const blockIdNorm = normalizeToolCallIdForLookup(block.id);
-                if (stateManager.deletedOffloadIds.has(block.id) || stateManager.deletedOffloadIds.has(blockIdNorm)) {
-                  content.splice(j, 1);
-                }
-              }
-            }
-          }
+          stripDeletedToolUseBlocks(msg, stateManager.deletedOffloadIds);
         }
         if (msg._offloaded) continue;
         if (tid && hasConfirmed && (stateManager.confirmedOffloadIds.has(tid) || (tidNorm && stateManager.confirmedOffloadIds.has(tidNorm)))) {
--- a/src/offload/hooks/llm-input-l3.ts
+++ b/src/offload/hooks/llm-input-l3.ts
@@ -23,6 +23,7 @@ import {
   replaceWithSummary,
   replaceAssistantToolUseWithSummary,
   compressNonCurrentToolUseBlocks,
+  stripDeletedToolUseBlocks,
   getCurrentTaskNodeIds,
 } from "../l3-helpers.js";
 import type { OffloadStateManager } from "../state-manager.js";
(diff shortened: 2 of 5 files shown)


Maxwell-Code07:

YOMXXX (approved):
Reviewed and verified locally. This is the stronger #233 fix because it centralizes deleted `tool_use` stripping in `stripDeletedToolUseBlocks()`, invalidates the token cache only when a block is actually removed, covers all three call sites including `src/offload/index.ts`, and adds a regression that proves the post-strip token snapshot shrinks instead of reusing the stale cached count.

Verification I ran:
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/offload/l3-helpers.test.ts` — 2/2 passed
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm test` — 5 files / 69 tests passed
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm build` — passed
- `git diff --check` — clean

PR #366, opened 2026-07-02
fix(search): build FTS5 queries from quoted literals

WHUTcjh-2024:
## Description | 描述

Fix FTS5 query construction so tokenizer output is always interpreted as literal search terms rather than SQLite FTS5 syntax.

### Root cause

`buildFtsQuery()` previously concatenated tokenizer output with `OR` without encoding individual terms. Tokens such as `AND`, `OR`, `NOT`, `NEAR`, quotes, parentheses, column filters, or wildcard characters could therefore alter the `MATCH` expression or produce an FTS5 syntax error.

The fallback and jieba paths also applied different normalization rules, which made malformed-input behavior harder to reason about.

### Changes

- Add a shared literal-only FTS5 query encoder.
- Extract Unicode letter, number, and underscore token parts from every tokenizer candidate.
- Quote every term with SQLite FTS5-compatible double-quote escaping.
- Preserve reserved words as searchable literals instead of silently dropping them.
- Deduplicate terms case-insensitively while preserving the first spelling.
- Use the same encoder for both jieba and fallback tokenization paths.
- Return `null` when no searchable literal remains.

### Compatibility

Benign fallback queries remain byte-for-byte compatible. Punctuation-heavy technical terms follow the same `unicode61` token boundaries used by the existing FTS5 indexes; this is covered for `C++`, `.NET`, `node.js`, and email-like text.

No schema, migration, ranking, or public API changes are introduced.

## Related Issue | 关联 Issue

Fix #160

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

### Validation

- `npm test`: 5 test files, 93 tests passed.
- Focused FTS5 suite: 26 tests passed.
- 2,000 deterministic fuzz payloads executed against real SQLite FTS5 without syntax errors.
- Real `VectorStore` L0 and L1 search paths verified.
- `fts5-query.ts`: 100% statements, branches, functions, and lines.
- `npm run build`: passed.
- `npm pack --dry-run --json`: passed.
- `git diff --check`: passed.

### DIFF
--- a/src/core/store/fts5-query.test.ts
+++ b/src/core/store/fts5-query.test.ts
@@ -1,0 +1,270 @@
+import { DatabaseSync } from "node:sqlite";
+import { mkdtemp, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+
+import type { MemoryRecord } from "../record/l1-writer.js";
+import type { L0Record } from "./types.js";
+import {
+  buildFts5LiteralOrQuery,
+  quoteFts5Literal,
+} from "./fts5-query.js";
+import {
+  _resetJiebaForTest,
+  _setJiebaForTest,
+  buildFtsQuery,
+  VectorStore,
+} from "./sqlite.js";
+
+describe("quoteFts5Literal", () => {
+  it("escapes quotes according to the FTS5 string grammar", () => {
+    expect(quoteFts5Literal('alpha"beta')).toBe('"alpha""beta"');
+  });
+});
+
+describe("buildFts5LiteralOrQuery", () => {
+  it.each([
+    ["travel plan API", '"travel" OR "plan" OR "API"'],
+    ['alpha" OR "beta', '"alpha" OR "OR" OR "beta"'],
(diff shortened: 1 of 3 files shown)


WHUTcjh-2024:
Hi maintainers, this PR fixes #160 by making every FTS5 token a quoted literal instead of stripping reserved words. Compared with a simple operator-removal approach, this preserves searchable inputs like "AND", "OR", ".NET", "C++", "node.js", and email-like text while avoiding FTS5 syntax errors.

I have added focused FTS5 tests and fuzz cases against real SQLite FTS5. If the preferred direction is a smaller patch, I can simplify the implementation.

YOMXXX:
Triage note: #178 has been locally verified and approved as the canonical narrow fix for #160. This PR appears to cover the same FTS5 operator-sanitization scope, so I recommend treating it as superseded unless it intentionally adds coverage or behavior beyond #178.

WHUTcjh-2024:
Thanks for the triage note. I agree #178 is the canonical narrow fix for #160, so I am closing this PR as superseded rather than competing with the already approved fix. If maintainers later want additional FTS5 regression coverage or literal-query behavior after #178 lands, I can open a smaller follow-up PR.

PR #368, opened 2026-07-02
fix(profile): sanitize pulled scene filenames

YOMXXX:
## Summary
- normalize remote L2 profile filenames before writing pulled scene blocks
- keep pulled scene content inside `scene_blocks/` even when remote filenames contain path segments
- add a regression test for `../../escaped.md` profile filenames

## Why
`pullProfilesToLocal()` reads profile records from the remote store and previously joined `record.filename` directly into the temporary `scene_blocks` path. A malformed or stale remote profile filename could escape the scene block directory during pull.

## Tests
- `npm test -- src/core/profile/profile-sync.test.ts`
- `npm test`
- `npm run build`
- `git diff --check`

### DIFF
--- a/src/core/profile/profile-sync.test.ts
+++ b/src/core/profile/profile-sync.test.ts
@@ -1,0 +1,64 @@
+import { createHash } from "node:crypto";
+import { access, mkdtemp, readFile, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
+
+import { afterEach, describe, expect, it } from "vitest";
+
+import { pullProfilesToLocal } from "./profile-sync.js";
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Thanks for your attention! We’ll review this and get back to you as soon as possible.

YOMXXX:
Ready triage note: I checked this with the path-boundary group.

This PR is not a duplicate of #369 or the offload path-boundary fixes. #368 covers remote profile filenames before writing pulled scene blocks into `scene_blocks/`; #369 covers filenames read from `.metadata/scene_index.json`. These are adjacent but distinct read/write boundaries.

I did not find merge-order conflicts for this PR in the path-boundary group.

PR #369, opened 2026-07-02
fix(scene): normalize indexed filenames on read

YOMXXX:
## Summary
- normalize filenames loaded from `.metadata/scene_index.json`
- prevent stale or malformed scene metadata from feeding path segments to downstream readers/navigation
- add a regression test for `../../secret.md` metadata filenames

## Why
`scene_index.json` is normally generated by `syncSceneIndex()`, but stale/corrupt metadata can be consumed by persona generation and recall navigation. Normalizing on read keeps downstream code operating on safe single-segment scene filenames.

## Tests
- `npm test -- src/core/scene/scene-index.test.ts`
- `npm test`
- `npm run build`
- `git diff --check`

### DIFF
--- a/src/core/scene/scene-index.test.ts
+++ b/src/core/scene/scene-index.test.ts
@@ -1,0 +1,45 @@
+import { mkdtemp, mkdir, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
+
+import { afterEach, describe, expect, it } from "vitest";
+
+import { readSceneIndex } from "./scene-index.js";
+
+const tempDirs: string[] = [];
+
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Thanks for your attention! We’ll review this and get back to you as soon as possible.

YOMXXX:
Ready triage note: I checked this with the path-boundary group.

This PR is not a duplicate of #368 or the offload path-boundary fixes. #369 covers filenames loaded from `.metadata/scene_index.json`; #368 covers remote profile filenames before writing pulled scene blocks. These are adjacent but distinct read/write boundaries.

I did not find merge-order conflicts for this PR in the path-boundary group.

PR #413, opened 2026-07-07
fix(checkpoint): recalibrate counters on startup to fix drift after cleanup

Lfan-ke:
Closes #157

Memory-cleaner deletes records from JSONL/SQLite but never decrements the checkpoint counters (`total_memories_extracted`, `l0_conversations_count`, `memories_since_last_persona`), causing counter drift over time that accumulates silently after every cleanup pass.

## Root cause

`CheckpointManager` has no way to reconcile its counters against actual store state. The `LocalMemoryCleaner` deletes L0/L1 records but leaves counters inflated, and the scheduler startup reads these stale counters without correction.

## Fix

Adds `CheckpointManager.recalibrate(store?)` that:
- Reads actual L1/L0 counts concurrently via `Promise.all([store.countL1(), store.countL0()])`
- Adjusts `total_memories_extracted` and `l0_conversations_count` to match actual store state
- Decrements `memories_since_last_persona` by the L1 delta (so persona-update decisions stay accurate)
- Clamps all values to `>= 0` — no counter goes negative
- Is a no-op when `store` is unavailable; skips silently on transient store errors (non-fatal)

Wired in two places:
1. **`pipeline-factory.ts`** — called at gateway startup before reading checkpoint state
2. **`memory-cleaner.ts`** — called after each cleanup pass when records were actually removed

## Tests (10 cases)

- No-op without store, no file created
- Writes correct L1/L0 counts into checkpoint
- Corrects inflated counters (drift simulation)
- Handles zero counts (empty store)
- Returns `undefined` and skips write on store error (no partial write)
- Preserves unrelated checkpoint fields
- Decrements `memories_since_last_persona` by removed-L1 delta
- Floors `memories_since_last_persona` at 0 when removal exceeds current value
- Does not change `memories_since_last_persona` when L1 count increases

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -31,6 +31,7 @@ import {
 } from "./src/utils/clean-context-runner.js";
 import { SessionFilter } from "./src/utils/session-filter.js";
 import { LocalMemoryCleaner } from "./src/utils/memory-cleaner.js";
+import { CheckpointManager } from "./src/utils/checkpoint.js";
 import { registerMemoryTdaiCli } from "./src/cli/index.js";
 import { initDataDirectories, resetStores } from "./src/utils/pipeline-factory.js";
 import { getOrCreateInstanceId, initReporter, report, resetReporter } from "./src/core/report/reporter.js";
@@ -305,6 +306,7 @@ export default function register(api: OpenClawPluginApi) {
         retentionDays: cfg.memoryCleanup.retentionDays,
         cleanTime: cfg.memoryCleanup.cleanTime,
         logger: api.logger,
+        checkpoint: new CheckpointManager(pluginDataDir, api.logger),
       });
       sharedMemoryCleaner.start();
       api.logger.debug?.(`${TAG} Memory cleaner started (singleton)`);
--- a/src/utils/checkpoint.test.ts
+++ b/src/utils/checkpoint.test.ts
@@ -1,0 +1,155 @@
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+
+import { CheckpointManager } from "./checkpoint.js";
+import type { IMemoryStore } from "../core/store/types.js";
+
+function makeDir(): Promise<string> {
+  return fs.mkdtemp(path.join(os.tmpdir(), "cp-test-"));
+}
+
+function mockStore(l1: number, l0: number): Pick<IMemoryStore, "countL1" | "countL0"> & IMemoryStore {
+  return {
+    countL1: async () => l1,
+    countL0: async () => l0,
+  } as unknown as IMemoryStore;
+}
+
+describe("CheckpointManager.recalibrate", () => {
+  let dir: string;
+
+  beforeEach(async () => {
+    dir = await makeDir();
+    await fs.mkdir(path.join(dir, ".metadata"), { recursive: true });
+  });
+
+  afterEach(async () => {
+    await fs.rm(dir, { recursive: true, force: true });
+  });
+
+  it("is a no-op when store is not provided", async () => {
+    const cm = new CheckpointManager(dir);
+    await cm.recalibrate();
+    // No checkpoint file should be created
+    const filePath = path.join(dir, ".metadata", "recall_checkpoint.json");
+    await expect(fs.access(filePath)).rejects.toThrow();
+  });
+
+  it("writes store counts into the checkpoint", async () => {
+    const cm = new CheckpointManager(dir);
+    const store = mockStore(42, 7);
+    const result = await cm.recalibrate(store);
+
+    expect(result).toEqual({ total_memories_extracted: 42, l0_conversations_count: 7 });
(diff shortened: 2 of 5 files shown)


Maxwell-Code07:

Lfan-ke:
@Yuntong8888 @jackson-jia-914 @zhangxiaoshuai98 Could you take a look at this PR when you have time?

jackson-jia-914:
One concern I'd like to raise:

  Missing error handling on countL1() / countL0() could block gateway startup

  countL1() and countL0() can throw on transient SQLite/VectorDB failures. memory-cleaner.ts (lines 116-117) already wraps these same calls in try/catch as non-fatal.
  However, recalibrate() doesn't catch them — if either throws, createPipeline() will reject and block gateway startup, which seems too strict for a status-counter hygiene
   fix.

  Suggest wrapping the count reads:

  let actualL1: number, actualL0: number;
  try {
    [actualL1, actualL0] = await Promise.all([store.countL1(), store.countL0()]);
  } catch (err) {
    this.logger.info(
      `[checkpoint] recalibrate skipped: ${err instanceof Error ? err.message : err}`,
    );
    return undefined;
  }

  This way a transient store issue just skips recalibration (counters stay as-is, fixed on next startup) rather than taking the whole pipeline down. A corresponding test
  case for the rejection path would also be helpful.

  Happy to take a look again once this is addressed.

Lfan-ke:
Fixed — wrapped the `Promise.all([store.countL1(), store.countL0()])` call in a try/catch; on any transient store error, `recalibrate()` logs and returns `undefined` so the pipeline starts normally with counters as-is. Added a test case that seeds known counters (10/5), feeds a throwing store, asserts `undefined` return, and verifies the counters are unchanged (no partial write). All 7 tests pass.

jackson-jia-914:
Thanks for the fix — looks good. The error handling on countL1() / countL0() is now in place, and the new test (returns undefined and skips write when store throws)
nicely covers the no-partial-write invariant by seeding known counter values and asserting they're preserved when the store throws.

LGTM. 🚢

YOMXXX:
Triage note: #177 has been locally verified and approved as the canonical #157 fix. It recalibrates checkpoint counters against actual store counts, adjusts memories_since_last_persona safely, and wires recalibration after cleanup plus scheduler startup. This PR appears to cover the same checkpoint-counter drift scope, so I recommend treating it as superseded unless it intentionally adds behavior beyond #177.

Lfan-ke:
Thanks for the triage. Here is a concrete diff between the two approaches:

**API design**
- #177 `recalibrateCounts({ l0ConversationsCount, totalMemoriesExtracted })` — takes pre-read counts as plain values; caller owns the store queries
- #413 `recalibrate(store?)` — accepts an `IMemoryStore` and reads `countL1`/`countL0` concurrently inside via `Promise.all`; optional store makes it safe to call unconditionally at startup

**Counter coverage**
- #177 also decrements `memories_since_last_persona` by the delta of removed L1 records — this is the right thing to do and a genuine gap in #413
- #413 only recalibrates `total_memories_extracted` and `l0_conversations_count`

**Integration points**
- #177 wires into `memory-cleaner.ts` (after every cleanup pass) and `tdai-core.ts` (scheduler startup)
- #413 wires into `pipeline-factory.ts` (gateway startup)

Bottom line: #177's `memories_since_last_persona` adjustment is a real improvement our PR lacks. Happy to defer to maintainers — if you'd like to close #413 as superseded, that is a reasonable call.

Lfan-ke:
Update after latest push: the gap I acknowledged earlier has been addressed.

**`memories_since_last_persona` adjustment is now included** — `recalibrate()` computes the L1 delta (`prev - actual`) and decrements `memories_since_last_persona` by that amount, clamped to `>= 0`. This matches the intent of #177's adjustment.

**`memory-cleaner.ts` integration added** — `LocalMemoryCleaner` now accepts an optional `checkpoint` in its constructor options and calls `recalibrate(vectorStore)` after each cleanup pass where records were actually removed.

#413 is now a strict superset of #177:

| Feature | #177 | #413 |
|---|---|---|
| `total_memories_extracted` correction | ✓ | ✓ |
| `l0_conversations_count` correction | ✓ | ✓ |
| `memories_since_last_persona` delta | ✓ | ✓ |
| Floor counters at 0 | ✓ | ✓ |
| Called after cleanup pass | ✓ | ✓ |
| Called at scheduler/gateway startup | ✓ | ✓ |
| Store read inside the method (simpler callers) | — | ✓ |
| Transient store error is non-fatal (returns `undefined`) | — | ✓ |
| No-op without store (safe unconditional call) | — | ✓ |
| Test: `memories_since_last_persona` decremented correctly | — | ✓ |
| Test: `memories_since_last_persona` floored at 0 | ✓ | ✓ |
| Test: no change when L1 increases | — | ✓ |
| Test: no partial write on store error | — | ✓ |

Lfan-ke:
Closing - the task is handled and there are already many equivalent PRs, no reason to keep this open.

Maxwell-Code07:

PR #418, opened 2026-07-07
fix(hermes): mirror explicit memory writes (#417)

### LINKED ISSUE #417: [Bug] Hermes memory_tencentdb does not mirror explicit memory() writes into TencentDB memory search
## Bug Description
When Hermes is configured with `memory.provider: memory_tencentdb`, explicit durable-memory writes made through the built-in Hermes `memory` tool are accepted locally but do not become searchable through TencentDB memory search.

In practice:
- `memory(action="add", ...)` succeeds
- `memory_tencentdb_conversation_search` works
- `memory_tencentdb_memory_search` does **not** return the freshly written marker

This makes explicit "save memory" behavior diverge from TencentDB L1 memory search.

## Root Cause
The Hermes-side provider hook appears to be a no-op.

In the installed provider code (`hermes-plugin/memory/memory_tencentdb/__init__.py` in the TencentDB plugin tree), `on_memory_write(...)` is stubbed:

```python
    pass
```

Hermes core does call this hook after built-in memory writes, but the provider currently does nothing with it.

## Reproduction
1. Configure Hermes with:
   - `memory.memory_enabled: true`
   - `memory.user_profile_enabled: true`
   - `memory.provider: memory_tencentdb`
2. Call the built-in Hermes memory tool with a unique marker, e.g.:
   - `memory(action="add", target="memory", content="TencentDB retry probe memory: durable memory round-trip verification marker.")`
3. Query TencentDB memory search with that marker / substring.
4. Query TencentDB conversation search with the same marker.

## Expected Behavior
The explicit memory write should be mirrored into TencentDB's memory index so that `memory_tencentdb_memory_search` can retrieve it.

## Actual Behavior
Only conversation search works. Explicit Hermes memory writes are not ingested into TencentDB memory search.

## Environment
- Host OS: Linux
- Hermes repo observed locally: `NousResearch/hermes-agent`
- TencentDB memory repo: `TencentCloud/TencentDB-Agent-Memory`
- GitHub auth verified via `gh`

## Notes
This does not look like a pure search-quality problem. The provider hook for explicit memory writes appears unimplemented, so the written memory never reaches TencentDB's searchable memory store.

A clean fix would likely require implementing an explicit-memory ingest path and wiring `on_memory_write(...)` to it, rather than relying only on conversation capture.

kumar-shivang:
*Description | 描述*
This PR fixes Hermes explicit memory writes not being mirrored into TencentDB memory search.

Before this change:
- Hermes `memory(action="add", ...)` writes were stored in Hermes built-in memory
- the `memory_tencentdb` provider received the callback
- but `on_memory_write(...)` in the Hermes plugin was effectively a no-op
- result: explicit memories could not be retrieved through TencentDB memory search, even though conversation search still worked

This PR adds a dedicated explicit-memory ingest path instead of faking conversation capture:
- implements `on_memory_write(...)` in the Hermes TencentDB plugin
- adds client support for forwarding explicit memory writes to the gateway
- adds a new explicit-memory endpoint in the gateway/core path
- writes explicit Hermes memories directly into TencentDB L1 memory
- maps Hermes targets explicitly:
  - `target="memory"` -> `instruction`
  - `target="user"` -> `persona`

This makes Hermes explicit memory round-trip work correctly with TencentDB memory search.

*Related Issue | 关联 Issue*
Fix #417

*Change Type | 修改类型*
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

*Self-test Checklist | 自测清单*

*Additional Notes | 其他说明*
Local verification completed with:
- `npm test -- src/core/record/explicit-memory.test.ts`
- `npm run build:plugin`
- provider Python tests passed
- real Hermes round-trip verified in a fresh process:
  explicit `memory add` -> `memory_tencentdb_memory_search` returned the inserted marker

### DIFF
--- a/hermes-plugin/memory/memory_tencentdb/__init__.py
+++ b/hermes-plugin/memory/memory_tencentdb/__init__.py
@@ -1048,9 +1048,27 @@ class MemoryTencentdbProvider(MemoryProvider):
 
     def on_memory_write(self, action: str, target: str, content: str) -> None:
         """Mirror built-in memory writes to memory-tencentdb for indexing."""
-        # TODO: Implement mirroring of Hermes builtin MEMORY.md/USER.md writes
-        # to memory-tencentdb's recall index for conflict suppression and dedup.
-        pass
+        if action.strip().lower() != "add":
+            return
+        if not content.strip():
+            return
+        if not self._ensure_alive_for_request() or not self._client:
+            return
+
+        try:
+            self._client.write_explicit_memory(
+                action=action,
+                target=target,
+                content=content,
+                session_key=self._session_id,
+                session_id=self._session_id,
+                user_id=self._user_id,
+            )
+            self._record_success()
+        except Exception as e:
+            self._record_failure()
+            logger.warning("memory-tencentdb explicit memory sync failed: %s", e)
+            self._try_recover_gateway()
 
     def on_session_end(self, messages: List[Dict[str, Any]]) -> None:
         """Trigger session-level flush on the Gateway."""
--- a/hermes-plugin/memory/memory_tencentdb/client.py
+++ b/hermes-plugin/memory/memory_tencentdb/client.py
@@ -139,6 +139,28 @@ class MemoryTencentdbSdkClient:
             body["user_id"] = user_id
         return self._post("/capture", body)
 
+    def write_explicit_memory(
+        self,
+        action: str,
+        target: str,
+        content: str,
+        session_key: str,
+        session_id: str = "",
+        user_id: str = "",
+    ) -> Dict[str, Any]:
+        """Mirror a host-native durable-memory write directly into L1."""
+        body: Dict[str, Any] = {
+            "action": action,
+            "target": target,
+            "content": content,
+            "session_key": session_key,
+        }
+        if session_id:
+            body["session_id"] = session_id
+        if user_id:
+            body["user_id"] = user_id
+        return self._post("/memories/explicit", body)
+
     def search_memories(self, query: str, limit: int = 5, type_filter: str = "", scene: str = "") -> Dict[str, Any]:
         """Search L1 structured memories."""
         body: Dict[str, Any] = {"query": query, "limit": limit}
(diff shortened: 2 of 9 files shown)


Maxwell-Code07:
Thank you for your attention. We will conduct a review and respond to you as soon as possible.

YOMXXX (changes requested):
Thanks for wiring the Hermes hook through to the Gateway. I think this still needs one correctness fix before merge.

The endpoint currently returns `stored: true` as soon as `writeMemory()` appends the JSONL record, but `memory_tencentdb_memory_search` only searches through `vectorStore`. If store init is degraded/unavailable, or if `vectorStore.upsertL1()` returns false/throws, this PR will still report success while the exact issue remains: the explicit memory is durable in JSONL but not searchable. The new test only covers the happy path where `upsertL1` succeeds, so it would not catch this.

Please make `/memories/explicit` only report success when the record is actually indexed for memory search, or return a degraded/failure response when indexing is unavailable, and add a regression test covering the indexing-failure/no-vector-store path plus the write-then-search behavior. Relevant code paths: `src/core/tdai-core.ts` `ingestExplicitMemory`, `src/core/record/l1-writer.ts` vector dual-write, and `src/core/tools/memory-search.ts` no-vector/no-FTS behavior.

YOMXXX:
I opened #434 as a replacement/supplemental branch for this fix. It keeps the Hermes hook and Gateway explicit-memory path from this PR, then adds the missing strict indexing semantics: explicit memory writes only report `stored:true` after `VectorStore.upsertL1()` succeeds, and the Hermes provider treats `stored:false` as a sync failure instead of silently recording success.

YOMXXX:
Marking this as superseded by #434 for #417.

#434 builds on this approach and adds the important guarantee that explicit Hermes memory writes must reach the searchable L1 index before the Gateway reports stored=true. I verified #434 locally with the focused explicit-memory test, full test suite, build, Python py_compile, and diff check.

PR #427, opened 2026-07-08
fix(otel): align OTLP backend Resource API with v2 SDK (#420)

### LINKED ISSUE #420: v1.0.0 tag: OTel SDK version mismatch — `new Resource()` not compatible with `resources@^2.x`
**Bug:** The v1.0.0 tag's `package.json` specifies `@opentelemetry/resources@^2.7.1` (v2.x SDK), but the codebase uses `new Resource({...})` which is a v1.x API (e.g. in `src/core/report/otlp-backend.ts`). The v2.x SDK fails with:

```
TypeError: Resource is not a constructor
```

This means OTel observability is silently broken on the v1.0.0 tag — any code path that initializes `Resource` via `new Resource()` will throw.

**Root cause:** The v1.0.0 tag was published with v2.x OTel SDK versions that don't match the v1.x API the code actually uses. A `try/catch` around OTel initialization masks the error (falls back to console logging), so the breakage is silent.

**Fix (verified on `gugu23456789/TencentDB-Agent-Memory` branch `r3-v1`):** Align to the v1.x line:

```
@opentelemetry/resources@^1.30.1
@opentelemetry/sdk-node@^0.54.0
```

These are peer-compatible with `@opentelemetry/api@^1.9.1` and work correctly with `new Resource()`.

**Full OTLP pipeline verified (on r3-v1 fork):**

```
trace.report() -> OTel SDK -> OTLP/HTTP -> Jaeger (:4318)
```

Setup guide at `docs/otlp-setup.md` on the r3-v1 branch.

**Affects:** Anyone using the v1.0.0 tag as a base for adapter development or production deployment with OTel observability enabled.

**Suggested fix:** Either (a) update `package.json` dependency versions to the v1.x line as above, or (b) update the code to use the v2.x API (`resourceFromAttributes()`).

RerankerGuo:
## Description

Fixes #420 — `new Resource({...})` in `src/core/report/otlp-backend.ts` is the
v1 OTel SDK API, but `package.json` locks `@opentelemetry/resources@^2.7.1`
which has dropped the `Resource` class entirely in favor of the
`resourceFromAttributes()` factory function. The `new Resource(...)` call
therefore throws `TypeError: Resource is not a constructor` at every Gateway
startup, and the surrounding `try/catch` in `initOTelSDK()` swallows it into a
`console.warn`, so OTLP observability has been silently broken on the v1.0.0
tag.

## Root cause

```ts
// before
const { Resource } = await import("@opentelemetry/resources");
...
resource: new Resource({ [ATTR_SERVICE_NAME]: serviceName }),
```

`@opentelemetry/resources@2.x` exports no `Resource` class:

```
typeof Resource               === "undefined"
typeof resourceFromAttributes === "function"
```

So `new Resource({...})` is a hard runtime error, masked by the outer catch
(`console.warn("... Trace/Log will use console fallback.")`).

## Fix

```ts
// after
const { resourceFromAttributes } = await import("@opentelemetry/resources");
...
resource: resourceFromAttributes({ [ATTR_SERVICE_NAME]: serviceName }),
```

This matches the migration pattern already used in
`src/core/report/otel-sdk-init.ts:120-127` (which keeps a runtime-detect
helper for backward compat). Here we drop the v1 fallback because
`package.json` is pinned to v2.x, so the v1 path is dead code anyway.

### Error handling upgrade

The catch block was rewritten to surface failures instead of silently
downgrading:

- Default: `console.error` with the error message **and stack trace**.
- `TDAI_OTEL_FAIL_OPEN=true` env knob restores the previous `console.warn`
  behavior for CI / debugging scenarios.

The function signature still returns `boolean` (callers don't change), but
operators now see the full failure mode in production logs instead of a
one-line warning that is easy to miss.

## Tests

Added `src/core/report/otlp-backend-resource.test.ts` with 4 cases:

1. Asserts v2.x `@opentelemetry/resources` does not export `Resource` and
   does export `resourceFromAttributes` — direct regression marker for the
   API surface change.
2. Asserts `resourceFromAttributes({...})` returns a `Resource` with the
   supplied attributes.
3. End-to-end: `OtlpObservabilityBackend.initialize()` against an
   unreachable endpoint must not throw `Resource is not a constructor`. The
   test captures console output and explicitly greps for the regression
   marker so a future downgrade to v1.x API would fail this case.
4. Lifecycle: `otel.enabled: false` is treated as a no-op.

Local verification:

```
$ PATH=~/.nvm/versions/node/v24.15.0/bin:$PATH npx vitest run \
    src/core/report/otlp-backend-resource.test.ts
 Test Files  1 passed (1)
      Tests  4 passed (4)

$ PATH=~/.nvm/versions/node/v24.15.0/bin:$PATH npm test
 Test Files  1 passed (1)
      Tests  4 passed (4)

$ PATH=~/.nvm/versions/node/v24.15.0/bin:$PATH npm run build
... clean

$ git diff --check
(empty)
```

## Relation to #416

The `L2_schedule` → `offload-l2` timer misrouting reported in #416 was
harder to diagnose precisely because #420 silently broke OTel tracing for
the `[local-timer] Timer fired: ${member} → enqueued ${taskType} task` log
line at `src/gateway/server.ts:1193`. With OTel working again, the same
misroute would have been visible as a `taskType=offload-l2` trace span
within seconds, instead of requiring a full source-read of the gateway
timer routing module.

This PR is intentionally narrow: it fixes the OTel Resource API mismatch
on `feat/server` (the v1.0.0 release branch) without touching the wider
OTel stack layout that `gugu23456789:r3-v1` addresses. Maintainers can
compare both implementations and pick whichever is cleaner.

## Compatibility notes

- No `package.json` change: we stay on the v2.x line.
- Public API unchanged: `OtlpObservabilityBackend.initialize()` signature
  and semantics are identical.
- New env knob `TDAI_OTEL_FAIL_OPEN=true` is opt-in and only affects the
  log level on init failure; default behavior is fail-closed (loud).
- `otel-sdk-init.ts` already uses a `createResource()` runtime-detect
  helper for v1/v2 compat, so this PR aligns `otlp-backend.ts` with that
  pattern but without the v1.x fallback (since package.json pins v2.x).

## Files changed

- `src/core/report/otlp-backend.ts` — use `resourceFromAttributes`, surface
  init failures via `console.error` + stack, add `TDAI_OTEL_FAIL_OPEN`.
- `src/core/report/otlp-backend-resource.test.ts` — new regression test
  (4 cases).

Closes #420.

### DIFF
--- a/src/core/report/otlp-backend-resource.test.ts
+++ b/src/core/report/otlp-backend-resource.test.ts
@@ -1,0 +1,111 @@
+/**
+ * Regression coverage for #420: v1.0.0 OTel SDK 版本不匹配 — `new Resource()`
+ * v1 API 与 `@opentelemetry/resources@^2.x` SDK 不兼容，会抛
+ * `TypeError: Resource is not a constructor`，导致 OTLP 后端在生产环境静默
+ * 退化为 console 输出，observability 完全失效。
+ *
+ * 这份测试在两种层面验证修复：
+ * 1. 直接验证 v2.x `resourceFromAttributes()` factory 可以替换 v1 `new Resource()`。
+ * 2. 验证 `OtlpObservabilityBackend.initialize()` 调用真实 OTel runtime 时，
+ *    不会因为 Resource API 变化而抛错，并能在 OTLP endpoint 不可达时优雅降级。
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+describe("otlp-backend Resource API compatibility (#420)", () => {
+  it("v2.x @opentelemetry/resources exports resourceFromAttributes (not Resource class)", async () => {
+    const resourcesModule = await import("@opentelemetry/resources");
+    // v2.x 移除 `Resource` 类，改为 `resourceFromAttributes()` factory。
+    // 这是 v1 API (`new Resource({...})`) 抛出 `TypeError` 的根因。
+    expect(typeof (resourcesModule as { Resource?: unknown }).Resource).toBe("undefined");
+    expect(typeof (resourcesModule as { resourceFromAttributes?: unknown }).resourceFromAttributes).toBe(
+      "function",
+    );
+  });
+
+  it("resourceFromAttributes({...}) returns a Resource with the supplied attributes", async () => {
+    const { resourceFromAttributes } = await import("@opentelemetry/resources");
+    const { ATTR_SERVICE_NAME } = await import("@opentelemetry/semantic-conventions");
+
+    const resource = resourceFromAttributes({
+      [ATTR_SERVICE_NAME]: "tdai-memory-test",
+    });
+
+    expect(resource).toBeDefined();
+    expect(resource.attributes[ATTR_SERVICE_NAME]).toBe("tdai-memory-test");
+  });
+
+  it("does not throw Resource constructor error when initializing OTLP backend", async () => {
+    // 真实回归：v1.0.0 tag 上的 `otlp-backend.ts` 用 `new Resource({...})`，
+    // 在 v2 SDK 下会抛 `TypeError: Resource is not a constructor`，
+    // 被 try/catch 静默吞掉。修复后应当不再触发 Resource constructor 路径。
+    const { OtlpObservabilityBackend } = await import("./otlp-backend.js");
+
+    const errorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+    const warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+    const logSpy = vi.spyOn(console, "log").mockImplementation(() => {});
+
+    try {
+      const backend = new OtlpObservabilityBackend();
+
+      // 用一个不存在的 endpoint，应当由网络层失败（OTLP exporter 内部超时），
+      // 而不是被 `TypeError: Resource is not a constructor` 阻断。
+      // 我们的 catch 会输出 console.error，所以这里 expect 调用过 console.error 即可。
+      await backend.initialize({
+        type: "otlp",
+        otel: {
+          enabled: true,
+          endpoint: "http://127.0.0.1:1", // 不存在端口
+          protocol: "http",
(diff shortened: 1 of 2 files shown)


YOMXXX:
I reviewed the actual patch and the `resourceFromAttributes()` change looks aligned with the OTel v2 API. One verification note: I could not reproduce the listed local checks with the current `feat/server` dependency state.

`pnpm install --frozen-lockfile` fails because `package.json` and `pnpm-lock.yaml` are out of sync on the target branch, and then the new test cannot import `@opentelemetry/resources`; `pnpm build` also fails on a missing `dayjs` module from the same install issue. Since this PR only changes `src/core/report/otlp-backend.ts` and adds the regression test, this may be a branch baseline problem rather than a bug in this patch, but CI/default installs will still fail until the lockfile is refreshed for `feat/server`.

Maxwell-Code07:
Thank you for your attention！ We will conduct a review and respond to you as soon as possible.

RerankerGuo:
Thanks for the verification note. I rechecked this from a fresh checkout of the current PR head (`5b3f495`) rather than an existing worktree.

One important baseline detail: neither `feat/server` at `105c335` nor this PR head tracks `pnpm-lock.yaml` or `package-lock.json`, and this PR does not change `package.json`. The target branch already declares both `dayjs` and the OTel v2 dependencies. A frozen pnpm install therefore depends on a locally generated lockfile and is not a reproducible install path for the tracked branch.

Fresh-checkout verification with Node 24.15.0:

```text
npm install
  added 567 packages; exit 0
  git status --short: clean

npx vitest run src/core/report/otlp-backend-resource.test.ts
  1 file / 4 tests passed

npm test
  1 file / 4 tests passed

npm run build
  passed (plugin and all scripts)

git diff --check
  passed
```

This confirms that the `resourceFromAttributes()` patch and regression test are reproducible from the repository's tracked state without adding unrelated lockfile churn. The earlier missing-module result appears consistent with an incomplete install or a stale local pnpm lock rather than a change introduced by this PR.

PR #428, opened 2026-07-08
security: harden supervisor and export script against injection/traversal

dandycrypto:
## Summary

Three hardening fixes for the Python supervisor and shell export script.

---

### 1. supervisor.py — Gateway command injection prevention

`GatewaySupervisor.ensure_running()` now validates `gateway_cmd` against a blocklist of shell metacharacters before passing it to `Popen`. While `shlex.split()` already prevents most injection vectors, a carefully crafted path on some platforms could slip through. Defense-in-depth.

**Blocked characters:** `| & ; $ \ ` ! < > ( ) { } [ ] " ' * ?`

### 2. export-diagnostic.sh — Tar flag injection prevention

`tar -czf "$ARCHIVE_PATH"` could interpret a path starting with `-` as a flag. Fixed by:
- Rejecting archive paths that start with `-`
- Using `--` to terminate option parsing
- Using `-P` to reject absolute paths

### 3. export-diagnostic.sh — Grep pattern injection prevention

The fallback config redaction used `grep -iE "$pattern"` with a shell-expanded variable. Fixed by writing patterns to a temporary file and using `grep -f` instead, eliminating any possibility of pattern injection.

### 4. export-diagnostic.sh — Path traversal guard

Added a case guard that rejects `EXPORT_DIR` starting with `/` or `..`, preventing directory traversal if a caller passes a malicious path.

---

## Scope note

These fixes cover the Python supervisor and shell export script. The TypeScript codebase (TCVDB HTTP default, backend `rejectUnauthorized: false`) is out of scope for this PR — those are Tencent-internal network/HTTPS edge cases requiring separate review as they affect users with specific non-default configurations.

---

## Testing

All edits are additive security checks. Existing behavior is unchanged for legitimate inputs. The export script continues to function identically for all valid paths.

### DIFF
--- a/hermes-plugin/memory/memory_tencentdb/supervisor.py
+++ b/hermes-plugin/memory/memory_tencentdb/supervisor.py
@@ -31,6 +31,11 @@ HEALTH_CHECK_RETRIES = 3     # retries for is_running check
 # Log file rotation parameters
 LOG_TAIL_BYTES_ON_CRASH = 2048  # bytes of stderr log to surface on startup crash
 
+# Shell metacharacters blocked in gateway_cmd to prevent command injection.
+# Defense-in-depth: shlex.split() already mitigates this, but we reject
+# suspicious input before it reaches Popen.
+_FORBIDDEN_CMD_CHARS = frozenset("|&;$\\`!<>(){}[]\"'*?")
+
 
 class GatewaySupervisor:
     """Manages the memory-tencentdb Gateway sidecar lifecycle."""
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Thank you for your attention！ We will conduct a review and respond to you as soon as possible.

YOMXXX (approved):
Approved as a focused security hardening PR for the Python supervisor and export diagnostic script.

Local verification:
- `python3 -m py_compile hermes-plugin/memory/memory_tencentdb/supervisor.py`
- `bash -n scripts/export-diagnostic.sh`
- `git diff --check`

This stays narrowly scoped to supervisor command validation and export script injection/traversal hardening. I did not run an end-to-end export smoke test, because this PR does not add one.

PR #442, opened 2026-07-09
fix(checkpoint): recalibrate counters from storage

xy200303:
## Description | 描述
Fixes checkpoint counter drift after cleanup or manual pruning by recalculating derived counters from actual persisted data.

- Add `CheckpointManager.recalibrate()` to refresh `total_processed`, `l0_conversations_count`, `total_memories_extracted`, and `memories_since_last_persona`.
- Prefer current store counts via `countL0()`, `countL1()`, and `queryL1Records({ updatedAfter })`, with JSONL shard counting as the fallback when the store is unavailable.
- Run recalibration during pipeline/core startup, after the memory cleaner completes, and in the existing L2 checkpoint repair path.
- Preserve non-derived checkpoint fields and keep the existing `scenes_processed` regression guard.
- Add regression tests for JSONL fallback counting, persona cursor handling, and vector-store-backed recalibration.

## Related Issue | 关联 Issue
Fix #157

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明
Verification run locally:

- `npx vitest run src/utils/checkpoint.test.ts`
- `npm test`
- `npm run build`

### DIFF
--- a/src/core/tdai-core.ts
+++ b/src/core/tdai-core.ts
@@ -409,6 +409,8 @@ export class TdaiCore {
       const stores = await initStores(this.cfg, this.dataDir, this.logger);
       this.vectorStore = stores.vectorStore;
       this.embeddingService = stores.embeddingService;
+      const checkpoint = new CheckpointManager(this.dataDir, this.logger);
+      await checkpoint.recalibrate(this.vectorStore);
       this.logger.debug?.(`${TAG} Stores initialized: backend=${this.cfg.storeBackend}, embedding=${this.cfg.embedding.provider}`);
     } catch (err) {
       this.logger.warn(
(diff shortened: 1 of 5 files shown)


YOMXXX:
Triage note: this overlaps with the already-approved canonical fix #177 for issue #157.

#177 is the selected baseline for checkpoint counter recalibration after cleanup and has already been locally verified/approved. This PR appears to solve the same #157 counter-drift problem with a very similar recalibration path.

Recommendation: close this as superseded by #177. If there is a specific improvement here that #177 lacks (for example a narrower fallback-counting edge case or an extra regression test), please split that into a small follow-up after #177 rather than keeping another full #157 implementation open.

PR #410, opened 2026-07-06
fix: stabilize cache-aware context handling

quan020406:
## Description | 描述

本 PR 针对 #120 中的 prompt cache 不稳定问题，调整插件上下文构建方式，减少每轮动态数据对可复用 prompt 前缀的影响。

核心目标：

- 稳定会话事实通过确定性 session snapshot 输出。
- 动态记忆默认按需工具召回，而不是每轮自动 prepend。
- 大工具结果在进入 prompt-facing history 前先卸载，避免长期污染历史。
- Mermaid/MMD 任务状态使用 snapshot/delta 结构，降低旧消息反复变化的风险。
- 为后续 Cache Epoch 长上下文治理补齐基础合约和测试。

主要改动：

- 新增 `src/core/session/session-snapshot.ts`，用于构建确定性 session snapshot。
- 默认启用 `recall.mode = "tool-only"`，同时保留 `recall.mode = "auto"` 兼容旧的自动召回路径。
- 拆分 `showInjected` 与 `persistInjected` 语义；当 `persistInjected=false` 时，在持久化消息前清理 `<relevant-memories>` 注入块。
- 新增工具结果 normalize/offload 逻辑，大结果写入 `refs/*.md`，历史中只保留 compact metadata。
- 新增 `tdai_offload_read`，支持按 `result_ref`、query 和行号范围受控读取完整结果片段。
- 新增 Cache Epoch helper，用于判断长上下文压缩边界。
- 新增 Mermaid/MMD task snapshot 和 delta helper。
- 补充相关回归测试，覆盖 session snapshot、tool-only recall、注入清理、工具结果卸载、ref 读取、epoch 判断和 task snapshot/delta。

## Related Issue | 关联 Issue

Fix #120

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [x] New feature | 新功能
- [ ] Documentation update | 文档更新
- [x] Code optimization | 代码优化

## Self-test Checklist | 自测清单

- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

本地验证命令：

```powershell
npm.cmd test
npm.cmd run build:plugin
node --check dist\index.mjs
git diff --check
```

验证结果：

- `13` 个测试文件通过。
- `82` 个测试通过。
- 插件构建通过。
- `dist/index.mjs` 语法检查通过。
- diff whitespace 检查通过。

额外做过本地长上下文验证：

- 48 turn。
[...]

PR #375, opened 2026-07-02
fix(recall): stabilize prompt-cache memory injection

Ricky-7-Yan:
## Description | 描述

Fixes #120.

Add explicit prompt-cache controls for TencentDB recall injection while keeping the default behavior backward compatible.

This PR keeps the TencentDB-side mitigation scoped:
- [x] Add `recall.injectionMode` with `prepend` as the backward-compatible default
- [x] Add opt-in `append` mode that emits dynamic L1 recall as OpenClaw `appendContext`
- [x] Add `recall.showInjected` so preserving `<relevant-memories>` in persisted history is explicit opt-in
- [x] Keep persisted history clean by default to avoid replaying stale dynamic recall blocks
- [x] Move OpenClaw recall shaping and injected-history stripping into a tested adapter helper
- [x] Document both options in the plugin manifest and README files

## Cache impact | 缓存影响
- [x] Stable memory context remains in `appendSystemContext`
- [x] Dynamic L1 recall can stay out of the user prompt prefix with `recall.injectionMode="append"` on compatible OpenClaw hosts
- [x] Persisted history avoids dynamic recall growth by default with `recall.showInjected=false`

## Related Issue | 关联 Issue

Fix #120

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [x] New feature | 新功能
- [x] Documentation update | 文档更新
- [x] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

## Verification | 验证

```bash
npx.cmd vitest run src/config.test.ts src/adapters/openclaw/recall-injection.test.ts
npx.cmd vitest run
npm.cmd run build
git diff --check
```

Results:
- [x] Targeted tests: 2 files / 8 tests passed
- [x] Full test suite: 6 files / 75 tests passed
- [x] Build passed
- [x] Whitespace check passed

PR #448, opened 2026-07-09
feat(offload): add bounded tool-result recovery

quan020406:
## Description | 描述

本 PR 从 #410 中拆分，仅聚焦工具结果卸载与按需读取行为。

主要改动：
- 在大工具结果进入 prompt-facing history 前进行确定性序列化和 token 估算。
- 超过 `inlineToolResultMaxTokens` 时，将完整结果写入 `refs/*.md`，prompt 中仅保留摘要、`result_ref`、内容 hash 和恢复指引。
- 新增 `tdai_offload_read`，支持按 query、行号范围和 token 预算读取最小必要片段。
- 为 ref 路径增加单层 `refs/*.md` 校验，并要求存在活动 session。
- L1 后续处理复用已生成的 `result_ref`，避免重复写入完整结果。

不包含以下内容：
- recall injection placement
- injected history cleanup
- cache epoch
- task snapshot / MMD snapshot 逻辑

## Related Issue | 关联 Issue

Related to #120

从 #410 拆分，按评审意见收敛为独立的工具结果卸载/读取 PR。

## Change Type | 修改类型

- [ ] Bug fix | Bug 修复
- [x] New feature | 新功能

## Self-test Checklist | 自测清单

验证命令：
- `npm.cmd test`
- `npm.cmd run build:plugin`
- `node --check dist\index.mjs`
- `git diff --check`

验证结果：
- 8 个测试文件、73 个测试通过。
- 插件构建通过。
- `dist/index.mjs` 语法检查通过。
- diff whitespace 检查通过。

## Additional Notes | 其他说明

该 PR 与 #375 独立，评审范围限定在有界卸载与恢复行为本身，不包含 recall 注入布局调整。

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -7,7 +7,7 @@
     "onStartup": true
   },
   "contracts": {
-    "tools": ["tdai_memory_search", "tdai_conversation_search"]
+    "tools": ["tdai_memory_search", "tdai_conversation_search", "tdai_offload_read"]
   },
   "configSchema": {
     "type": "object",
@@ -161,6 +161,10 @@
           "mildOffloadRatio": { "type": "number", "default": 0.5, "description": "温和压缩触发比例（占 context window）" },
           "aggressiveCompressRatio": { "type": "number", "default": 0.85, "description": "激进压缩触发比例" },
           "mmdMaxTokenRatio": { "type": "number", "default": 0.2, "description": "MMD 注入 token 预算比例" },
+          "inlineToolResultMaxTokens": { "type": "number", "default": 1200, "description": "单个工具结果超过该 token 估算值时，写入 refs 并在 prompt 中替换为 result_ref 摘要" },
+          "summaryMaxTokens": { "type": "number", "default": 350, "description": "工具结果卸载摘要 token 预算" },
+          "previewMaxChars": { "type": "number", "default": 1200, "description": "用于生成确定性摘要的预览字符数" },
+          "readChunkMaxTokens": { "type": "number", "default": 1600, "description": "tdai_offload_read 单次返回 token 预算" },
           "backendUrl": { "type": "string", "description": "后端服务 URL（如 https://offload-api.example.com），配置后 L1/L1.5/L2/L4 走后端" },
           "backendApiKey": { "type": "string", "description": "后端 API 认证 token" },
           "backendTimeoutMs": { "type": "number", "default": 10000, "description": "后端调用超时（毫秒）" }
--- a/src/config.tool-offload.test.ts
+++ b/src/config.tool-offload.test.ts
@@ -1,0 +1,23 @@
+import { describe, expect, it } from "vitest";
+import { parseConfig } from "./config.js";
+
+describe("tool-result offload config", () => {
+  it("provides bounded defaults and parses explicit limits", () => {
+    const defaults = parseConfig({});
+    expect(defaults.offload.inlineToolResultMaxTokens).toBe(1200);
+    expect(defaults.offload.readChunkMaxTokens).toBe(1600);
+
+    const configured = parseConfig({
+      offload: {
+        inlineToolResultMaxTokens: 64,
+        summaryMaxTokens: 24,
+        previewMaxChars: 320,
+        readChunkMaxTokens: 512,
+      },
+    });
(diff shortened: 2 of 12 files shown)


Maxwell-Code07:

YOMXXX (changes requested):
I cannot approve this as-is.

`tdai_offload_read` needs a stricter session boundary. The new tool reads full offloaded tool-result refs, but when `ctx.sessionKey` is missing it falls back to `_lastActiveMgr`:

```ts
const mgr = ctx?.sessionKey ? await _resolveSession(ctx.sessionKey, ctx?.sessionId) : _lastActiveMgr;
```

That is not safe for a ref-reading tool. A caller without a session key can read from whichever session happened to be active last, as long as it supplies a `refs/*.md` name. Tool-result refs may contain raw command output, file contents, diagnostics, or other sensitive task data, so this should not inherit the looser fallback pattern used by internal lifecycle hooks.

Please require an active `ctx.sessionKey` (or another explicit, validated session binding) for `tdai_offload_read`, and add tests covering:

- missing `ctx.sessionKey` is rejected instead of using `_lastActiveMgr`
- a valid session can read its own ref
- traversal/nested/absolute paths remain rejected
- requested `max_tokens` cannot exceed the configured hard limit

There is also a correctness gap worth testing: when `replacePromptFacingToolResult()` cannot find the matching tool-result message, the raw large result may remain in `event.messages` even though `event.result` was replaced. The front-offload path should either guarantee replacement or log/fallback in a way that does not leave the prompt-facing history unbounded.

quan020406:
Addressed in e760342.

`tdai_offload_read` now requires an explicit `ctx.sessionKey` and no longer falls back to `_lastActiveMgr`. The read is resolved only through the validated session registry entry.

Added coverage for:

- rejecting a missing `ctx.sessionKey`;
- reading a ref from the explicitly bound session;
- traversal, nested, and absolute ref-path rejection;
- capping requested `max_tokens` at the configured hard limit.

Also made front-offload replacement atomic: `event.result` is replaced only after the matching prompt-facing tool-result message is replaced. If no matching message exists, the raw result is retained and a warning is logged for normal L3 handling.

Validation:
- `npm.cmd test` (`76/76`)
- `npm.cmd run build:plugin`
- `git diff --check`

quan020406:
Fixed in 9fc5563.

The ref directory is agent-scoped, so requiring a valid sessionKey alone was insufficient. Refs now carry an owner sessionKey marker when they are written, and `tdai_offload_read` validates that marker against the caller's `ctx.sessionKey` before returning content.

A ref written by `agent:main:session-1` is now rejected for `agent:main:session-2` with the same not-found response used for missing refs. The owning session can still read it.

Legacy refs without an owner marker are intentionally treated as unreadable through `tdai_offload_read`, which is the safe default.

Validation:
- `npm.cmd test` (76/76)
- `npm.cmd run build:plugin`
- `git diff --check`

YOMXXX (approved):
Approved. Thanks for the follow-up fixes.

I re-reviewed the latest commit (`9fc5563`) against the previous blocking feedback:

- `tdai_offload_read` now requires an explicit `ctx.sessionKey` and no longer falls back to `_lastActiveMgr`.
- Ref reads are now owner-bound via the stored session marker, so a ref from `agent:main:session-1` is rejected from `agent:main:session-2`.
- Legacy refs without an owner marker are unreadable through `tdai_offload_read`, which is the safer default.
- The front-offload replacement path now only replaces `event.result` after the prompt-facing message replacement succeeds; otherwise it retains the raw result for normal L3 handling and logs the fallback.

Verified locally on the PR branch:

- `COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/offload/index.test.ts src/offload/storage.test.ts src/offload/tool-result-normalizer.test.ts src/config.tool-offload.test.ts` — passed: 4 files, 9 tests
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm test` — passed: 8 files, 76 tests
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm build:plugin` — passed
- `git diff --check upstream/main...HEAD` — passed

Only local environment note: my Node is `v22.14.0`, below the project engine `>=22.16.0`, so pnpm/tsdown emitted engine/deprecation warnings, but the tests and build completed successfully.

Maxwell-Code07:

PR #391, opened 2026-07-04
fix(security): 代码审查安全加固 — Gateway/沙箱/TLS/存储/Offload 缺陷修复

rainforest888:
## 背景

对 `src/` 与 `index.ts` 做了一轮安全 + 健壮性代码审查，修复发现的缺陷。本 PR 仅含审查修复，不改变公开 API 行为。

## 验证

- `vitest run` → **67/67 通过**
- `tsc`（strict，含 src + index.ts）→ **零新增错误**（baseline 38 行历史错误 = 修复后 38 行；本 PR 引入的唯一一处已在提交前修正）

### HIGH

| 位置 | 问题 | 修复 |
|:---|:---|:---|
| `src/gateway/server.ts` | `parseJsonBody` 无 body 大小限制，且默认免鉴权 → 远程 OOM DoS | 加 16 MiB 上限，超限返回 413 并断开连接 |
| `src/adapters/standalone/llm-runner.ts` | `resolveSandboxedPath` 用 `startsWith` 无尾部分隔符 → `../tdai-backup` 类前缀穿越 | 改为 `resolved === root \|\| resolved.startsWith(root + path.sep)` |
| `src/offload/backend-client.ts` | 所有后端 HTTPS 调用无条件 `rejectUnauthorized:false` → MITM 窃取 API key | 默认校验；仅 `TDAI_BACKEND_INSECURE_TLS=1` 时 opt-in |
| `src/core/store/tcvdb.ts` | `session_key`/`session_id` 直接插值进过滤表达式 → 注入跨会话泄露 | 新增 `escapeTcvdbFilterString`，转义 `\` 与 `"` |

### MED

- **gateway/server.ts**：`/seed` 禁止覆盖 `llm.*` 防端点劫持；500 回 generic "Internal error" 防内部信息泄露；`stop()` 用 `closeAllConnections()` 防关闭挂起；`safeEqual` 改 SHA-256 后比较防时序泄漏 key 长度；seed outputDir 加随机后缀防同秒并发碰撞
- **offload/opik-tracer.ts**：`createRequire(import.meta.url)` 加载 opik（ESM 下 `require` 失效导致 tracing 永久关闭）；trace 内容统一截断到 2000 字符防未脱敏 tool 输出外泄
[...]

PR #289, opened 2026-06-30
fix(offload): sanitize ref and skill paths

YOMXXX:
## Summary
- sanitize timestamp-derived ref filenames before writing `refs/*.md`
- resolve `readRefMd` only within the refs directory
- sanitize backend-provided L4 skill names before creating skill directories

## Why
Ref filenames and L4 skill names are string inputs that were joined into filesystem paths. Dot segments or separators could escape the intended offload storage subdirectories.

## Tests
- `npm test -- src/offload/storage-ref.test.ts`
- `npm test`
- `npm run build`
- `git diff --check`

PR #288, opened 2026-06-30
fix(offload): keep MMD files inside storage dir

YOMXXX:
## Summary
- normalize L1.5 task labels and continuation MMD filenames before activating/writing MMDs
- route MMD read/write/patch/delete through a safe single-segment `.mmd` filename
- add regression tests for unsafe labels and direct MMD writes escaping `mmds/`

## Why
L1.5 judgments come from the backend/LLM. A `newTaskLabel` or `continuationMmdFile` containing dot segments could make `writeMmd` operate outside the agent `mmds/` directory.

## Tests
- `npm test -- src/offload/hooks/before-agent-start.test.ts`
- `npm test -- src/offload/hooks/before-agent-start.test.ts src/offload/auth-profile-key.test.ts`
- `npm test`
- `npm run build`
- `git diff --check`

PR #287, opened 2026-06-30
fix(offload): sanitize session storage paths

YOMXXX:
## Summary
- sanitize offload agent/session path segments at StorageContext creation
- reuse the same sanitizer for invalid session-key fallback routing
- store sanitized session ids/offload filenames in the session registry

## Why
Host-provided session ids are used in `offload-<sessionId>.jsonl`. Path separators or dot segments could escape the intended agent storage directory, and fallback session keys like `..` could point the agent directory outside the data root.

## Tests
- `npm test -- src/offload/session-registry.test.ts`
- `npm test -- src/offload/auth-profile-key.test.ts src/offload/session-registry.test.ts`
- `npm test -- src/utils/sanitize.test.ts src/offload/session-registry.test.ts`
- `npm test`
- `npm run build`
- `git diff --check`

PR #461, opened 2026-07-11
fix(core): storage hardening — Buffer safety, embedding abort, atomic instanceId

rainforest888:
# fix(offload): Tracing & Token Hardening

拆分自 #391，per [YOMXXX review](https://github.com/TencentCloud/TencentDB-Agent-Memory/pull/391#issuecomment-4923202779)。

## 修复范围

聚焦 `src/offload/` 层：token estimate、local LLM、Opik tracing。不涉及 store、runtime、gateway 层。

### `src/offload/fast-token-estimate.ts` — Astral-平面 Token 修正

| 问题 | 修复 |
|:---|:---|
| astral-plane 字符（emoji / CJK extension B，U+10000+）编码为 surrogate pair，原实现把 high surrogate 和 low surrogate 各计为一个 token | high surrogate 检测后跳过低代理（`i += 2`），修正 ~2× token 高估 |

### `src/offload/local-llm/index.ts` — L1.5 Fallback 修正

| 问题 | 修复 |
|:---|:---|
| 解析失败返回 `{taskCompleted: false, ...}` — normalizeJudgment 把 `false` 当有效判断，永远不触发 "LLM 不可用" 重试 | 改为 `{taskCompleted: null, ...}`，`== null` 检查触发重试路径 |

### `src/offload/opik-tracer.ts` — ESM 兼容 + Trace 截断

| 问题 | 修复 |
|:---|:---|
| ESM 下 `require("opik")` 为 `undefined`，tracer 永久静默关闭 | 改为 `createRequire(import.meta.url)` |
| 未脱敏 tool 输出（文件内容、环境变量）完整写入 trace，泄露到 Opik backend | 统一截断到 2000 字符 |

## 验证

- 3 files, +41/-15
- 无 API 变更，无新增依赖
- 与 #39 / #232 / #242 / #287 / #288 / #289 / #347 均不重叠

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,35 @@
+# fix(core): Storage Hardening
+
+拆分自 #391，per [YOMXXX review](https://github.com/TencentCloud/TencentDB-Agent-Memory/pull/391#issuecomment-4923202779)。
+
+## 修复范围
+
+聚焦 `src/core/store/` + `src/core/report/`，不涉及 Gateway、Offload、Runtime 层。
+
+## 具体修复
+
+### `src/core/store/sqlite.ts` — Buffer 安全 + Extension 锁定
+
+| 问题 | 修复 |
+|:---|:---|
+| `Buffer.from(emb.buffer)` 缺少 byteOffset/byteLength，Float32Array subarray 视图会写出整个底层 buffer | 6 处补全 `byteOffset, byteLength` 参数 |
+| `enableLoadExtension(true)` 加载 sqlite-vec 后未重置，未来 SQL 路径可能加载任意 extension | 加载后立即 `enableLoadExtension(false)` |
+
+### `src/core/store/embedding.ts` — Abort 信号贯穿
+
+| 问题 | 修复 |
+|:---|:---|
+| recall 超时后 embedding HTTP 请求继续占用 API slot / 连接 | `EmbeddingCallOptions` 新增 `abortSignal`，`AbortSignal.any` 合并超时与外部信号 |
+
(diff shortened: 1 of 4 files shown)


Maxwell-Code07:
Thank you for your contribution! We will review your code and get back to you as soon as possible.

YOMXXX (changes requested):
Thanks for the focused core hardening split. I verified the branch locally with:

- `pnpm test` — passed: 4 files, 67 tests
- `pnpm build` — passed
- `git diff --check HEAD~2..HEAD` — passed
- a local Float32Array subarray Buffer check — passed

I think this still needs one change before approval: the new `EmbeddingCallOptions.abortSignal` is wired into the embedding HTTP fetch, but the actual auto-recall timeout path does not pass an abort signal. `performAutoRecall()` still uses `Promise.race` with a timer that resolves after `recall.timeoutMs`; once that race resolves, `performAutoRecallInner()` and the embedding HTTP request can continue running until the embedding request's own timeout. With defaults, `recall.timeoutMs` is 5000ms while `embedding.timeoutMs` is 10000ms and `embedding.recallTimeoutMs` may be unset, so the PR does not fully deliver the documented "recall timeout aborts the in-flight embedding request" behavior.

Please create an `AbortController` for the recall timeout path and pass its signal through `searchMemories()` / `EmbeddingCallOptions` so the same timeout that skips memory injection also cancels the embedding fetch. The Buffer byteOffset/byteLength fixes and the instance-id in-flight guard look directionally good.

PR #39, opened 2026-05-18
fix(gateway): harden CORS / Host header / startup host binding

YOMXXX:
## Description | 描述

TDAI Gateway 的 HTTP 层和启动入口有三处独立于任何在飞 feature 的安全缺陷，在当前 `main` 上已经可利用。本 PR 全部修复。

Three security holes in the TDAI Gateway HTTP layer and startup path that exist independently on `main` today, all fixed here.

### 1. CORS hardcoded `Access-Control-Allow-Origin: *` (`src/gateway/server.ts:177-186`)

`handleRequest` 无条件下发 `Access-Control-Allow-Origin: *` 并对所有 OPTIONS 返回 204。daemon 默认绑 `127.0.0.1`，但浏览器通过 **DNS rebinding**（攻击者域名 `evil.com` 短 TTL DNS → 解析为 `127.0.0.1`）发起的 fetch 直接通过 —— 加上 #2，daemon 数据端点对恶意浏览器页面开放。

`handleRequest` unconditionally emitted `Access-Control-Allow-Origin: *` and acked all OPTIONS with 204. Although the daemon binds loopback by default, a malicious page using DNS rebinding (short-TTL `evil.com → 127.0.0.1`) can issue fetch() and the daemon accepts — combined with #2 this exposes data endpoints to any browser tab.

### 2. Host header unvalidated (`src/gateway/server.ts:172-174`)

`req.headers.host` 不校验，配合 #1 形成 DNS rebinding 攻击面。

`req.headers.host` was not validated; pairs with #1 to enable DNS rebinding.

### 3. `server.ts main()` did not validate `TDAI_GATEWAY_HOST` (`src/gateway/server.ts:500-512`)

`tsx src/gateway/server.ts` 直接启动路径不校验 `TDAI_GATEWAY_HOST`。环境里若设了 `TDAI_GATEWAY_HOST=0.0.0.0`，daemon 静默绑全网卡，把记忆数据端口暴露到 LAN。

The auto-start path (`tsx src/gateway/server.ts`) bound whatever `TDAI_GATEWAY_HOST` was set in env — a stray `0.0.0.0` would silently expose the daemon to the LAN.

## Fix | 修复

- **CORS opt-in**: `Access-Control-Allow-*` 头仅在 `TDAI_GATEWAY_CORS_ORIGIN` 显式设置时下发；OPTIONS preflight 仅在该模式下返回 204。默认状态下 OPTIONS 落入正常路由 404，不再作为永久未鉴权 probe 端点。
[...]

PR #462, opened 2026-07-11
fix(runtime): race/timeout hardening — recall abort, profile backup-swap, sandbox, SQLITE_BUSY

rainforest888:
# fix(offload): Tracing & Token Hardening

拆分自 #391，per [YOMXXX review](https://github.com/TencentCloud/TencentDB-Agent-Memory/pull/391#issuecomment-4923202779)。

## 修复范围

聚焦 `src/offload/` 层：token estimate、local LLM、Opik tracing。不涉及 store、runtime、gateway 层。

### `src/offload/fast-token-estimate.ts` — Astral-平面 Token 修正

| 问题 | 修复 |
|:---|:---|
| astral-plane 字符（emoji / CJK extension B，U+10000+）编码为 surrogate pair，原实现把 high surrogate 和 low surrogate 各计为一个 token | high surrogate 检测后跳过低代理（`i += 2`），修正 ~2× token 高估 |

### `src/offload/local-llm/index.ts` — L1.5 Fallback 修正

| 问题 | 修复 |
|:---|:---|
| 解析失败返回 `{taskCompleted: false, ...}` — normalizeJudgment 把 `false` 当有效判断，永远不触发 "LLM 不可用" 重试 | 改为 `{taskCompleted: null, ...}`，`== null` 检查触发重试路径 |

### `src/offload/opik-tracer.ts` — ESM 兼容 + Trace 截断

| 问题 | 修复 |
|:---|:---|
| ESM 下 `require("opik")` 为 `undefined`，tracer 永久静默关闭 | 改为 `createRequire(import.meta.url)` |
| 未脱敏 tool 输出（文件内容、环境变量）完整写入 trace，泄露到 Opik backend | 统一截断到 2000 字符 |

## 验证

- 3 files, +41/-15
- 无 API 变更，无新增依赖
- 与 #39 / #232 / #242 / #287 / #288 / #289 / #347 均不重叠

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,41 @@
+# fix(runtime): Race & Timeout Hardening
+
+拆分自 #391，per [YOMXXX review](https://github.com/TencentCloud/TencentDB-Agent-Memory/pull/391#issuecomment-4923202779)。
+
+## 修复范围
+
+聚焦 runtime 层：hooks、profile、sandbox、gateway lifecycle。不涉及 store、offload 层。
+
+## 具体修复
+
+### `src/core/hooks/auto-recall.ts` — Recall 超时 Abort 贯穿
+
+| 问题 | 修复 |
+|:---|:---|
+| recall 超时后 Promise.race 返回 undefined，但 embedding/VectorStore 仍在运行 | 新增 `AbortController`，超时时 abort，信号贯穿到 embedding |
+
+### `src/core/profile/profile-sync.ts` — Scene Blocks Backup-Swap
+
+| 问题 | 修复 |
+|:---|:---|
+| 原实现先 `rm -rf scene_blocks` 再 rename，rename 失败则本地 blocks 永久丢失 | 改为 backup-swap：先 rename 旧目录为 `.old-*`，rename 失败则恢复 |
+| 并发 pull 时 rename race 检测保留 | 保留 `isRenameRaceError` 判断 |
+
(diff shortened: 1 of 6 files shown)


Maxwell-Code07:
Thank you for your contribution! We will review your code and get back to you as soon as possible.

YOMXXX (changes requested):
Thanks for splitting this runtime hardening work. I verified the PR patch locally by applying it to its base commit and running:

- `pnpm test` — passed: 4 files, 67 tests
- `pnpm build` — passed
- `git diff --check --cached` — passed

However, the recall abort fix is incomplete as a standalone PR. `performAutoRecall()` now creates an `AbortController` and passes the signal through `searchMemories()`, but this PR does not update `src/core/store/embedding.ts`: `EmbeddingCallOptions` still has no `abortSignal` field, and the embedding HTTP request path does not combine or use that signal. Because the object literal uses a conditional spread, TypeScript still builds, but at runtime the extra `abortSignal` option is ignored by the current embedding services. So the documented behavior — recall timeout aborts the in-flight embedding request — does not actually take effect in this PR by itself.

Please either include the embedding-layer wiring in this PR, or make this PR explicitly depend on/stack after the storage/embedding PR that adds `EmbeddingCallOptions.abortSignal` and passes it to fetch. Once the abort signal reaches the actual fetch call, the rest of the timeout propagation shape here looks reasonable.

PR #195, opened 2026-06-11
fix(hermes): split recall context for prefetch

RerankerGuo:
## Description | 描述
Split Gateway recall responses into stable system context and dynamic L1 prefetch context for Hermes.

This fixes the Hermes path where `/recall` only returned `appendSystemContext`, so L1 memories recalled by auto-recall were not injected through `prefetch()`. The Gateway now returns:
- `system_context` for persona / scene navigation / tool guidance
- `prepend_context` for dynamic L1 memories
- `context` as a backward-compatible combined field for older clients

The Hermes provider prefers `prepend_context` in `prefetch()` and caches `system_context` for `system_prompt_block()`, while still falling back to the legacy `context` response.

## Related Issue | 关联 Issue
Fix #101

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [ ] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

## Additional Notes | 其他说明
Verified with `npx vitest run src/gateway/server.test.ts`, `npm test`, and `npm run build` using Node v24.15.0.
Also ran `python3 -m py_compile` for the Hermes provider files. `pytest` is not installed in this environment, so the Python pytest suite could not be executed here.

PR #492, opened 2026-07-13
fix(hermes): inject stable system context

RerankerGuo:
## Description | 描述

Fixes the main Hermes persona/system-prompt bug in #205. The two additional write/tool bullets in that issue remain out of scope.

This is a focused follow-up to the approved split-context contract in #195. `system_prompt_block()` now obtains and caches only the stable `system_context` returned by that contract, so generated L3 persona/system guidance is visible to Hermes without moving dynamic L1 recall into the stable system prompt.

## Why this shape | 方案依据

This also addresses the verification/runtime gaps noted on the current #206 candidate:

- the Gateway rejects `query=""`; this implementation uses a reserved non-empty probe;
- no second `recalledL3_persona` wire field is added because #195 already exposes the correct stable `system_context` boundary;
- the backward-compatible combined `context` and dynamic `prepend_context` fields are explicitly ignored by the system-prompt refresh path;
- regression tests live outside the provider package and stub `agent.memory_provider` before import, so both unittest and pytest run from a repo-only checkout;
- the root test path is excluded from the npm package.

## Behavior | 行为

- Refresh stable system context on the first prompt build and after a 60-second TTL.
- Let a normal user-query `prefetch()` prime the same cache, avoiding a second Gateway call.
- Prime the prompt cache only when prefetch belongs to the provider's own session, preventing cross-session context leakage.
- Cache successful empty responses so a Gateway without generated persona is not polled on every prompt.
- On recall failure, keep the last-known stable context, record the failure, and trigger the existing recovery path without breaking prompt construction.
- Reset cached context when the provider is initialized for a session/user.

## Dependency | 依赖

This PR depends on #195 and starts from its approved head commit `e644d0b`.

Until #195 lands, the diff against `main` also includes that PR's split-context commit. The unique change here is the signed commit `1c85c31`; relative to #195 it changes only the Hermes provider and one focused repo-only test file. I will rebase onto `main` after #195 merges.

## Verification | 验证

```text
$ python3 -m pytest tests/hermes/test_memory_tencentdb_system_context.py
7 passed

$ python3 tests/hermes/test_memory_tencentdb_system_context.py
Ran 7 tests - OK

$ npm test
Test Files  5 passed (5)
Tests       68 passed (68)

$ npm run build
Build complete

$ npm pack --dry-run --json
Passed; 148 package entries and tests/hermes is excluded

$ python3 -m py_compile hermes-plugin/memory/memory_tencentdb/__init__.py tests/hermes/test_memory_tencentdb_system_context.py
Passed

$ git diff --check
Passed
```

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

- [x] Commit includes the required DCO sign-off

### DIFF
--- a/hermes-plugin/memory/memory_tencentdb/__init__.py
+++ b/hermes-plugin/memory/memory_tencentdb/__init__.py
@@ -375,6 +375,7 @@ class MemoryTencentdbProvider(MemoryProvider):
         self._user_id = ""
         self._gateway_available = False
         self._initialized = False  # Track if initialize() has been called
+        self._latest_system_context = ""
 
         # Background sync threads.
         # We allow at most _MAX_INFLIGHT_SYNCS in-flight sync threads at any
@@ -816,7 +817,7 @@ class MemoryTencentdbProvider(MemoryProvider):
     def system_prompt_block(self) -> str:
         if not self._gateway_available:
             return ""
-        return (
+        block = (
             "# memory-tencentdb Memory\n"
             f"Active. User: {self._user_id}.\n"
             "Four-layer memory system (L0→L1→L2→L3) with automatic conversation "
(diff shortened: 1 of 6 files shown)


Maxwell-Code07:
Thanks for your attention! We will review your code and get back to you as soon as possible.

PR #509, opened 2026-07-16
[codex] fix checkpoint counter recalibration after cleanup

XiZu233:
## Summary

Fixes #157.

This PR adds a startup recalibration path for checkpoint counters that can drift after memory cleanup. total_memories_extracted and l0_conversations_count were increment-only fields in recall_checkpoint.json; after records were removed from the underlying store, the checkpoint could continue reporting stale inflated totals.

## Approach

- Add CheckpointManager.recalibrate() and keep it inside the existing mutate() path, so the update still uses the per-file async lock and atomic tmp+rename write.
- Recompute the two affected counters from the existing IMemoryStore abstraction via countL1() and countL0() during TdaiCore store startup.
- Adjust memories_since_last_persona conservatively when L1 totals decrease, so persona generation thresholds do not fire early because of deleted memories.
- Skip startup recalibration when the store is unavailable, degraded, or reports both counts as zero while the checkpoint is non-empty. That avoids accidentally clearing counters if a fault-tolerant count implementation returns 0 during a transient store/count failure.

## Notes

l0_conversations_count is recalibrated from countL0(), matching the issue's expectation to count actual l0_conversations rows. This means the calibrated value reflects stored L0 records rather than historical capture rounds.

## Validation

- npm test -- src/utils/checkpoint.test.ts src/core/tdai-core.test.ts
- npm test
- npm run build

### DIFF
--- a/src/core/tdai-core.test.ts
+++ b/src/core/tdai-core.test.ts
@@ -1,0 +1,196 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+
+// Mock only initStores — the rest of pipeline-factory (initDataDirectories,
+// resetStores, ...) stays real so TdaiCore can run against a temp dataDir.
+vi.mock("../utils/pipeline-factory.js", async (importOriginal) => {
+  const orig = await importOriginal<typeof import("../utils/pipeline-factory.js")>();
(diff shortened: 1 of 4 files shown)


XiZu233:
主要使用codex+kimi完成本次issue的修复，另外编码部分使用k3模型，针对本任务的测试结果还算OK

Maxwell-Code07:

Maxwell-Code07:

PR #519, opened 2026-07-18
fix(queue): recover from synchronous task throws

### LINKED ISSUE #518: Bug: SerialQueue deadlocks after a synchronous task throw
﻿## Bug description

## Reproduction

```ts

await queue.add(() => {
}).catch(() => undefined);

  next,
]);

```

Actual result:

```text
timeout 1 true false
```

## Proposed fix

yangjj-iso:
﻿## Description | ??

`SerialQueue.drain()` previously invoked `entry.task()` before attaching the promise cleanup chain. A synchronous exception therefore escaped without running `.finally()`, leaving `running = true` forever. Later tasks stopped executing and `onIdle()` never resolved.

This PR starts each task from a resolved promise so both synchronous throws and asynchronous rejections follow the same rejection and cleanup path. It also adds a regression test covering error propagation, continued FIFO processing, and the final idle state.

## Related Issue | ?? Issue

Fixes #518

## Change Type | ????

- [x] Bug fix | Bug ??
- [ ] New feature | ???
- [ ] Documentation update | ????
- [ ] Code optimization | ????

## Self-test Checklist | ????

- [x] Verified locally | ??????
- [x] No existing features affected | ???????

## Additional Notes | ????

- `npm test` ? 5 test files / 68 tests passed
- `npm run build` ? plugin and all three scripts built successfully
- `git diff --check` ? passed
- Commit includes the required DCO sign-off

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@
   - 修复 offload local-llm 模式下每次 LLM 调用都重新创建 fetch wrapper 的性能问题（现在在 `LocalLlmClient` 构造函数中创建一次并缓存）。
   - 注入逻辑抽取到 `src/utils/no-think-fetch.ts` 共享，新增 vitest 单测覆盖全部策略 / 跳过 embedding / 非 JSON 容错。
 
+### 🐛 修复
+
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:
Thank you for your attention. We will schedule an internal review of your PR as soon as possible and get back to you with feedback.

PR #558, opened 2026-07-22
fix(recall): prompt-cache-friendly injection mode and history strippi…

wzxsdf:
## Description

Fixes #120.

Adds explicit, backward-compatible prompt-cache controls for TencentDB recall injection. **Default behaviour is unchanged** (`injectionMode="prepend"` + `showInjected=false`), so this is a zero-impact upgrade unless the operator opts in.

- **`recall.injectionMode`** (default `"prepend"`): opt-in `"append"` routes dynamic L1 recall to the host's `appendContext` (after the user prompt) instead of `prependContext` (before it), so the user-prompt prefix stays byte-stable across turns — the primary lever for prefix-matching cache providers (DeepSeek, MiMo). Requires a host that consumes `appendContext`.
- **`recall.showInjected`** (default `false`): makes the existing `before_message_write` stripping of `<relevant-memories>` explicit and opt-in, so persisted history doesn't replay stale dynamic recall and keep growing the prompt prefix.
- **Lightweight prefix diagnostics**: each `before_prompt_build` logs the stable-block hash and dynamic placement (`stable=Nchars(hash=…)`, `dynamic=prepend|append/Nchars`). Observation-only — no behaviour change — lets operators verify prefix stability and the effect of `injectionMode`.
- **Layering**: the strip helper is host-neutral (`src/utils/relevant-memories.ts`); injection routing is OpenClaw-specific (`src/adapters/openclaw/recall-injection.ts`); diagnostics live in `src/utils/recall-shape-diagnostics.ts`.

## Scope decisions

- **No session-level freeze / dedupe of the stable block.** The "stable block byte drift" chain does not exist in this repo: `generateSceneNavigation` is deterministic (sort by heat, no timestamps/random) and `heat` is updated only during L2 scene extraction — not on the recall path. Therefore `appendSystemContext` (persona + scene nav + tools guide) is already byte-stable across turns while source files are unchanged. A freeze would add no value and would additionally delay L2/L3 persona updates.
- **No persona relocation to `prependSystemContext`.** Final placement relative to the host's cache boundary depends on host-side semantics that cannot be verified offline, so it is intentionally left out of scope.

## Cache impact

- `injectionMode="append"`: dynamic L1 recall moves out of the user-prompt prefix → prefix-matching cache stays warm (on hosts supporting `appendContext`).
- `showInjected=false` (default): persisted history avoids dynamic-recall growth → less truncation jitter over multi-turn conversations.
- Default config = legacy behaviour.

## Verification

```bash
npx vitest run          # 8 files / 84 tests passed
npm run build:plugin    # passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,12 @@
   - 环境变量 `TDAI_LLM_DISABLE_THINKING` 支持策略名（如 `deepseek`）和布尔值。
   - 修复 offload local-llm 模式下每次 LLM 调用都重新创建 fetch wrapper 的性能问题（现在在 `LocalLlmClient` 构造函数中创建一次并缓存）。
   - 注入逻辑抽取到 `src/utils/no-think-fetch.ts` 共享，新增 vitest 单测覆盖全部策略 / 跳过 embedding / 非 JSON 容错。
+- **提示词缓存友好性：`recall.injectionMode` + `recall.showInjected`** ([#120](https://github.com/TencentCloud/TencentDB-Agent-Memory/issues/120))：为召回注入增加显式的缓存控制，默认行为完全向后兼容。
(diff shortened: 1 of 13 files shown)


Maxwell-Code07:

Maxwell-Code07:

PR #564, opened 2026-07-23
fix(config): add cross-platform config path resolution

March-77:
## Summary\n- Add shared config path helpers (home dir + user config dirs + optional file checks).\n- Gateway config resolution now checks XDG and ~/.config/tencentdb-agent-memory before data-dir fallback.\n- Hook policy resolution now reuses safe config-path helpers and resolves ~/<app>/openclaw.json more reliably.\n- OpenClaw state-dir resolution now uses the same home resolution path.\n- Add regression tests for XDG config and USERPROFILE fallback behaviors (no HOME).\n\n## Fixes\n- #103\n

### DIFF
--- a/src/gateway/config.ts
+++ b/src/gateway/config.ts
@@ -12,6 +12,7 @@ import fs from "node:fs";
 import path from "node:path";
 import YAML from "yaml";
 import { getEnv } from "../utils/env.js";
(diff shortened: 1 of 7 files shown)


Maxwell-Code07:
Thann you!Please provide more detailed and standardized PR information.

PR #433, opened 2026-07-08
fix(recall): move dynamic L1 memory to appendContext to preserve prefix cache

L-yifan:
## Summary | 描述

Fix #120 — 修复 OpenAI-compatible provider（DeepSeek、MiMo）的 prompt prefix-matching 缓存命中率退化。

### 根因

动态 L1 记忆召回（~500-1700 tokens）每轮注入到 `prependContext`（用户消息开头），导致每轮的 prompt 前缀哈希都不同 → prefix-matching 缓存无法复用。

### 方案

将动态 L1 召回从 `prependContext` 移动到 `appendContext`（用户消息之后），保持 prompt prefix 稳定，同时模型在当前轮仍看到完整的记忆内容。

## 关键改动

| 文件 | 说明 |
|------|------|
| `src/config.ts` | 新增 `recall.dynamicContextPlacement`（"append" 默认、"prepend" 回退）和 `recall.showInjected` 开关 |
| `src/core/hooks/auto-recall.ts` | 将动态 L1 从 placement 决策中抽取出来；core 始终输出 prependContext，适配器决定最终位置 |
| `src/adapters/openclaw/recall-injection.ts` | OpenClaw 适配器：placement 路由（prepend ↔ appendContext）+ showInjected 清理控制 |
| `src/utils/sanitize.ts` | 集中化 `<relevant-memories>` 清理（string + parts），处理不完整标签边界 |
| `src/utils/prompt-shape-diagnostics.ts` | 四段上下文观察：appendSystemContext / history / prependContext / appendContext |
| `index.ts` | 集成适配器路由 + 统一清理 + 诊断日志 |
| `docs/issue-120-cache-safe-recall-measurement.md` | 完整 A/B 测量报告 |

## A/B 验证结果

**2 provider × 2 variant × 2 repeat × 3 turn = 24 cells**

| Provider | Warm-turn cacheRead (prepend → append) | cache miss |
|----------|---------------------------------------|------------|
| DeepSeek | 42,752 → **52,544** (+23%) | 5,795 → 4,586 (-21%) |
| MiMo | 36,576 → 32,736 | 1,434 → 2,857 |

- `append` 在所有 12 cells 中 prependContext=0、动态 L1 仅在 appendContext
- 持久化 history 在所有 24 cells 中 historyRelevantMemories=0
- MiMo turn-3 cacheRead=0 是 provider 端 cache-window 特性，非回归

## 配置

```json
{
  "recall": {
    "dynamicContextPlacement": "append",
[...]

PR #569, opened 2026-07-23
fix(hooks): cache-safe recall placement, deterministic ordering, session dedup

liu66-qing:
## Description | 描述

Fix prefix cache hit-rate regression on `openai-completions` providers (DeepSeek V4 Pro, MiMo V2.5 Pro) reported in #120.

Three independent cache-invalidating behaviors in `src/core/hooks/auto-recall.ts`:

1. **Placement** — `<relevant-memories>` emitted as `prependContext` merges into the current user message. On `openai-completions` the block enters conversation history and is subject to per-turn dynamic-budget truncation → user-message prefix shifts every turn.
2. **Non-deterministic memory ordering** — in `searchHybrid`, RRF ties break by the parallel-race order of keyword vs. embedding sub-searches. Same query on adjacent turns can render the same memories in different order.
3. **No repeat-query dedup** — adjacent turns on the same topic re-embed, re-rank, re-render. Score jitter surfaces in the rendered bytes even after (2) is fixed.

Introduces opt-in `recall.cacheSafe`. Defaults preserve legacy behavior.

```ts
recall.cacheSafe = {
  placement: "user-prefix" | "system-tail-dynamic" | "system-tail-cacheable",
  deterministicOrder: boolean,    // sort formatables by (id, type, content) pre-budget
  sessionDedup:       boolean,    // per-session LRU: same query → identical bytes
  sessionDedupMax:    number,     // default 32
  diagnostics:        boolean,    // log SHA-1 fingerprint of each region per turn
};
```

Placement modes:
- `user-prefix` — legacy (default, unchanged).
- `system-tail-dynamic` — new `appendContext` field on `RecallResult`; adapters write it AFTER cache boundary. User message stays cache-stable. Recommended for `openai-completions`.
- `system-tail-cacheable` — folds into `appendSystemContext` (before boundary). Safe only with `sessionDedup=true`.

`deterministicOrder` sorts by `(id, type, content)` BEFORE budget clipping so truncation is also stable.

`sessionDedup` bounded per-session LRU keyed on sanitized user text; repeat returns byte-identical block.

`diagnostics` logs `sha1(region).slice(0,8)` per region per turn for cache-hit forensics without leaking tokens.

### Recommended production config for #120 workload

```yaml
recall:
  cacheSafe:
    placement: system-tail-dynamic
    deterministicOrder: true
    sessionDedup: true
    diagnostics: true
```

## Related Issue | 关联 Issue

Fix #120

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

Local verification:
- `npx vitest run` → 77/77 pass (10 new).
- `npx tsdown` → clean build, 785 kB bundle.
- Defaults produce byte-identical output vs. `main` when `cacheSafe` fields are unset.

New tests:
- `src/config.test.ts` — defaults + valid/invalid placement + all fields.
- `src/core/hooks/auto-recall.test.ts` — `sortMemoriesForCache` order-invariant, non-mutating; `routeRecallPlacement` field routing per mode.

## Additional Notes | 其他说明

- No storage schema change. No breaking change to hook signatures.
- `RecallResult` gains `appendContext?: string`, unset unless `placement=system-tail-dynamic`. Adapters unaware of it still receive legacy fields; migration is one line (concatenate at system tail after boundary).
- Relative to prior discussion under PR #433: this PR additionally addresses the hybrid RRF ordering non-determinism and adds a bounded per-session dedup LRU — both load-bearing for prefix cache stability under score jitter and provider latency races.
- Follow-up (out of scope): wire prompt-shape SHA-1 fingerprints into the existing `agent_turn` metric event.

### DIFF
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -1,0 +1,42 @@
+/**
+ * Config-parse tests for recall.cacheSafe group (issue #120).
+ */
+import { describe, expect, it } from "vitest";
+import { parseConfig } from "./config.js";
+
+describe("parseConfig — recall.cacheSafe", () => {
+  it("defaults are backwards-compatible (legacy user-prefix, no dedup)", () => {
+    const cfg = parseConfig({});
+    expect(cfg.recall.cacheSafe).toEqual({
+      placement: "user-prefix",
+      deterministicOrder: false,
+      sessionDedup: false,
+      sessionDedupMax: 32,
+      diagnostics: false,
+    });
+  });
+
+  it("parses cache-safe settings when provided", () => {
+    const cfg = parseConfig({
+      recall: {
+        cacheSafe: {
+          placement: "system-tail-dynamic",
+          deterministicOrder: true,
+          sessionDedup: true,
+          sessionDedupMax: 8,
+          diagnostics: true,
+        },
+      },
+    });
+    expect(cfg.recall.cacheSafe.placement).toBe("system-tail-dynamic");
+    expect(cfg.recall.cacheSafe.deterministicOrder).toBe(true);
+    expect(cfg.recall.cacheSafe.sessionDedup).toBe(true);
+    expect(cfg.recall.cacheSafe.sessionDedupMax).toBe(8);
+    expect(cfg.recall.cacheSafe.diagnostics).toBe(true);
+  });
+
+  it("rejects unknown placement, falls back to user-prefix default", () => {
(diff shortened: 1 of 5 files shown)


Maxwell-Code07:

Maxwell-Code07:

PR #570, opened 2026-07-23
fix(data): recalibrate checkpoint counters after cleanup (#157)

KarlLeen:
## Summary

Fixes #157. `l0_conversations_count` and `total_memories_extracted` only ever incremented — after any cleanup (memory-cleaner, manual JSONL pruning, pipeline state reset) the counters permanently overstated reality, causing persona generation thresholds to miscalculate and incremental processing to skip records.

### Core fix: `recalibrate()` (`checkpoint.ts`)

Reads actual counts from the live vector store (falling back to JSONL line counts when no store is wired), then corrects `l0_conversations_count`, `total_memories_extracted`, `memories_since_last_persona`, and `scenes_processed` in a single `mutate()` call.

**Tradeoff — `total_processed` is intentionally excluded.** It is a monotonic epoch marker: `last_persona_at` stores a `total_processed` snapshot and persona eligibility is evaluated as `current - last_persona_at > threshold`. Lowering `total_processed` after cleanup would shrink that gap and could re-trigger persona generation immediately. Keeping it monotonic is the safer invariant; the drift from this counter is purely cosmetic (status display), not behavioral.

### Auto-trigger in `memory-cleaner` (`memory-cleaner.ts`)

`runOnce()` calls `recalibrate()` unconditionally at the end of every cleanup cycle.

**Tradeoff — removed the `changedFiles > 0` guard.** An earlier draft only recalibrated when the cycle actually deleted something. But pre-existing drift (from a previous manual prune or a crashed cleanup) must also be corrected, even when the min-retain guard skipped all DB deletes this cycle. The daily run frequency is low enough that the extra I/O is not a concern.

### `memories_since_last_persona` clamp (`checkpoint.ts`)

Old formula: `persona = Math.max(0, persona - deletedCount)` — subtracted the full deletion count regardless of when those records were created relative to the last persona run.

New formula attributes deletions to the correct zone:
- Records deleted **before** the last persona run don't affect the counter (they were already counted as "processed before persona")
- Only records deleted **since** the last persona run reduce the counter

**Tradeoff — more complex formula, but the old one would over-correct.** If 10 records were deleted and 8 were from before the last persona, the old code would reduce `memories_since_last_persona` by 10, potentially triggering persona generation prematurely on the next capture.

### `scenes_processed` recalibration from filesystem (`checkpoint.ts`)

Counts actual `.md` files in `scene_blocks/` instead of trusting the stored value.

**Tradeoff — filesystem read on every recalibrate.** Scene files are not stored in the vector store, so there is no cheaper source of truth. The alternative (keeping `scenes_processed` increment-only) would permanently disable the P3 "first scene" trigger after any scene file is deleted, which is a harder-to-diagnose bug. The `glob` is scoped to one directory and runs at most once per day.

### `drift_history` and `detectDrift()` do not track `scenes_processed` (`checkpoint.ts`)

**Tradeoff — intentional omission.** `l0/l1` drift signals counter corruption: the DB was mutated without the checkpoint being updated, which should not happen in normal operation. `scenes_processed` decreasing after cleanup is expected and correct behavior. Mixing them into the same drift signal would generate noise and make it harder to distinguish real anomalies from routine corrections.

### `last_recalibrated_at` written on every run, not just when `changed=true` (`checkpoint.ts`)

**Tradeoff — "last checked" semantics over "last corrected" semantics.** Writing on every run means operators can verify the cleaner is actually running by inspecting the checkpoint file, even during periods when there is no drift to correct. The downside is the timestamp is less informative about when drift was last found; `drift_history` fills that gap.

### Atomic `floorGlobalCounters()` replaces `checkpoint.write()` in L2 runner (`pipeline-factory.ts`, `checkpoint.ts`)

The original corruption repair read the checkpoint, checked for counter rollback, then called `checkpoint.write({...postState, ...})`. This full overwrite is TOCTOU-unsafe: any `captureAtomically` call between the `read()` and `write()` would be silently clobbered.

**Tradeoff — `floorGlobalCounters()` is technically dead code in normal operation.** With the full `write()` gone, there is no remaining code path that can decrease `total_processed`. The floor guard is kept as a defensive invariant against direct file manipulation or future refactors that reintroduce a full write. The cost is one no-op `mutate()` per L2 extraction.

### `StatusCache<T>` (`status-cache.ts`, `tdai-core.ts`)

TTL-based cache with in-flight promise coalescing for `getCheckpointStatus()`.

**Tradeoff — 30 s default TTL.** Short enough that the displayed status reflects recent cleanup runs; long enough to avoid a file read on every status poll. Callers that need fresh data (e.g., immediately after recalibrate) can call `cache.invalidate()`.

## Test Plan

- [x] `recalibrate()` corrects l0/l1 from vector store and JSONL fallback
- [x] `memories_since_last_persona` clamp: deletions before persona, after persona, mixed zones
- [x] `scenes_processed` recalibrated from `.md` file count on disk
- [x] `drift_history` ring buffer: appends on change, skips when unchanged, evicts at cap=10
- [x] Concurrent `recalibrate()` calls serialized by file lock — no lost writes
- [x] `detectDrift()` returns delta without mutating checkpoint
- [x] `floorGlobalCounters()` raises counters atomically, no-ops when sufficient
- [x] Configurable `minRetainL0/L1` — recalibrate still runs when DB skip fires
- [x] `StatusCache`: hit/miss, TTL expiry, concurrent coalescing, error non-poisoning, invalidate

59 tests pass across `checkpoint.test.ts` and `status-cache.test.ts`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - **checkpoint 计数器只增不减，清理后与真实数据长期漂移** ([#157](https://github.com/TencentCloud/TencentDB-Agent-Memory/issues/157))：`l0_conversations_count` 与 `total_memories_extracted` 此前只在采集/抽取时递增；memory-cleaner 删除过期 L0/L1 后从不回写，导致计数持续高估，并可能让依赖 `memories_since_last_persona` 的画像阈值提前触发。
   - 新增 `CheckpointManager.recalibrate(source?)`：以在线存储（`countL0()` / `countL1()`）为准回填；无存储时回退统计 `conversations/`、`records/` 日分片非空行数。写入走同一把 per-file 锁，幂等，并对 NaN/负数钳制。
   - **画像间隔钳制**：L1 缩水时同步按 delta 下调 `memories_since_last_persona`，并限制其不超过新的 L1 总量，避免清理后误触发 persona。
+  - 新增 `CheckpointManager.resetSession(sessionKey, source?)`：清除单个 session 的 `runner_states` / `pipeline_states` 后立刻 recalibrate，覆盖「删测试 pipeline 状态 / session 重置」漂移路径。
   - **调用时机**：Gateway 启动恢复 pipeline 状态前一次；每轮 cleaner 清理结束后一次；两处均为非致命。
 
 ### ⚠️ 升级注意（仅在显式配置 `timezone` 时生效）
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@
   - 修复 offload local-llm 模式下每次 LLM 调用都重新创建 fetch wrapper 的性能问题（现在在 `LocalLlmClient` 构造函数中创建一次并缓存）。
   - 注入逻辑抽取到 `src/utils/no-think-fetch.ts` 共享，新增 vitest 单测覆盖全部策略 / 跳过 embedding / 非 JSON 容错。
 
+### 🐛 修复
+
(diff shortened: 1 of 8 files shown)


Maxwell-Code07:

Maxwell-Code07:

PR #529, opened 2026-07-18
fix(store): sanitize FTS5 query tokens to prevent MATCH injection (Resolves #160)

Pomeeloo:
问题
buildFtsQuery() 把用户输入拼成 FTS5 MATCH 表达式,含 FTS5 操作符(`" ' * ( ) AND OR NOT NEAR : ^`)的输入会篡改查询语义或触发语法错误(注入)。

方案
- `sanitizeFtsToken()`:双引号包字面 phrase,内部 `"` 转义为 `""`(修正旧实现"删除双引号"导致的 recall 损伤)
- `sanitizeFtsWhitelist()`:字符级白名单(defence-in-depth)
- `buildFtsQuery` 复用 `sanitizeFtsToken`
- 与项目已有的 `MATCH ?` 参数化(SQL 防注入)组成三层防御

测试
新增 38 个:单元(21)+ 真实 FTS5 召回对比(4)+ 白名单(13)。本地全量 **105 passed**,零回归。

 验收对照
issue 四档全覆盖:基础 / 进阶 / 深入(真实 FTS5 recall 实证)/ 拓展(白名单 + 参数化三层防御)。

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [ ] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

PR #585, opened 2026-07-26
Lemons for issue #160

HnuLemons:
# Fix FTS5 query injection in `buildFtsQuery`

## Description | 描述

This PR fixes SQLite FTS5 operator injection in `buildFtsQuery()` (`src/core/store/sqlite.ts`).

`buildFtsQuery()` turns raw user text into an FTS5 `MATCH` expression. The previous
implementation quoted each token and stripped `"`, which partially mitigated injection —
but it did **not** explicitly sanitize FTS5 query syntax, and it forwarded jieba-produced
tokens verbatim. A token such as `content:alpha`, `alpha*`, or a reserved bareword
(`AND`/`OR`/`NOT`/`NEAR`) could therefore reach the `MATCH` expression and either change
search semantics or trigger an FTS5 syntax error — exactly the risk described in the issue.

This PR introduces an explicit, whitelist-based sanitizer and routes **both** the jieba
path and the fallback regex path through it before anything reaches `MATCH`:

- **`sanitizeFtsTokens(token)`** — keeps only Unicode letters (`\p{L}`), Unicode numbers
  (`\p{N}`), and `_`. Every other character — `"`, `'`, `(`, `)`, `*`, `:`, `^`, `-`,
  control chars (NUL/ESC/DEL), emoji, punctuation — is treated as a separator and removed,
  so a mixed token like `foo:bar` is *split* into `foo` and `bar` (preserving recall)
  rather than collapsed.
- **Drop reserved FTS5 bareword operators** — `AND`, `OR`, `NOT`, `NEAR`
  (case-insensitive) — when they survive the whitelist as whole tokens.
- **Quote each sanitized token** as an FTS5 phrase and OR-join the phrases. Since tokens
  are already whitelisted to `[letter|number|_]`, quoting is defense in depth: no
  character can escape the phrase.
- **Apply the same sanitizer to both** jieba segmentation output and fallback regex
  tokenization, so the two paths are uniformly safe.

Example:

```plain
alpha AND NOT beta
```

is now converted to:

```plain
"alpha" OR "beta"
```

rather than letting `AND NOT` control FTS5 query semantics (or, as the executable test
proves, throwing an FTS5 syntax error when forwarded verbatim).

## Related Issue | 关联 Issue

Fixes the FTS5 query sanitization issue described in  [[#160](https://github.com/TencentCloud/TencentDB-Agent-Memory/issues/160)](https://github.com/TencentCloud/TencentDB-Agent-Memory/issues/160 "超链接title")
(SQLite FTS5 operator injection in `buildFtsQuery`).

## Change Type | 修改类型

- Bug fix | Bug 修复
- New feature | 新功能 (exported `sanitizeFtsTokens` whitelist builder)
- Code optimization | 代码优化

### 基础：对 FTS5 特殊字符进行转义，普通用户搜索不受影响

`sanitizeFtsTokens()` strips FTS5 syntax characters and drops reserved operators before
the `MATCH` expression is built. Ordinary keyword search is unchanged:

```plain
travel plan API        → "travel" OR "plan" OR "API"
用户 编程 TypeScript   → "用户" OR "编程" OR "TypeScript"
```

Covered inputs include every FTS5 special character from the issue: `"`, `'`, `(`, `)`,
`*`, `:`, `^`, `-`, plus the reserved operators `AND`, `OR`, `NOT`, `NEAR`.

Test file: `src/core/store/sqlite.fts.test.ts` → `[基础]` (6 tests).

### 进阶：编写完整的安全测试用例，覆盖所有 FTS5 操作符和边界情况

A table-driven suite covers every issue-listed operator and common boundary cases,
asserting both the direct `sanitizeFtsTokens()` output and the end-to-end
`buildFtsQuery()` output (every emitted token must be a safe bareword):

- quote injection: `alpha" OR "beta`, `alpha' OR 'beta`
- parentheses: `(alpha) OR (beta)`
- boolean operators (upper / lower / mixed case): `alpha AND beta`, `alpha OR beta`,
  `alpha NOT beta`, `alpha and beta`, `Alpha Or Beta`
- NEAR syntax: `NEAR(alpha beta, 5)`, `alpha NEAR/5 beta`, `alpha NEAR beta`
- prefix / column-filter-like syntax: `alpha*`, `content:alpha`, `-content:alpha`
- pure operators → nothing survives: `AND OR NOT NEAR`, `and or not near`
- pure punctuation → nothing survives: `!!!`, `"""`, `(*):-`
- mixed alphanumeric + syntax: `foo:bar`, `C++`, `node-18`, `a(b)c`
- control characters (NUL / ESC / DEL) embedded in tokens are dropped
- underscore is kept (part of the whitelist)

Also added an **executable SQLite FTS5** check (via `node:sqlite`) that proves:

- the raw payload `alpha AND NOT beta` is an FTS5 **syntax error** when forwarded
  verbatim (the unsanitized risk);
- a valid boolean injection `alpha NOT beta` matches only the `alpha`-only document
  (boolean effect);
- after sanitization, `alpha AND NOT beta` → `"alpha" OR "beta"` matches **all** of
  A/B/C — i.e. it no longer behaves as a boolean expression;
- every hostile input yields a query that executes in real FTS5 without error.

Test file: `src/core/store/sqlite.fts.test.ts` → `[进阶]` (44 tests).

### 深入：对比转义前后 recall 效果，确认无明显退化

A recall fixture compares, on the **same** in-memory SQLite FTS5 corpus:

- the **previous** behavior simulated by `input.split(/\s+/).filter(Boolean).join(" OR ")`
  (bareword token-OR, no quoting, no sanitization);
- the **new** sanitized quoted-token behavior from `buildFtsQuery()`.

For each ordinary keyword query the two forms return the **same** result set (compared as
sorted `record_id` arrays), confirming no recall regression — while the new form is
strictly safer (every token quoted).

Covered ordinary queries:

```plain
travel plan API
TypeScript memory
coffee beans
project roadmap
用户 编程 TypeScript
```

Test file: `src/core/store/sqlite.fts.test.ts` → `[深入]` (6 tests).

### 拓展：考虑 FTS5 查询白名单或参数化查询方案

Addressed through a **whitelist-based** FTS5 query builder plus verification that
execution stays **parameterized**.

- **Whitelist**: `sanitizeFtsTokens()` is the allow-list defense — only Unicode letters,
  numbers, and `_` may become FTS5 query tokens; everything else is removed before SQLite
  receives the `MATCH` expression.
- **Parameterization**: SQLite execution already uses prepared statements with
  `MATCH ?` (`stmtL1FtsSearch` / `stmtL0FtsSearch`), so the bound value is *data*, not SQL.
  FTS5 still parses that data as a `MATCH` query expression, which is precisely why the
  whitelist is required on top of parameterization. A test verifies that a sanitized query
  containing `"` and ` OR ` binds and executes as data (returns the expected rows) rather
  than being concatenated into SQL.
- **Real `VectorStore` L1/L0 FTS fallback paths**: a temporary SQLite-backed `VectorStore`
  is initialized (sqlite-vec + FTS5 confirmed non-degraded), records are inserted through
  `upsertL1()` / `upsertL0()` (metadata + FTS, no embedding), and `searchL1Fts()` /
  `searchL0Fts()` are queried. The payload `alpha AND NOT beta` returns operator-neutral
  results (A/B/C) on both layers, and ordinary keyword search returns the relevant record.
- **Jieba-path parity**: a fake jieba emitting hostile tokens (`foo:bar`, `C++`, `AND`,
  `NEAR`, …) confirms the same sanitizer is applied to jieba output, not only to the
  fallback regex path.

Test file: `src/core/store/sqlite.fts.test.ts` → `[拓展]` (9 tests).

## Files Changed | 修改文件

- `src/core/store/sqlite.ts`
  - Added exported `sanitizeFtsTokens(token)` — whitelist sanitizer (the 拓展 builder).
  - Added `FTS5_RESERVED_OPERATORS` set.
  - Rewrote `buildFtsQuery()` to route both jieba and fallback tokenization through
    `sanitizeFtsTokens()`, dedupe, and quote each surviving token.
- `src/core/store/sqlite.fts.test.ts` (new) — 4-level test suite (65 tests).

No call-site changes are required: all four callers
(`memory-search.ts`, `conversation-search.ts`, `auto-recall.ts`, `l1-dedup.ts`) already
feed `buildFtsQuery()` output into parameterized `MATCH ?` statements, and the output
format for ordinary input is unchanged (`"a" OR "b"`).

## Self-test Checklist | 自测清单

  ✅ Verified locally | 本地验证通过
  ✅ No existing features affected | 无影响现有功能
  - `buildFtsQuery()` keeps ordinary keyword search behavior
    (`travel plan API` → `"travel" OR "plan" OR "API"`).
  ✅ FTS5 special characters are sanitized
  - Covered `"`, `'`, `(`, `)`, `:`, `*`, `^`, `-`, and pure punctuation input.
  - Verified that syntax-like input is split or removed before constructing the final
    `MATCH` expression.
  ✅ FTS5 reserved operators cannot control query semantics
  - Covered `AND`, `OR`, `NOT`, `NEAR`, lowercase variants, `NEAR(...)`, and `NEAR/5`.
  - Verified that `alpha AND NOT beta` becomes `"alpha" OR "beta"`.
  ✅ Sanitization works for both fallback tokenization and jieba-produced tokens
  - Forced fallback tokenization in tests with `_setJiebaForTest(null)`.
  - Added a fake-jieba test to verify mixed tokens such as `foo:bar`, `C++`, `AND`, and
    `NEAR` go through the same sanitizer.
  ✅ Sanitized queries execute correctly in real SQLite FTS5
  - Created an in-memory FTS5 table with `node:sqlite`.
  - Verified that `alpha AND NOT beta` does not behave as a boolean `AND NOT` expression
    after sanitization (returns A/B/C, not just C).
  - Verified the raw payload throws an FTS5 syntax error (the unsanitized risk).
  ✅ Real `VectorStore` L1 FTS fallback path works
  - Initialized a temporary SQLite-backed `VectorStore` (non-degraded, FTS5 available).
  - Inserted records through `upsertL1()`.
  - Queried through `searchL1Fts()`.
  - Verified operator-neutral results for `alpha AND NOT beta`.
  ✅ Real `VectorStore` L0 FTS fallback path works
  - Inserted records through `upsertL0()`.
  - Queried through `searchL0Fts()`.
  - Verified operator-neutral results for `alpha AND NOT beta`.
  ✅ Recall comparison shows no obvious regression for ordinary keyword queries
  - Compared previous token-OR vs new sanitized quoted-OR on the same SQLite FTS5 fixture.
  - Verified identical result sets for `travel plan API`, `TypeScript memory`,
    `coffee beans`, `project roadmap`, `用户 编程 TypeScript`.
  ✅ Full test suite passes

```plain
$ npx vitest run src/core/store/sqlite.fts.test.ts
 Test Files  1 passed (1)
      Tests  65 passed | 0 failed
```

```plain
$ npm test
 Test Files  5 passed (5)
      Tests  132 passed | 0 failed
```

```plain
$ npm run build
✔ Build complete
```

> **Note:** Some content was generated with AI assistance.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,10 @@
+# 默认忽略的文件
+/shelf/
+/workspace.xml
+# 已忽略包含查询文件的默认文件夹
+/queries/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
+# 基于编辑器的 HTTP 客户端请求
+/httpRequests/
--- a/.idea/TencentDB-Agent-Memory.iml
+++ b/.idea/TencentDB-Agent-Memory.iml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="JAVA_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$" />
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_25" default="true" project-jdk-name="25" project-jdk-type="JavaSDK">
+    <output url="file://$PROJECT_DIR$/out" />
+  </component>
+</project>
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectModuleManager">
+    <modules>
+      <module fileurl="file://$PROJECT_DIR$/.idea/TencentDB-Agent-Memory.iml" filepath="$PROJECT_DIR$/.idea/TencentDB-Agent-Memory.iml" />
+    </modules>
+  </component>
+</project>
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="" vcs="Git" />
+  </component>
+</project>
--- a/src/core/store/sqlite.fts.test.ts
+++ b/src/core/store/sqlite.fts.test.ts
@@ -1,0 +1,522 @@
+/**
+ * Security & recall tests for FTS5 query sanitization (issue #160).
+ *
+ * Covers the four acceptance levels from the issue:
+ *   - 基础  : FTS5 special characters are sanitized; ordinary search unaffected.
+ *   - 进阶  : Complete security suite — every FTS5 operator + boundary cases,
+ *             plus an executable SQLite FTS5 check that a payload such as
+ *             `alpha AND NOT beta` does NOT behave as a boolean expression.
+ *   - 深入  : Recall comparison (old token-OR vs new sanitized quoted-OR) on the
+ *             same SQLite FTS5 fixture — confirms no recall regression for
+ *             ordinary keyword queries.
+ *   - 拓展  : Whitelist-based builder + parameterized `MATCH ?` verification,
+ *             plus real `VectorStore` L1/L0 FTS fallback paths.
+ */
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it } from "vitest";
+import { DatabaseSync } from "node:sqlite";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+
+import type { L0Record } from "./types.js";
+import type { MemoryRecord } from "../record/l1-writer.js";
+import {
+  _resetJiebaForTest,
+  _setJiebaForTest,
+  buildFtsQuery,
+  sanitizeFtsTokens,
+  VectorStore,
+} from "./sqlite.js";
+
+// ──────────────────────────────────────────────────────────────────────────
+// Helpers
+// ──────────────────────────────────────────────────────────────────────────
+
+/** Fake jieba that emulates `cutForSearch` for deterministic ASCII tests. */
+function fakeJieba(tokens: string[]): Parameters<typeof _setJiebaForTest>[0] {
+  return {
+    cutForSearch: (_text: string, _hmm: boolean) => tokens,
+  } as Parameters<typeof _setJiebaForTest>[0];
+}
+
+/** All FTS5 query-syntax characters that must never reach MATCH as live ops. */
+const FTS5_SYNTAX_CHARS = ['"', "'", "(", ")", "*", ":", "^", "-"];
+
+/** A bare token is "safe" iff it contains only letters / numbers / `_`. */
+function isSafeToken(t: string): boolean {
+  return /^[\p{L}\p{N}_]+$/u.test(t);
+}
+
+/** Build a fresh in-memory FTS5 table for isolated execution tests. */
+function makeFtsTable(docs: Array<{ id: string; text: string }>): {
+  db: DatabaseSync;
+  search: (matchExpr: string) => string[];
+  close: () => void;
+} {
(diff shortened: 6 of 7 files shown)


Maxwell-Code07:

Maxwell-Code07:
感谢你的参与！本次犀牛鸟活动 issue [#160] 最终方案已选定，通过 PR #529 合并。你的提交质量很不错，可惜本次没有纳入本次版本。期待你后续继续参与犀牛鸟开源课题，欢迎继续提交贡献👏

PR #586, opened 2026-07-26
fix(hermes): bound Gateway sidecar logs

### LINKED ISSUE #583: [Bug] Hermes Gateway stdout/stderr logs grow without runtime bounds

### Plugin Version | 插件版本

### Operating System | 操作系统

macOS / Darwin 24.6.0 arm64

### Describe the bug | 问题描述

### To Reproduce | 复现步骤

### Expected behavior | 预期行为

- guard interval: 250 ms

```text
Task completed: flush
```

Observed storage impact:

```text
line count: 420,160,119
log directory: ~63 GiB
```

### Additional context | 补充信息

Heptazero:
## Description | 描述

Bound the Hermes-managed Gateway's append-only stdout/stderr logs at runtime.

- guard each active log every 250 ms;
- when a file exceeds 10 MiB, atomically preserve its final 1 MiB as `<path>.1`;
- truncate through the existing parent handle so the inode inherited by the child remains live;
- bound leftovers from previous runs before spawning a new child;
- stop/join the guard before closing handles or reading a startup-crash tail;
- fail open on filesystem errors, warning once per failing path without interrupting child logging.

The regression tests use real files and a real child process rather than mocked `Popen` output. Their A/B case writes about 1 MiB:

- control (no guard): grows beyond 10× the 32 KiB test threshold;
- experiment (guard enabled): rotates while the child is still writing, leaves the active file at or below 32 KiB, retains exactly 8 KiB in `.1`, preserves the active inode, and captures the child's final sentinel.

Production defaults remain the issue's proposed 10 MiB active limit, 1 MiB retained tail, and 250 ms interval.

## Related Issue | 关联 Issue

Fixes #583

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证

- `python3 hermes-plugin/memory/memory_tencentdb/tests/test_gateway_log_bounds.py` — 4/4 passed
- `python3 -m py_compile .../supervisor.py .../test_gateway_log_bounds.py` — passed
- `npm test` — 67/67 passed
- `npm run build` — passed
- `npm pack --dry-run --ignore-scripts --json` — passed; no `__pycache__`/`.pyc` artifacts
- broader Hermes pytest probe: 6 passed, 3 skipped, and one pre-existing `test_second_provider_does_not_reuse_stale_gateway` failure reproduced unchanged on `origin/main`

## Additional Notes | 其他说明

The active path is deliberately not renamed during rotation: renaming alone would leave the Gateway's inherited descriptor writing to the old inode and would not enforce a runtime disk bound.

### DIFF
--- a/hermes-plugin/memory/memory_tencentdb/supervisor.py
+++ b/hermes-plugin/memory/memory_tencentdb/supervisor.py
@@ -13,8 +13,9 @@ import os
 import shlex
 import signal
 import subprocess
+import threading
 import time
-from typing import IO, Optional
+from typing import IO, Optional, Set
 
 from .client import MemoryTencentdbSdkClient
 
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Thank you for your attention. We will schedule an internal review of your PR as soon as possible and get back to you with feedback.

PR #607, opened 2026-07-28
fix: sanitize FTS5 operators in buildFtsQuery to prevent query injection

RerankerGuo:
## Summary

This PR fixes #160 by sanitizing FTS5 operator characters in `buildFtsQuery()` to prevent query injection.

## Changes
- Added `sanitizeFtsToken()` function to strip all FTS5 special characters from user-supplied tokens
- Added `FTS5_SPECIAL_CHARS` regex covering: *, ^, (, ), :, -, +, /, \, !
- Applied sanitization to all tokens before building the FTS5 query
- Rejects tokens that become empty after sanitization (no alphanumeric content)
- Added proper TypeScript type narrowing for the sanitized token filter

## Problem
The `buildFtsQuery()` function only stripped double quotes from tokens, but did not handle other FTS5 operator characters. Users could inject characters like `*` (prefix), `^` (weight), `(` `)` (grouping), `:` (column restrict), `-` (boolean NOT), or `+` (boolean AND) into search queries, altering the query semantics.

## Solution
Added a dedicated sanitization step that strips all FTS5 special characters from each token before wrapping them in double-quoted phrases. This ensures that even if a user includes FTS5 operator characters in their input, they are neutralized before being embedded in the MATCH query.

## Security Considerations
- The fix is defense-in-depth: tokens were already quoted (which neutralizes most operators), but this adds explicit stripping for belt-and-suspenders safety
- The sanitization is applied after tokenization, so it only affects the final token set
- Tokens that become entirely empty after sanitization are rejected, preventing empty phrases

## Checklist
- [x] My code follows the project's style guidelines
- [x] The fix is minimal and focused on the issue
- [x] Comprehensive coverage of all FTS5 special characters
- [x] Proper TypeScript typing for the filter

### DIFF
--- a/src/core/store/sqlite.ts
+++ b/src/core/store/sqlite.ts
@@ -189,6 +189,58 @@ const ZH_STOP_WORDS = new Set([
  * match more tokens higher, so precision is preserved while recall is
  * significantly improved — especially for longer queries and when running
  * in FTS-only fallback mode (no embedding available).
+ */
+
+/**
+ * Characters that have special meaning in FTS5 query syntax.
+ * Must be stripped from user-supplied tokens to prevent query injection.
+ *
+ * FTS5 special characters:
+ * - *  prefix operator (e.g. "hello*")
+ * - ^  weight operator (e.g. "hello"^2)
+ * - (  )  grouping operators
+ * - "  phrase delimiter
+ * - :  column restrict (e.g. title:hello)
+ * - -  boolean NOT (e.g. -hello)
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:

RerankerGuo:
Closing as superseded by the approved canonical #178. #178 is the narrower FTS5 operator-sanitization fix, has been locally verified by the reviewer, and has all checks passing.

PR #608, opened 2026-07-28
fix(search): sanitize FTS5 operators in buildFtsQuery

heguang-mercy:
## Summary

`buildFtsQuery()` constructs an FTS5 MATCH expression from user-provided text by tokenizing and OR-joining tokens. It does not strip FTS5 special operators (AND, OR, NOT, NEAR) from the input before tokenization, allowing users to alter query semantics.

## Root Cause

In `src/core/store/sqlite.ts`, `buildFtsQuery()` wraps each token in double quotes, which mitigates most injection vectors. However, the lack of explicit sanitization of FTS5 operators before tokenization is a defense-in-depth gap.

## Fix

Strip FTS5 operators (AND, OR, NOT, NEAR) from the raw input string before tokenization, on both the jieba path and the fallback regex path.

### Before


Maxwell-Code07:

PR #612, opened 2026-07-28
fix: prevent SerialQueue deadlock on synchronous task throw

### LINKED ISSUE #518: Bug: SerialQueue deadlocks after a synchronous task throw

RerankerGuo:
## Description

This PR fixes issue where `SerialQueue` deadlocks when a task throws synchronously (before returning a Promise).

## Related Issue

Fixes #518

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Code optimization

## Root Cause

In `src/utils/serial-queue.ts`, the `drain()` method called `entry.task()` directly in a `.then()` chain. If a synchronous task throws before returning a Promise, the error bypasses the `.catch()` handler, leaving `running=true` forever and blocking the queue.

## Fix

Wrapped task execution in `Promise.resolve().then(() => entry.task())` so synchronous throws are captured by the Promise chain and properly routed to `.catch()` and `.finally()`.

## Changes

- **`src/utils/serial-queue.ts`**: Wrapped `entry.task()` call in `Promise.resolve().then()`
- **`src/utils/serial-queue.test.ts`**: New test suite with 4 regression tests for sync-throw scenarios

## Self-test Checklist

- [x] TypeScript diagnostics pass
- [x] Regression tests for all deadlock scenarios

### DIFF
--- a/src/utils/serial-queue.test.ts
+++ b/src/utils/serial-queue.test.ts
@@ -1,0 +1,166 @@
+import { describe, it, expect } from "vitest";
+import { SerialQueue } from "./serial-queue.js";
+
+describe("SerialQueue", () => {
+  describe("basic execution", () => {
+    it("runs tasks in FIFO order", async () => {
+      const q = new SerialQueue("test");
+      const order: number[] = [];
+
+      await Promise.all([
+        q.add(async () => order.push(1)),
+        q.add(async () => order.push(2)),
+        q.add(async () => order.push(3)),
+      ]);
+
+      expect(order).toEqual([1, 2, 3]);
+    });
+
+    it("resolves with task return value", async () => {
+      const q = new SerialQueue("test");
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Thank you for your attention! We will review your code and get back to you as soon as possible.

RerankerGuo:
Closing in favor of the earlier community implementation #519. #519 uses the same Promise.resolve().then(() => task()) fix, includes the exact sync-throw continuation/onIdle regression test, has all checks passing, and has already received maintainer acknowledgement. Keeping #612 would only duplicate review work.

Issue #549, opened 2026-07-22
L1 提炼链路有两个相关的健壮性缺陷，叠加后会导致 L1 提炼静默停摆数天，而 L0 捕获一切正常，用户难以察觉


qiaoyuanjun:
### OpenClaw Version | OpenClaw 版本

gateway部署

### Plugin Version | 插件版本

1.0.1

### Operating System | 操作系统

windows11

### System Specification | 系统配置

13th Gen Intel(R) Core(TM) i9-13900HX (2.20 GHz)

### Describe the bug | 问题描述

Two related pipeline-robustness defects in the standalone/stateful L1 extraction path. Together they can silently stall L1 extraction for days while capture (L0) keeps working.

L1 提炼链路有两个相关的健壮性缺陷，叠加后会导致 L1 提炼静默停摆数天，而 L0 捕获一切正常，用户难以察觉。

#### 缺陷 1：L1 drain 任务锁冲突超时后被直接丢弃，drain 链永久断裂

**Mechanism | 机制**

When a backlog exists, L1 extraction continues via a self-enqueue chain: after each batch, the worker executor checks `result.hasFullBacklog` and calls `enqueueL1Drain` for the next batch (`src/gateway/server.ts:1479-1487`). There is no supervisor — the next batch exists only because the previous one enqueued it.

积压消化靠"上一批入队下一批"的接力模型，没有任何监督或重试机制。

**Where it breaks | 缺陷位置**

`src/services/pipeline-worker.ts:302-343`: on lock conflict the worker retries with exponential backoff (200ms→5s cap) until `lockTtlMs` (~2 min) is exhausted, then:

```
Lock conflict timeout [L1] (task=L1-drain-...): pipeline:{default}:<session>, dropping task
```

The task is dropped **and ACKed** (the "CR-1 fix" comment, :335-341) — never re-enqueued, never dead-lettered, no alert. Since the drain chain is the only way a full backlog continues, one drop silently ends the whole drain.

锁冲突超时时任务被丢弃并 ACK（防重投），不重入队、不进死信、无告警。而 `hasFullBacklog` 分支故意不设置 `L1_idle` 定时器（:1475-1476 注释假设 drain 任务一定会跑），断链后只能靠"新对话凑满 everyNConversations"这一条路恢复。

**Why lock conflicts are the norm, not the exception | 锁冲突为何必然发生**

[...]

PR #621, opened 2026-07-29
fix: add runtime log rotation to prevent unbounded log growth

### LINKED ISSUE #583: [Bug] Hermes Gateway stdout/stderr logs grow without runtime bounds

RerankerGuo:
## Background

This PR addresses #583: the Hermes GatewaySupervisor opens `gateway.stdout.log` and `gateway.stderr.log` in append mode with no runtime size limit. A high-frequency logging defect (e.g. the flush loop in #549) can grow these files without bound — one observed incident reached 63 GiB / 420 billion lines in 4 days.

Closes #583

### `hermes-plugin/memory/memory_tencentdb/supervisor.py`

**New constants:**
- `LOG_MAX_BYTES = 10 MiB` — hard size limit per log stream
- `LOG_PRESERVE_BYTES = 1 MiB` — diagnostic tail preserved on rotation
- `LOG_ROTATE_INTERVAL = 2.0s` — guard check interval

**New methods:**
- `_start_log_rotator()` — starts a daemon thread after log files are opened
- `_stop_log_rotator()` — signals the thread to stop and waits for clean exit
- `_log_rotator_loop()` — periodic check loop (every `LOG_ROTATE_INTERVAL`)
- `_check_log_rotation(handle, path)` — checks size and triggers rotation if needed
- `_rotate_log_file(handle, path, current_size)` — performs in-place rotation:
  1. Reads last `LOG_PRESERVE_BYTES` from the active file
  2. Writes preserved tail to `.1` backup (e.g. `gateway.stdout.log.1`)
  3. Truncates active file in-place via `os.ftruncate(fd, 0)`
  4. Writes preserved tail back to the beginning
  5. Seeks to end for continued append

**Modified methods:**
- `ensure_running()` — calls `_start_log_rotator()` after opening log files
- `_close_log_handles()` — calls `_stop_log_rotator()` before closing handles
- Constructor — adds `_log_rotator` thread and `_log_rotator_stop` event

## Design Rationale

**In-place truncation via `os.ftruncate()`**: The child Gateway process inherits the file descriptors. Renaming the log file would leave the child writing to the old inode, creating a "phantom" log. In-place truncation keeps the fd valid.

**Daemon thread**: The rotator thread is daemonized so it never blocks process exit. It also stops cleanly before log handles are closed to avoid a race.

**Preserved tail**: Writing the last 1 MiB to a `.1` file provides crash diagnostics when the active log has been truncated.

## What this does NOT change
- Log file paths, formats, or naming conventions
- Child process stdout/stderr redirection mechanism
- Gateway startup/shutdown sequence
- The health check or watchdog logic

## Verification
- [x] Python syntax validated: `ast.parse()` passes
- [x] Rotator thread is properly started/stopped with log handle lifecycle
- [x] In-place truncation preserves child's fd validity
- [x] Backup .1 file preserves diagnostic tail for postmortem analysis

## Impact
- **Breaking Change?** → No
- **Files changed**: 1 (`hermes-plugin/memory/memory_tencentdb/supervisor.py`)
- **Risk**: Low — new background thread, log-only change, no functional impact on Gateway behavior

### DIFF
--- a/hermes-plugin/memory/memory_tencentdb/supervisor.py
+++ b/hermes-plugin/memory/memory_tencentdb/supervisor.py
@@ -13,6 +13,7 @@ import os
 import shlex
 import signal
 import subprocess
+import threading
 import time
 from typing import IO, Optional
 
@@ -20,17 +21,18 @@ from .client import MemoryTencentdbSdkClient
 
 logger = logging.getLogger(__name__)
 
-# Default Gateway address
 DEFAULT_HOST = "127.0.0.1"
 DEFAULT_PORT = 8420
 
-# Health check parameters
-HEALTH_CHECK_INTERVAL = 0.5  # seconds between checks
-HEALTH_CHECK_MAX_WAIT = 30   # max seconds to wait for Gateway to start
-HEALTH_CHECK_RETRIES = 3     # retries for is_running check
+HEALTH_CHECK_INTERVAL = 0.5
+HEALTH_CHECK_MAX_WAIT = 30
+HEALTH_CHECK_RETRIES = 3
 
-# Log file rotation parameters
-LOG_TAIL_BYTES_ON_CRASH = 2048  # bytes of stderr log to surface on startup crash
+LOG_TAIL_BYTES_ON_CRASH = 2048
+
+LOG_MAX_BYTES = 10 * 1024 * 1024
+LOG_PRESERVE_BYTES = 1 * 1024 * 1024
+LOG_ROTATE_INTERVAL = 2.0
 
 
 class GatewaySupervisor:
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:
Thank you for your attention! We will review your code and get back to you as soon as possible.

RerankerGuo:
Closing in favor of the earlier community implementation #586. #586 addresses the same Gateway sidecar log-bounding problem and includes dedicated runtime-bound regression tests, so retaining #621 would duplicate review work.

PR #638, opened 2026-07-30
fix(checkpoint): recalibrate append-only counters on startup (fixes #157)

liu66-qing:
## Description | 描述

Fixes the drift bug reported in #157: `total_memories_extracted` and `l0_conversations_count` in `recall_checkpoint.json` only ever increment. After any cleanup (`memory-cleaner` pruning JSONL shards, SQLite `deleteL0Expired`, or manual edits to the checkpoint), the counters permanently overstate reality. That biases L2/L3 persona-threshold logic (`memories_since_last_persona`, `total_processed`) and misleads every status readout that surfaces them.

Adds `CheckpointManager.recalibrate(dataDir, store?)` and calls it once on `TdaiCore` startup (fire-and-forget, after `storeReady`).

### Design decisions worth reviewer attention

1. **Downward-only correction.** A recount larger than the stored value is *ignored*. Otherwise a shard that was written to disk before a crash but never got a checkpoint flush would leave `records/` transiently ahead of the counter; if we raised the counter to match, the pending `markL1ExtractionComplete` that was queued in memory would double-count when it eventually lands.
2. **MAX(jsonl, dbL1) for L1.** JSONL is the persistent source-of-truth per `l1-writer.ts`, but the vector store deletes eagerly on update/merge while JSONL is append-only. When the two disagree, taking the max avoids clobbering the healthier side.
3. **Structural `CheckpointStoreProbe`.** A tiny interface exposing only `countL0` / `countL1`, so callers pass `IMemoryStore` directly without dragging `core/store/*` into `utils/checkpoint.ts` (that would introduce a circular import — `checkpoint` is imported everywhere the store lives).
4. **Clamp `memories_since_last_persona`** to the new L1 ceiling so persona triggers can't fire on phantom counts after pruning.
5. **Recount outside the file lock, correction inside.** Walking JSONL can be slow on large installs; we don't want to block concurrent `captureAtomically` / `markL1ExtractionComplete` on that. The write phase acquires the existing per-file lock as usual.
6. **Fire-and-forget from `initialize()`.** Best-effort. Any error is logged as `warn` and never blocks or fails startup — the checkpoint is still usable in its pre-recount state if recalibration errors.

### Why not the alternatives

- **Recount unconditionally (grow too):** rejected — see (1) above.
- **JSONL-only, drop the probe:** rejected — L0 lives only in SQLite, so `l0_conversations_count` would never recalibrate.
- **Recount lazily on every read:** rejected — breaks the append-only semantics the pipeline relies on and adds a hot-path cost.

## Related Issue | 关联 Issue

Closes #157

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

### Verification commands

```
npx vitest run                # 91 pass, 0 fail (12 new + 79 existing)
npx tsc --noEmit              # no errors
npx tsdown --dts=false        # build clean
```

### Tests added (12 cases in `src/utils/checkpoint.test.ts`)

- **Exact issue #157 repro** — 50 accumulated → operator prunes 8 pipeline_states + 8 JSONL lines → recalibrate corrects to 42, and the operator's remaining `pipeline_states` are preserved.
- Downward correction (50 → 42).
- No-grow guarantee (JSONL=100, stored=42 → stays at 42).
- L0 probe path (stored=50, probe=8 → 8).
- Persona clamp (total 50→3, `memories_since_last_persona` 25 → 3).
- Persona left alone when already within bounds.
- `MAX(jsonl, dbL1)` tie-break (jsonl=10, dbL1=40 → 40).
- Malformed / blank JSONL lines skipped.
- Missing `records/` treated as zero.
- Throwing store probe falls back to JSONL-only.
- Idempotence (twice yields the same result).
- Concurrent-mutation safety (`recalibrate` + `markL1ExtractionComplete` serialized correctly).

## Additional Notes | 其他说明

- Commit signed off per CONTRIBUTING.md DCO requirement.
- `CHANGELOG.md` updated under `[Unreleased] → 🐛 修复 → 数据一致性`.
- No public API break — `recalibrate()` is additive, existing callers unchanged.
- Runtime cost: one directory scan + one JSON parse per JSONL line at boot, bounded by the current shard count (default retention keeps this small). Zero cost after startup — nothing else calls into it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,9 @@
 
 ### 🐛 修复
 
+#### 数据一致性
+- **Checkpoint 计数器只增不减，cleanup 后与真实数据漂移** ([#157](https://github.com/TencentCloud/TencentDB-Agent-Memory/issues/157))：`recall_checkpoint.json` 中的 `total_memories_extracted` 和 `l0_conversations_count` 仅在 L0 capture / L1 完成时递增，`memory-cleaner` 清理 JSONL 分片或 SQLite 过期行、以及手工编辑 checkpoint 后从不下调；结果计数长期偏高，导致 `memories_since_last_persona` / `total_processed` 触发 L3 persona 生成阈值时机错乱，状态查询也持续高报。新增 `CheckpointManager.recalibrate(dataDir, store?)`：JSONL (`records/*.jsonl`) 为 L1 数据源、可选 store probe (`countL0` / `countL1`) 为 L0 数据源；`MAX(jsonl, dbL1)` 容忍两侧未同步；**只下调不上调**（避免 crash 前未 flush 的写入落盘后被双计）；对 `memories_since_last_persona` 做 clamp 防止 persona 幻触发。`TdaiCore.initialize()` 中在 `storeReady` 后 fire-and-forget 调用一次，best-effort，永不阻塞启动。12 组 vitest 覆盖 issue 复现、downward 修正、no-grow 保护、probe 抛错回退、malformed JSONL 跳过、并发写入锁。
+
 #### 数据安全 / 数据隔离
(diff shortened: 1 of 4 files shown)


Maxwell-Code07:

Maxwell-Code07:

PR #507, opened 2026-07-15
fix(seed): flush pending L1 batches

WHUTcjh-2024:
﻿## Description | 描述
Fix `/seed` hanging during historical imports and ensure tail rounds are flushed through L1.

The seed runtime now uses explicit L1 flushes at seed batch/session boundaries instead of polling `waitForL1Idle`. Seed runs also disable live warmup behavior so historical rounds are processed according to the configured `everyNConversations` batching semantics.

`flushSession()` now treats DB-backed pending conversations as L1 work even when the in-memory message buffer is empty, matching the current design where L1 reads persisted L0 records.

## Related Issue | 关联 Issue
Fix #505

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [ ] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

## Additional Notes | 其他说明
Validation:
- `./node_modules/.bin/vitest.CMD run src/core/seed/seed-runtime.test.ts src/utils/pipeline-manager.test.ts`
- `./node_modules/.bin/vitest.CMD run`
- `./node_modules/.bin/tsdown.CMD`
- `git diff --check`

PR #645, opened 2026-07-31
fix(seed): wait for L2/L3 before closing pipeline

RerankerGuo:
## Description | 描述

Fix the remaining seed lifecycle defect from #115: the seed runtime previously waited only for L1 to become idle, then called the gateway-oriented `destroy()` path. That path has an intentional 2-second shutdown budget and marks the scheduler destroyed before flushing, so a scheduled/running L2 extraction or the L3 persona task triggered by L2 could be skipped or interrupted before artifacts reached disk.

This PR adds a finite-batch completion barrier:

- Add `MemoryPipelineManager.flushAll()` to drain pending L1 work, flush scheduled L2 work, then wait for L3.
- Wait for L2 and L3 **sequentially**. The old concurrent wait could observe an initially-idle L3 queue before L2 enqueued the L3 task.
- Cancel recurring L2 timers after the finite batch drain, preventing the following `destroy()` call from triggering a duplicate L2 pass.
- Replace seed's final L1-only polling barrier with `flushAll()` so scene and persona artifacts are complete before storage resources close.
- Add a regression test with independently blocked L2 and L3 runners. It proves the barrier does not resolve during either phase and that destroy does not run a second extraction afterward.

Scope boundary: this does not change the JSON correction/retry behavior already covered by #190, or the seed tail-batch flush behavior covered by #507.

## Related Issue | 关联 Issue

Refs #115

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Validation (Node v26.5.0, npm v11.17.0):

```text
npm test -- src/utils/pipeline-manager.test.ts
# 1 file, 1 test passed

npm test
# 5 files, 68 tests passed

npm run build
# plugin bundle + all three script TypeScript builds passed

git diff --check
# passed
```

No configuration, storage schema, or public HTTP API changes.

### DIFF
--- a/src/core/seed/seed-runtime.ts
+++ b/src/core/seed/seed-runtime.ts
@@ -5,15 +5,9 @@
  * L1 runner, L2 runner, L3 runner, and persister wiring — keeping this
  * module focused on seed-specific concerns:
  * - Synchronous per-round L0 capture with progress reporting
- * - waitForL1Idle polling (L1 only — see FIXME below)
+ * - Per-batch L1 idle polling to preserve extraction boundaries
+ * - Final L1→L2→L3 drain before storage resources are closed
  * - Ctrl+C graceful shutdown
- *
- * FIXME: Currently we only wait for L1 to become idle before destroying the
- * pipeline.  L2 (scene extraction) and L3 (persona generation) may still be
- * in-flight when `pipeline.destroy()` is called.  This is intentional for now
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

PR #646, opened 2026-07-31
fix(capture): skip pipeline scheduling for empty L0 batches

RerankerGuo:
## Background

`performAutoCapture()` currently notifies `MemoryPipelineManager` whenever a
scheduler is present, even when checkpoint filtering recorded zero new L0
messages.

This happens for duplicate hook delivery and is also the expected shape of a
checkpoint-based seed replay. The extra notification increments
`conversation_count`, advances L1 thresholds, and schedules extraction for work
that does not exist.

Refs #115.

## Changes

- Require `filteredMessages.length > 0` before calling
  `scheduler.notifyConversation()`.
- Keep the existing no-scheduler behavior unchanged.
- Return `schedulerNotified: false` and emit a debug-level reason for an empty
  filtered batch.
- Add a regression test that captures the same timestamped round twice and
  verifies:
  - the first call records two messages and notifies once;
  - the duplicate call records zero messages and does not notify;
  - checkpoint `l0_conversations_count` and `total_processed` remain stable.

### Why this approach

The checkpoint already decides whether the invocation produced new durable L0
data. Reusing that result at the scheduler boundary is narrower and safer than
adding duplicate detection inside `MemoryPipelineManager`.

### Out of scope

- No change to L0 filtering or cursor semantics.
- No change to L1/L2 thresholds for real captured conversations.
- No seed CLI resume option in this PR.

## Verification

- `npm test -- src/core/hooks/auto-capture.test.ts` — 1 passed
- `npm test` — 68 passed
- `npm run build` — passed
- `git diff --check` — clean

## Impact

- **Breaking change:** No.
- **Affected path:** L0 capture to pipeline scheduling boundary.
- **Performance:** Avoids unnecessary no-op L1 scheduling.
- **Persistence:** No schema or file-format changes.

## Checklist

- [x] Change is limited to one behavioral guard.
- [x] Regression test covers duplicate checkpoint filtering.
- [x] Existing tests pass.
- [x] Full build passes.
- [x] No unrelated changes included.

### DIFF
--- a/src/core/hooks/auto-capture.test.ts
+++ b/src/core/hooks/auto-capture.test.ts
@@ -1,0 +1,74 @@
+import { mkdtemp, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { parseConfig } from "../../config.js";
+import { CheckpointManager } from "../../utils/checkpoint.js";
+import type { MemoryPipelineManager } from "../../utils/pipeline-manager.js";
+import type { Logger } from "../types.js";
+import { performAutoCapture } from "./auto-capture.js";
+
+const logger: Logger = {
+  debug() {},
+  info() {},
+  warn() {},
+  error() {},
+};
+
+describe("performAutoCapture", () => {
+  const tempDirs: string[] = [];
+
+  afterEach(async () => {
+    await Promise.all(
+      tempDirs.splice(0).map((dir) => rm(dir, { recursive: true, force: true })),
+    );
+  });
+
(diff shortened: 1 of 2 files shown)


RerankerGuo:
Closing after a deeper recovery-path review. The unconditional scheduler notification is also a recovery signal for the crash window where L0 was durably captured but the process exited before notifyConversation. A duplicate delivery with zero newly filtered messages can still be the event that replays that pending L0 through L1. Skipping the notification based only on filteredMessages.length could strand that data until an unrelated future conversation. This guard therefore needs an explicit persisted-pending-work signal, not just an empty-batch check. I am withdrawing this version rather than weakening recovery semantics.

PR #647, opened 2026-07-31
fix(seed): wait for L2/L3 before closing v2 pipeline

RerankerGuo:
## Description | 描述

Port the seed completion fix from the legacy `main` line (#645) to the repository's current default `feat/server_team` / `MemoryCore` line. These branches have unrelated histories, and both contain the same L1-only seed FIXME, so the v2 package needs its own equivalent change.

Before this PR, `MemoryCore/src/core/seed/seed-runtime.ts` waited only for L1 idle and then called the gateway-oriented `destroy()` path. That path has an intentional 2-second shutdown budget and marks the scheduler destroyed before flushing, so scheduled/running L2 scene extraction or the L3 persona task triggered by L2 could be skipped or interrupted before artifacts reached disk.

Changes:

- Add `MemoryPipelineManager.flushAll()` to drain pending L1 work, flush scheduled L2 work, then wait for L3.
- Wait for L2 and L3 **sequentially**. A concurrent wait can observe an initially-idle L3 queue before L2 enqueues its L3 task and return too early.
- Cancel recurring L2 timers after a finite batch drain so the following `destroy()` does not execute an unnecessary second L2 pass.
- Replace seed's final L1-only barrier with `flushAll()` before storage teardown.
- Add a regression test with independently blocked L2 and L3 runners, proving the barrier waits for both and destroy does not trigger a duplicate extraction.

Scope boundary: no Gateway request behavior, store schema, stateful worker scheduling, or LLM extraction logic changes.

## Related Issue | 关联 Issue

Refs #115

Related implementation for the legacy `main` line: #645

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Validation in `MemoryCore` (Node v26.5.0, npm v11.17.0):

```text
npm test -- src/utils/pipeline-manager.test.ts
# 1 file, 1 test passed

npm test
# 1 file, 1 test passed (the branch currently contains this test file)

npm run build:plugin
npm run build:migrate-sqlite-to-vdb
npm run build:export-tencent-vdb
npm run build:read-local-memory
# all passed

git diff --check
# passed
```

Full `npm run build` reaches and passes the plugin plus the three existing script builds, then fails at the upstream baseline command `build:seed-v2` because `MemoryCore/scripts/seed-v2/tsconfig.json` is absent on `feat/server_team`. This PR does not add that unrelated missing build file.

### DIFF
--- a/MemoryCore/src/core/seed/seed-runtime.ts
+++ b/MemoryCore/src/core/seed/seed-runtime.ts
@@ -5,15 +5,9 @@
  * L1 runner, L2 runner, L3 runner, and persister wiring — keeping this
  * module focused on seed-specific concerns:
  * - Synchronous per-round L0 capture with progress reporting
- * - waitForL1Idle polling (L1 only — see FIXME below)
+ * - Per-batch L1 idle polling to preserve extraction boundaries
+ * - Final L1→L2→L3 drain before storage resources are closed
  * - Ctrl+C graceful shutdown
- *
- * FIXME: Currently we only wait for L1 to become idle before destroying the
- * pipeline.  L2 (scene extraction) and L3 (persona generation) may still be
- * in-flight when `pipeline.destroy()` is called.  This is intentional for now
- * to avoid excessively long seed runs, but means seed output may not include
- * the latest L2/L3 artifacts.  Re-evaluate adding a full L1+L2+L3 idle wait
- * once pipeline-manager exposes reliable L2/L3 idle signals.
  */
 
 import path from "node:path";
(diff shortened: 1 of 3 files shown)


RerankerGuo:
Closing this parallel v2 port to keep the current contribution batch within the two-PR limit and avoid asking maintainers to review the same fix twice. PR #645 is the canonical, CI-covered implementation on the branch currently used by all recent community PRs. The MemoryCore port is preserved on  and can be reopened/rebased when the protected  line has an active PR CI path and its current missing  build baseline is resolved.

RerankerGuo:
Correction to the branch references stripped by shell quoting above: the preserved port is on RerankerGuo:fix/issue-115-seed-drain-l2-l3-v2. It targets feat/server_team. Its full build currently reaches the pre-existing missing MemoryCore/scripts/seed-v2/tsconfig.json path; the focused pipeline-manager regression test passes.

RerankerGuo:
Reopening this v2-line port now that the local Auto-Pilot workflow no longer has a fixed PR-count cap. This remains an independent implementation for the unrelated `feat/server_team` history (the `main` implementation is #645). Full v2 build validation depends on #652, which removes the pre-existing missing `scripts/seed-v2/tsconfig.json` target; the focused regression test and all existing build stages before that baseline failure already pass.

Maxwell-Code07:

RerankerGuo:
Closing again as the cross-branch duplicate of #645. #645 contains the same seed drain fix against main and has all four required checks passing. This feat/server_team port has no PR checks and should only be reconsidered after that branch has an independent validated need.

PR #648, opened 2026-07-31
feat(seed): support explicit checkpoint resume

RerankerGuo:
## Background

The seed command already persists per-session capture and L1 cursors in
`.metadata/checkpoint.json`, but it unconditionally rejects every output
directory containing a checkpoint. An interrupted historical import therefore
cannot reuse its durable progress and must start from an empty directory.

This PR adds a conservative, explicit resume mode. It replays the full input
through the existing cursor-based capture path, which skips already captured
timestamps and continues with appended rounds.

Refs #115.

## Changes

- Add `--resume` to `openclaw memory-tdai seed`.
- Require all resume preconditions before starting:
  - an explicit `--output-dir`;
  - an existing output directory;
  - an existing `.metadata/checkpoint.json`;
  - stable timestamps already present on every input message.
- Preserve the existing fail-safe behavior without `--resume`: checkpointed and
  non-empty directories remain rejected.
- Move output-plan validation before timestamp confirmation so an invalid resume
  fails without prompting or mutating normalized input.
- Add six focused tests for valid resume and every rejection branch.
- Document the append-only resume contract and example command.

### Why stable timestamps are mandatory

Auto-filled timestamps use the current time. Generating them again on a resumed
run would make old messages appear new and duplicate the import. Requiring
source timestamps makes the existing per-session cursor idempotent without a
new checkpoint schema or input-side index.

### Out of scope

- No checkpoint format or storage migration.
- No arbitrary mid-file rewrite detection.
- No change to L1/L2/L3 scheduling.
- No automatic resume; users must opt in with `--resume`.

## Verification

- `npm test -- src/cli/commands/seed.test.ts` — 6 passed
- `npm test` — 73 passed
- `npm run build` — passed
- `npm pack --dry-run` — passed, package size 716.4 kB
- `git diff --check` — clean

## Impact

- **Breaking change:** No.
- **Default behavior:** Unchanged.
- **Persistence:** Reuses existing checkpoint fields; no schema changes.
- **Safety:** Resume is rejected unless the input can be replayed
  deterministically.

## Checklist

- [x] Resume requires explicit user intent.
- [x] Missing/unstable inputs fail before pipeline startup.
- [x] Existing overwrite protections remain intact.
- [x] Focused and full test suites pass.
- [x] Full build and package dry-run pass.
- [x] Documentation includes constraints and example usage.

### DIFF
--- a/src/cli/README.md
+++ b/src/cli/README.md
@@ -26,6 +26,7 @@ openclaw memory-tdai seed --input <file> [options]
 | `--config <file>` | — | 配置覆盖文件（JSON，与 openclaw.json 插件配置深度合并） |
 | `--strict-round-role` | — | 严格校验每轮对话必须包含 user 和 assistant 消息 |
 | `--yes` | — | 跳过交互确认（如时间戳自动填充确认） |
+| `--resume` | — | 从指定输出目录的 checkpoint 恢复；必须同时提供 `--output-dir`，且输入消息必须有稳定时间戳 |
 
 ### 示例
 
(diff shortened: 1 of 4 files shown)


Maxwell-Code07:

PR #336, opened 2026-07-01
fix(l1): filter trivial extraction inputs

RerankerGuo:
## Description | 描述
Restore a conservative L1 quality gate for trivial extraction inputs.

This addresses the Problem 1 path in #82 without taking on the broader pre-extraction, retry, or post-validation scope:
- keep L0 capture permissive so raw conversation archival still records short user turns
- filter trivial L1 extraction inputs such as `OK`, `hi`, and `好的`
- reject very short non-CJK/CJK fragments before they spend LLM extraction tokens
- reject oversized L1 extraction inputs above 5000 characters
- preserve short but meaningful identity/preference statements such as `I am Bob`, `我叫小王`, and `记住我喜欢茶`

## Related Issue | 关联 Issue
Related to #82

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [ ] Documentation update | 文档更新
- [x] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

## Additional Notes | 其他说明
Verified with:
- `npx vitest run src/utils/sanitize.test.ts`
- `git diff --check`
- `npm run build`

This PR intentionally keeps the scope small so it can be reviewed independently from larger L1 pipeline redesigns.

PR #649, opened 2026-07-31
feat(l1): add high-confidence rule pre-extraction (#82)

RerankerGuo:
## Background

Issue #82 proposes four independent L1 extraction improvements. The maintainer triage explicitly asked that the remaining work be split into narrow follow-ups after the canonical fixes for Problem 1 (#336), JSON retry (#190), and robust parsing (#181).

This PR implements **Problem 2 only**: conservative, HIGH-confidence rule-based extraction for explicit persona and long-term instruction statements.

Refs #82

## What changed

- Added `preExtractHighConfidence()` with anchored full-message rules for explicit persona/identity/preferences and long-term instructions in Chinese and English.
- Runs the pre-extractor **after** the background/new-message split and only on new user messages.
- Never pre-extracts assistant or background context.
- Removes deterministic matches from the LLM prompt; when every new user message is handled, the LLM call is skipped entirely.
- Keeps unmatched/ambiguous messages in their original order for the normal LLM path.
- Places deterministic results before LLM results so `maxMemoriesPerSession` does not discard the higher-confidence memory first.

### Why this scope

The earlier broad implementation review identified a semantics bug where background messages could be stored as new memories. This version makes that boundary explicit and directly tests it.

It also uses anchored, full-message patterns rather than substring matching. A sentence such as “I think this plan might work, but compare both databases” remains in the LLM path; only statements such as “I am a backend engineer” or “Please always use concise status updates” bypass it.

### Explicitly out of scope

- MEDIUM-confidence hints
- episodic/date rules
- malformed JSON retry (#190 already covers it)
- L1 quality gates (#336 already covers them)
- post-LLM confidence validation (separate follow-up)
- schema or persistence changes

## Verification

- `COREPACK_ENABLE_AUTO_PIN=0 npm test -- --run src/core/record/pre-extractor.test.ts` — 5/5 passed
- `COREPACK_ENABLE_AUTO_PIN=0 npm test` — 73/73 passed
- `COREPACK_ENABLE_AUTO_PIN=0 npm run build` — passed
- `git diff --check` — passed

Focused coverage proves:

1. explicit user persona/instruction statements are extracted deterministically;
2. assistant messages are never pre-extracted;
3. background messages remain context-only;
4. all-deterministic user input skips the LLM even when an assistant acknowledgement is present;
5. mixed input removes matched text from the prompt and merges deterministic + LLM results;
6. ambiguous messages remain on the LLM path.

## Impact

- Breaking change: No
- Storage/schema change: No
- Performance: Saves one LLM call when all new user inputs are deterministic; otherwise reduces prompt size for matched statements.
- Affected modules: L1 extraction only

## Checklist

- [x] Narrow scope: one #82 subproblem
- [x] Focused regression tests added
- [x] Full test suite passes
- [x] Full build passes
- [x] No background/assistant extraction
- [x] No overlap with canonical #336/#190/#181

### DIFF
--- a/src/core/record/l1-extractor.ts
+++ b/src/core/record/l1-extractor.ts
@@ -23,6 +23,7 @@ import type { IMemoryStore } from "../store/types.js";
 import type { EmbeddingService } from "../store/embedding.js";
 import { report } from "../report/reporter.js";
 import type { LLMRunner, Logger } from "../types.js";
+import { preExtractHighConfidence } from "./pre-extractor.js";
 
 const TAG = "[memory-tdai][l1-extractor]";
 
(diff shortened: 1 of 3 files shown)


RerankerGuo:
Closing to avoid duplicate review work. Issue #82 already has canonical implementations for this exact subproblem: #269 covers high-confidence direct pre-extraction, while #83 covers the broader four-part proposal and #336 covers the trivial-input gate. This parallel PR would overlap those changes, so it should not remain in the maintainer queue.

PR #650, opened 2026-07-31
feat(l1): add auditable replay for historical L0 rows (#541)

### LINKED ISSUE #541: [Feature] Add supported L1 replay for L0 skipped after extraction failures
## Problem

L0 records that were skipped after an L1 extraction failure have no supported recovery path.

On versions where an LLM exception is converted into an unsuccessful L1 result but the pipeline still marks the batch complete, `last_l1_cursor` advances past the failed L0 rows. After the LLM is restored, normal pipeline runs query only rows with `recorded_at > last_l1_cursor`, so the historical rows are never attempted again.

PR #430 prevents this for future failures by treating unsuccessful extraction as a pipeline failure and preserving the cursor. However, it does not recover L0 data whose cursor already advanced before the fix.

## Reproduction

This can be reproduced without any private conversation data:

1. Run the standalone gateway with L1 extraction enabled and a threshold of one conversation.
2. Configure a test LLM endpoint to fail once (for example, return HTTP 500).
3. Add a generic conversation such as "Project Atlas has a deployment freeze until Friday" through `POST /v2/conversation/add`.
4. Wait for or trigger the L1 pipeline.
5. Observe that the L0 row exists, no corresponding L1 record is stored, and `last_l1_cursor` has advanced beyond the row.
6. Restore the LLM endpoint and let the pipeline run again.
7. Observe that new post-cursor L0 rows can be processed, but the original failed row is not retried.

The important invariant is not that every L0 row must produce an L1 memory; valid extraction may produce zero memories. The invariant is that every L0 row must receive at least one successful extraction attempt, or remain replayable until it does.

## Current alternatives and their limitations

- `POST /session/end` only flushes sessions with pending `conversation_count`; it does not replay rows behind an already-advanced cursor.
- The V2 API has no replay or force-extraction endpoint.
- Re-seeding through `seed-v2` inserts duplicate L0 rows with new IDs and write timestamps, so it does not preserve original lineage.
- Manually editing checkpoint state is an unsupported operation and may reprocess already-successful rows without an auditable receipt.

## Requested recovery path

Please add a supported, idempotent L1 repair/replay command or admin API that:

- Reads existing L0 rows directly, without inserting duplicate L0 data.
- Can be scoped by instance/session and optionally a recorded-at range.
- Supports `--dry-run` and reports the rows/batches that would be attempted.
- Preserves original L0 message IDs and lineage.
- Avoids duplicate L1 records, or clearly reports dedup/update/skip decisions.
- Advances or repairs the checkpoint only after successful extraction.
- Produces an auditable replay receipt so operators can verify at-least-once processing.

For example:

```text
memory-tencentdb replay-l1 --session-key session-retry --from <timestamp> --to <timestamp> --dry-run
```

## Related work

- #104 reports cursor loss when extraction is blocked by content filtering.
- #115 reports silent seed failures and missing retries.
- #430 fixes cursor preservation for future unsuccessful L1 extractions.

I also validated #430 in an isolated worktree by applying it to `feat/server` at `2d1fa5a`: a synthetic transient-failure regression test confirmed that the first failed run leaves the cursor at zero and the second run processes the same L0 row. A separate regression confirmed that an L0 row at or before an already-advanced cursor is not sent to the LLM. The focused test suite passed (4 tests), and the plugin build completed successfully. This validation confirms the prevention fix, but historical replay still needs a supported operator path.

RerankerGuo:
## Summary

Closes #541.

Adds a supported Gateway admin API for replaying historical L0 rows that were skipped after an earlier L1 extraction failure and are now behind `last_l1_cursor`.

## What changed

- Add `POST /admin/replay-l1` with required `session_key` and optional inclusive `from`, `to`, `limit`, and `dry_run` controls.
- Query existing L0 rows directly from SQLite or Tencent Cloud VectorDB; no duplicate L0 seed/import writes are created.
- Preserve original L0 message IDs in the extraction prompt and resulting `source_message_ids` lineage.
- Reuse the existing L1 extraction and dedup pipeline.
- Persist audit receipts to `.metadata/l1-replay-receipts.jsonl` for completed, failed, and idempotently skipped replay attempts.
- Fingerprint exact completed batches so repeating the same replay does not invoke the LLM or write L1 again.
- Keep the live `last_l1_cursor` unchanged. Historical replay uses an independent receipt ledger rather than moving the online scheduler cursor backward and accidentally reprocessing unrelated rows.

## API example

```bash
curl -X POST http://127.0.0.1:8420/admin/replay-l1 \
  -H "Authorization: Bearer $TDAI_GATEWAY_API_KEY" \
  -H 'Content-Type: application/json' \
  -d '{
    "session_key": "session-retry",
    "from": "2026-07-01T00:00:00Z",
    "to": "2026-07-02T00:00:00Z",
    "limit": 100,
    "dry_run": true
  }'
```

The admin route is unavailable when `TDAI_GATEWAY_API_KEY` is not configured; this prevents an unauthenticated caller from triggering LLM calls and persistent memory writes.

## Design decisions

- **No checkpoint rewind:** replaying rows behind the cursor must not reset the live scheduler cursor. The response and receipt include `checkpoint_cursor_before` and `checkpoint_cursor_after` so operators can verify this invariant.
- **Bounded execution:** `limit` defaults to 100 and is capped at 500.
- **Range correctness:** both storage backends apply session and recorded-at bounds before the limit.
- **Fail visibly:** unsuccessful L1 extraction records a failed receipt and returns an error instead of being treated as a successful zero-memory replay.

## Verification

- [x] `npm test` — 70/70 tests passed
- [x] `npm run build:plugin`
- [x] `git diff --check`

Focused replay tests cover:
- dry-run range planning with no LLM call or receipt write;
- successful extraction with original L0 IDs and unchanged checkpoint;
- exact replay idempotency;
- failed extraction receipt with unchanged checkpoint.

## Impact

- Breaking change: no
- Existing capture, extraction, seed, and scheduler behavior: unchanged
- New persistent data: append-only replay receipt JSONL under `.metadata/`

### DIFF
--- a/src/core/record/l1-replay.test.ts
+++ b/src/core/record/l1-replay.test.ts
@@ -1,0 +1,197 @@
+import fs from "node:fs/promises";
+import path from "node:path";
+import { describe, expect, it, vi } from "vitest";
+import { parseConfig } from "../../config.js";
+import { CheckpointManager } from "../../utils/checkpoint.js";
+import type { MemoryRecord } from "./l1-writer.js";
+import type { IMemoryStore, L0QueryRow, L0ReplayQuery } from "../store/types.js";
+import type { LLMRunner, Logger } from "../types.js";
+import {
+  L1_REPLAY_RECEIPT_PATH,
+  L1ReplayExecutionError,
+  replayL1,
+} from "./l1-replay.js";
+
+const logger: Logger = {
+  debug: () => {},
+  info: () => {},
+  warn: () => {},
+  error: () => {},
+};
+
+const rows: L0QueryRow[] = [
+  {
+    record_id: "m1",
+    session_key: "session-retry",
+    session_id: "conversation-a",
+    role: "user",
+    message_text: "Project Atlas has a deployment freeze until Friday.",
+    recorded_at: new Date(1_000).toISOString(),
+    timestamp: 1_000,
+  },
+  {
+    record_id: "m2",
+    session_key: "session-retry",
+    session_id: "conversation-a",
+    role: "assistant",
+    message_text: "I will remember the deployment freeze and its deadline.",
+    recorded_at: new Date(2_000).toISOString(),
+    timestamp: 2_000,
+  },
+];
+
+function makeStore(stored: MemoryRecord[]): IMemoryStore {
+  return {
+    isDegraded: () => false,
+    queryL0ForReplay: vi.fn(async (query: L0ReplayQuery) =>
+      rows.filter((row) => {
+        const recordedAt = Date.parse(row.recorded_at);
+        return row.session_key === query.sessionKey &&
+          (query.fromRecordedAtMs == null || recordedAt >= query.fromRecordedAtMs) &&
+          (query.toRecordedAtMs == null || recordedAt <= query.toRecordedAtMs);
+      }).slice(0, query.limit)),
+    countL1: () => 0,
+    isFtsAvailable: () => false,
+    upsertL1: (record: MemoryRecord) => {
+      stored.push(record);
+      return true;
+    },
+  } as unknown as IMemoryStore;
+}
+
+function makeRunner(run: LLMRunner["run"]): LLMRunner {
+  return { run };
+}
+
+function makeDeps(baseDir: string, store: IMemoryStore, llmRunner: LLMRunner) {
+  return {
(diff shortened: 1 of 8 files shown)


Maxwell-Code07:

PR #661, opened 2026-07-31
perf(hooks): stabilize recall context to restore prefix cache hits

TPJHtpjh:
## Description | 描述

OpenAI-compatible providers (DeepSeek, MiMo, Qwen, GLM) cache by automatic prefix matching — any byte change in the system prompt or an earlier message invalidates everything after it. Two memory-injection behaviours were destroying the prefix every turn:

1. **System-prompt churn.** `appendSystemContext` (persona + scene navigation + tool guide) is concatenated to the system-prompt tail. Background L2/L3 pipelines regenerate persona and scene index on their own schedule, and the scene index carries a monotonically increasing `heat` counter + `updated` timestamp — so every scene hit or persona refresh rewrites the tail and busts the entire conversation cache.

2. **Per-turn re-injection.** `prependContext` (recalled L1 memories) is injected on every turn. A memory already in the model's context window is re-injected anyway, adding a gratuitous prefix break with no new information.

This PR introduces `RecallCachePolicy` (`src/core/hooks/recall-cache-policy.ts`), a per-session in-memory policy that:

- **Freezes `appendSystemContext`** for the session. A refresh is only allowed once both a minimum turn count (default 20) and a minimum wall-clock interval (default 30 min) have elapsed. A transient read failure (persona.md mid-write) replays the frozen revision instead of dropping it.
- **De-duplicates recalled L1 memories** at the session level. A memory already injected is skipped; only memories that survive the character budget are committed, so a budget-dropped memory remains eligible for a later turn.
- **Stabilizes scene navigation rendering** — coarse heat tiers + deterministic filename sort instead of raw counters and mtime, so an intra-tier heat bump produces byte-identical output.

All three are gated behind `recall.cacheStability` (default: enabled), with per-knob overrides in `openclaw.plugin.json`. Disabling the policy restores the previous behaviour exactly.

## Related Issue | 关联 Issue

Closes #120

## Change Type | 修改类型

## Self-test Checklist | 自测清单

**Test results:** 100 unit tests pass (`npm test`), including 20 new tests for `RecallCachePolicy` and 13 simulation tests modelling prefix reuse and `showInjected` bloat.

**Real A/B measurement (DeepSeek `deepseek-chat`, 20 turns × 3 repeats, cache isolation per arm):**

| Arm | Steady-state hit rate |
|---|---|
| Churning (v0.3.6 behaviour) | 30.8% |
| Stabilized (this PR) | 94.6% |

The churning arm decays monotonically (62% → 25%) while the stabilized arm holds at ~97–98%, confirming the mechanism: the churn point truncates the cacheable prefix so the whole transcript can never cache.

A granularity probe confirms DeepSeek uses token-level longest-common-prefix matching with 64-token blocks, validating the model's assumptions. The probe and A/B scripts are included in `__tests__/cache/` (outside the `files` whitelist, not shipped in the npm package).

## Additional Notes | 其他说明

- `RecallCachePolicy` state is in-memory and per-process; losing it degrades to the previous behaviour (re-inject + refresh), never to incorrect output.
- Session state is garbage-collected after a configurable TTL (default 12 h) and explicitly reset on `handleSessionEnd`.
- All config knobs are clamped to safe minimums (`clampMin`) to prevent a negative value from disabling a gate.
- Analysis tooling (simulator, bloat model, live scripts) lives in `__tests__/cache/` — outside `src/`, so it never enters the published package.


Maxwell-Code07:

PR #666, opened 2026-07-31
fix(recall): add bounded prompt-cache memory epochs

HelicasECoode42:
## Summary

- place the stable persona/scene snapshot in `prependSystemContext`, before OpenClaw's cache boundary
- keep per-turn L1 recall fresh while persisting only append-only Registry/Focus changes
- advance an active session's Stable Cache Epoch after explicit L2/L3 publication, causing one intentional cache miss before the new snapshot stabilizes
- correlate `before_prompt_build` retries by `runId` and drain pending writes FIFO per session, preventing concurrent turns from overwriting each other
- reconstruct memory epochs after restart, rotate them after host compaction, and cap persistent growth by token budget
- require OpenClaw `2026.5.28+`; the official package contract was inspected instead of silently falling back on unknown older hosts

## Why

Issue #120 is caused by both prompt placement and history growth: stable memory was composed after the provider cache boundary, while `showInjected` could persist changing recall text every turn. A stable snapshot alone fixes local rebuilding but does not control the dynamic tail.

This change models the dynamic tail as a bounded Memory Epoch. Memory content is registered once, subsequent turns update only focused IDs, and provider-visible event bytes are persisted exactly as sent. The epoch is sealed before unbounded growth; after sealing, current recall remains available ephemerally until host compaction.

The host adapter now treats Stable Cache Epoch separately from Memory Epoch. Repeated calls within one cache epoch keep identical system bytes; an explicit Core epoch change updates the active session on its next turn. Because OpenClaw's `before_message_write` hook has no `runId`, prepared turns are made retry-idempotent and queued in transcript-write order.

## Host contract

The published `openclaw@2026.5.28` package was inspected directly:

- `prependSystemContext` is composed before the base system prompt and its `CACHE_BOUNDARY`
- `before_prompt_build` exposes `runId`/`sessionId`
- `before_message_write` exposes `sessionKey` but no `runId`
- `after_compaction` is public

The plugin metadata, peer dependency, and README now declare `2026.5.28` as the minimum supported OpenClaw version.

## Validation

- `npm test`: 87 passed
- `npm run build:plugin`
- state-machine coverage includes retries, concurrent pending turns, restart, cache-epoch advance, sealing, high-cardinality growth, and compaction rotation
- 100-turn high-cardinality stress with a 512-token epoch budget: sealed at turn 5; turns 6-100 did not grow the transcript
- DeepSeek A/B weighted cache hit rate: 99.5% in epoch mode vs 63.9% with append + clean history
- MiMo 2.5 Pro A/B weighted cache hit rate: 84.43% in epoch mode vs 64.53% with append + clean history
- low-budget MiMo smoke retained answer correctness after sealing while keeping the transcript bounded

Provider results are cache-read ratios, not promises of equivalent TTFT or end-to-end latency improvement; those measurements remain noisy.

## Scope

The adapter's restart, sealing, and `after_compaction` transitions are covered by state-machine tests. Provider A/B validates prompt-cache behavior. The plugin reacts to OpenClaw's public `after_compaction` lifecycle; host-controlled automatic compaction scheduling itself remains outside plugin control.

Fixes #120


Maxwell-Code07:

Maxwell-Code07:

PR #671, opened 2026-08-01
feat(memory-core): L0→L3 replay/rerun CLI tool

### LINKED ISSUE #670: feat: L0→L3 replay/rerun CLI tool for memory extraction debugging
## Motivation

When a memory is extracted (or mis-extracted), there is no way to reproduce
**why** the pipeline produced that result. The extraction runs in the
background worker, and the prompt / model output are not persisted.

A replay / rerun tool addresses the reproducible half of that problem:
feed an existing data dir through the real L1→L2→L3 runners and record every
LLM call (systemPrompt / prompt / response / duration) into a JSON report.

## What this change adds

New standalone CLI under `MemoryCore/scripts/replay-pipeline/`:

- Re-runs the **production** `createL1Runner` / `createL2Runner` /
  `createL3Runner` (same prompt + extraction logic, not a parallel impl).
- Wraps the LLM runner with a recording decorator to capture each call's
  prompt / response / duration.
- Safe by default: copies the data dir to a temp workdir; **only the temp
  dir is auto-deleted**. User-specified `--work-dir` is never deleted.
- `--clean` (default) wipes derived data (L1 records / scene_blocks /
  persona / checkpoint) for an isolated L0→L3 rebuild; `--keep-state`
  continues incrementally.
- Drains all L1 batches (loops on `hasMore` / `hasFullBacklog`), not just
  the first batch per session.
- Per-stage `status` (ok/failed/skipped); detects LLM failures that runners
  swallow internally; exits 1 on any stage failure.
- `--compare` diffs two report files.
- `--redact` masks emails/phones/keys; reports written with `0600`.

## Scope / limitations

This is a **rerun** tool, not a historical replay. It shows what re-running
the pipeline on current data produces. It cannot reconstruct why a *past*
run produced its result (historical inputs / model version / retrieval
candidates were not persisted).

## Tests

- `recording-runner.test.ts`: 3 unit tests (record success / failure / factory).
- Verified e2e: workdir safety (user dir retained), batch drain (25 L0 → 3
  LLM rounds), failure → exit 1, redaction, report mode 0600.

luojiyin1987:
## Summary

Adds `MemoryCore/scripts/replay-pipeline/`, a standalone CLI that re-runs the
production L1→L2→L3 extraction pipeline against an existing data dir while
recording every LLM call into a JSON report. Goal: make extraction
reproducible for debugging ("why did this memory come out like this?").

Closes #670.

## Changes

- `scripts/replay-pipeline/replay-pipeline.ts` — CLI (list-sessions /
  replay L1/L2/L3 / compare / redact).
- `scripts/replay-pipeline/recording-runner.ts` — `RecordingLLMRunner`
  + `RecordingLLMRunnerFactory`, transparent prompt/response capture.
- `scripts/replay-pipeline/recording-runner.test.ts` — 3 unit tests.
- `scripts/replay-pipeline/README.md`, `tsconfig.json`.
- `MemoryCore/package.json` — `replay-pipeline` script; files entry.
- `MemoryCore/vitest.config.ts` — include the script's tests.
- `.gitignore` — exclude local `.codegraph/` index and `replay-report*.json`.

## Safety design

- Copies data dir to a temp workdir by default. **Only the mkdtemp-created
  temp dir is auto-deleted**; a user-specified `--work-dir` is never deleted.
- Rejects `--work-dir` equal to / nested in `--data-dir`.
- `--clean` (default) wipes derived state for an isolated L0→L3 rebuild;
  `--keep-state` for incremental continuation.
- `--redact` masks emails/phones/keys; report written with mode `0600`;
  `replay-report*.json` gitignored.

## Design decisions

- Reuses the production `createL1Runner`/`createL2Runner`/`createL3Runner`
  factories, so the replay runs the same prompts and logic as live mode.
- Loops L1 on `hasMore`/`hasFullBacklog` so history larger than one batch
  is fully drained.
- Per-stage `status` + detection of LLM failures that the runners swallow
  internally; non-zero exit on any stage failure.

## Scope / known limitations

This is a **rerun** tool, not a historical replay. It cannot reconstruct why
a past run produced a result (historical inputs / model version / retrieval
candidates were not persisted). Documented in README.

## Verification

- `npx vitest run scripts/replay-pipeline/` — 3 passed.
- `npx tsc --project scripts/replay-pipeline/tsconfig.json` — clean for the
  tool files (pre-existing repo-wide `src/` errors are unrelated).
- E2E with a mock OpenAI endpoint:
  - user-specified workdir retained after run;
  - `--work-dir == --data-dir` rejected;
  - 25 L0 messages → 3 L1 LLM rounds (batch drain);
  - unreachable LLM → `failed: true`, exit code 1;
  - `--redact` + report mode `0600`.


luojiyin1987:
Addressed all 6 review findings in `aa15652`.

## Fixes

**P1 profile-scoped L2/L3** — L2/L3 write to `profiles/<encoded scope>/`, but clean/report only touched the root. Now:
- enumerate all scopes (`profiles/` dirs + legacy root),
- clean per scope,
- report gains a per-scope `profiles` section with scenes/persona before/after.

**P1 `--clean` vs `--stages`/`--session-key`** — clean is now dependency-aware:
- `L1`: clears L1 records + checkpoint + downstream L2/L3.
- `L2` only: keeps L1, clears scenes/persona per scope.
- `L3` only: keeps L1/L2, clears persona only.
Dangerous combos are guarded: `--no-copy --clean` requires `--dangerous`;
`--no-copy --session-key --clean` is refused outright.

**P1 `--keep-state` L2 incremental** — now reads the L2 cursor
(`pipeline_states[*].l2_last_extraction_time`) before the run and writes
back the returned `latestCursor`, so L2 processes only data after the
cursor instead of re-reading all history.

**P1 exception-path temp-dir cleanup** — workdir copy, store init, report
write all moved inside a `try/finally`; no `process.exit()` after workdir
creation. A temp dir created by this tool is always removed even when the
run throws.

**P1 report 0600 on overwrite** — atomic write: random 0600 tmp file →
fsync → rename → `chmod 0600`. Overwriting an existing 0644 file now
tightens to 0600 (covered by a test).

**P2 WAL snapshot** — `vectors.db` is copied via SQLite `VACUUM INTO` for a
consistent single-file snapshot; falls back to plain cp with a warning if
unavailable. README documents the residual time-skew between L0 JSONL and
SQLite on a live directory.

## Tests added

`replay-pipeline.test.ts` (5 CLI integration tests):
1. `--no-copy --clean` without `--dangerous` → rejected.
2. `--no-copy --session-key --clean` → rejected.
3. `--list-sessions` lists keys from SQLite.
4. overwriting an existing 0644 report yields 0600.
5. exception path (store init failure) leaves no temp dir behind.

Total: 8 tests pass (`recording-runner` 3 + `replay-pipeline` 5).

## Not addressed

- Historical replay remains out of scope: the tool proves what re-running
  on current data produces, not what a past run produced (historical
  inputs/model versions/retrieval candidates were never persisted). Kept
  the tool named "replay/rerun" and documented the limitation in README.

luojiyin1987:
Addressed all 6 third-round findings in `cd643a7`.

## 1. L2 cursor / state key / write-back

Verified against production `PipelineManager.runL2` (pipeline-manager.ts:929):
- reads `last_extraction_updated_time` as the data cursor (was `l2_last_extraction_time`),
- writes back `latestCursor` to `last_extraction_updated_time` and sets
  `l2_last_extraction_time` to wall-clock now,
- uses read-modify-write of the checkpoint so **all other
  PipelineSessionState fields are preserved** (previously reset to 0/""),
- collects `profileScopes` returned by L1 and drives L2 with those
  `profile:team:...|session:...` keys (production-compatible); falls back
  to checkpoint-derived keys, then bare session keys when L1 did not run.

## 2. before/after snapshots

- `snapshotProfiles(workDir)` runs **before** the stage, then
  `snapshotProfilesAfter(workDir, before)` **after**, and unions the scope
  sets so a profile dir created during this run still appears in the report.
- L2 and L3 both use this pattern; `scenesBefore`/`scenesAfter`/
  `personaBeforeChars`/`personaAfterChars` now reflect real before/after
  state instead of two reads with no intervening work.

## 3. dependency-aware clean

- `cleanScenes = hasL1 || hasL2`; `cleanPersona = cleanScenes || hasL3`.
  `--stages L1` now clears downstream scenes/persona (they are rebuilt by
  L1), matching the documented semantics.
- `cleanScopeL2L3` also removes the scoped `.metadata/checkpoint.json`, so
  `PersonaTrigger` state (`scenes_processed` / `last_persona_at` /
  `request_persona_update`) is not carried into a clean replay.

## 4. copy-failure temp-dir leak

- `prepareWorkDir` now removes the temp dir it created inside its own
  `try/catch` when `copyDataDir` throws, instead of relying on an outer
  `finally` that only starts after `prepareWorkDir` returns.
- The existing exception-path test now asserts the temp dir actually
  disappears (previously it only checked a non-zero exit code).

## 5. SQLite snapshot sidecars

- `copyDataDir` excludes `vectors.db*` (wal/shm/journal) during the copy and
  deletes any sidecar afterwards, so the target is a clean single-file
  snapshot after `VACUUM INTO`.
- `snapshotSqliteDb` now loads `sqlite-vec` when present before `VACUUM INTO`,
  so vec0 virtual tables are handled by the registered module rather than
  silently falling back. A test asserts no `-wal`/`-shm`/`-journal` remain
  after copy.

## 6. symlink workdir bypass

- `prepareWorkDir` compares `realpath()` of data-dir and work-dir (resolving
  the deepest existing ancestor for non-existent paths), so a symlink that
  aliases work-dir to data-dir is rejected instead of bypassing the
  `--no-copy --dangerous` guard. Covered by a test.

## Tests

Total 11 pass (3 unit + 8 CLI integration). Added: symlink rejection,
sidecar cleanup, exception-path temp dir actually removed, L1-only clean
clears downstream scenes/persona.

Remaining known limitation (unchanged, documented): this is a rerun tool —
it cannot reconstruct what a past production run produced because that
input/model/retrieval state was never persisted.

PR #683, opened 2026-08-02
fix: make local embedding provider (provider=local) reachable from config

### LINKED ISSUE #678: Local embedding provider (provider=local) is unreachable from config and startWarmup() is never called
### Summary

### Impact

### Suggested fix

Schrotti77:
## Summary

Fixes #678 — `provider=local` (fully-offline embedding via node-llama-cpp + embeddinggemma-300m, 768-dim) was unreachable from config:

1. **config.ts** rewrote `provider=local` to `provider=none` with `embeddingEnabled=false` ("Local embedding provider is not available in user config").
2. **core/store/factory.ts** (sqlite branch) only created an embedding service when `provider !== "local" && apiKey` was set.
3. After creation, **nothing called `startWarmup()`** — the GGUF model was never downloaded, so `embed()` threw `EmbeddingNotReadyError`.

## Changes

- **config.ts**: `provider=local` now passes through with `embeddingEnabled=true` (no API key required).
- **core/store/factory.ts**: creates the embedding service for `provider=local` without requiring `apiKey`, and calls `startWarmup()` once at startup (safe no-op for remote providers).
- **Regression tests**: `src/core/store/__tests__/fix-678-local-embedding.test.ts` — 4 tests covering config parsing, factory creation, warmup availability, and the remote-without-apiKey guard.

## Test Plan

- `npx tsc --noEmit` — 0 errors
- `npx vitest run src/core/store/__tests__/fix-678-local-embedding.test.ts` — 4/4 passed

Verified on main @ 0.3.6 with Standalone Gateway (src/gateway/server.ts), as reported in the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@
 
 ## [Unreleased]
 
+### 🐛 修复
+
+- **本地嵌入提供方 (`provider=local`) 现在可从配置启用** ([#678](https://github.com/TencentCloud/TencentDB-Agent-Memory/issues/678))：`config.ts` 之前将 `provider=local` 改写为 `none`，导致完全离线的嵌入（node-llama-cpp + embeddinggemma-300m, 768-dim）不可用。
+  - `config.ts`：`provider=local` 现在直接通过，`embeddingEnabled=true`（无需 API key）。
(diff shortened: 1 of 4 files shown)


Maxwell-Code07:
Thanks for your contribution! We will schedule this PR for internal review and get back to you later.

PR #687, opened 2026-08-02
fix: bound Gateway stdout/stderr log growth with in-place rotation

### LINKED ISSUE #583: [Bug] Hermes Gateway stdout/stderr logs grow without runtime bounds

Schrotti77:
Fixes #583 — GatewaySupervisor opened `gateway.stdout.log` / `gateway.stderr.log` in append mode with no runtime size limit. In one observed incident a single stdout log grew to **63 GiB / 420M lines in ~4 days** (triggered by the #549 flush recursion).

## Changes (`hermes-plugin/memory/memory_tencentdb/supervisor.py`)
- **`LOG_MAX_BYTES`** (64 MiB per stream): hard runtime cap independent of application correctness
- **`_rotate_logs_if_needed()`**: when a log exceeds the cap, preserves a bounded diagnostic tail as `<path>.1`, then truncates the active file **in place** — keeps the child process's inherited descriptor valid (renaming alone would keep appending to the old inode)
- **`ensure_running()`**: periodic size check with a 300s monotonic guard (cheap `getsize` + threshold when the Gateway is already healthy)

## Test Plan
- Rotation unit-verification: active file truncated to 0 bytes, `.1` tail preserved with expected byte count — PASS
- `python3 -m py_compile` — OK

Closes #583

### DIFF
--- a/hermes-plugin/memory/memory_tencentdb/supervisor.py
+++ b/hermes-plugin/memory/memory_tencentdb/supervisor.py
@@ -32,6 +32,14 @@ HEALTH_CHECK_RETRIES = 3     # retries for is_running check
 # Log file rotation parameters
 LOG_TAIL_BYTES_ON_CRASH = 2048  # bytes of stderr log to surface on startup crash
 
+# Fix #583: hard runtime cap on active Gateway stdout/stderr logs.
+# When a log exceeds this, a bounded diagnostic tail is preserved as `.1`
+# and the active file is truncated in place (keeps the child's inherited
+# descriptor valid — renaming alone would keep appending to the old inode).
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:
Thanks for your contribution! We will schedule this PR for internal review and get back to you later.

PR #692, opened 2026-08-03
fix(embedding): unlock local provider and trigger startWarmup (#678)

### LINKED ISSUE #678: Local embedding provider (provider=local) is unreachable from config and startWarmup() is never called

RerankerGuo:
## 背景 (Background)

Closes #678

`createEmbeddingService()` supports `LocalEmbeddingService` (node-llama-cpp), but two layers blocked it:
1. `config.ts` rewrites `provider=local` to `provider=none`, disabling local embedding.
2. `factory.ts` only creates an embedding service when `config.embedding.provider !== "local"`.

Additionally, after a local service was created, `startWarmup()` was never called, so the GGUF model was never downloaded and `embed()` would fail with `EmbeddingNotReadyError`.

## 改动内容 (What & Why)

1. **`src/config.ts`**:
   - Changed logic for `provider="local"`: Now sets `embeddingEnabled = true` and `embeddingProvider = "local"` instead of disabling it.

2. **`src/core/store/factory.ts`**:
   - Modified condition in `createSqliteBackend`: Now creates the embedding service for `provider="local"` even without an API key.
   - After creating the service, triggers `embeddingService.startWarmup()` as a fire-and-forget promise for local providers, ensuring the GGUF model is downloaded and ready.

3. **`src/core/store/factory.test.ts`**:
   - Added vitest cases to verify that `startWarmup()` is called for `provider="local"` and is NOT called for remote providers.

## 验证方式 (Verification)

- [x] Added unit tests in `src/core/store/factory.test.ts` covering both local and remote provider behaviors.
- [x] Verified logic manually via code review.

## 影响范围 (Impact)

- Users who want fully-offline embedding (no external API key) can now enable it via config.
- The model will be loaded automatically on startup.
- No breaking changes for remote embedding providers.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -389,13 +389,10 @@ export function parseConfig(raw: Record<string, unknown> | undefined): MemoryTda
     embeddingProvider = "none";
     embeddingEnabled = false;
   } else if (embeddingProviderRaw === "local") {
-    // Local embedding is not exposed to users; treat as disabled at entry level.
-    // Internal LocalEmbeddingService code is preserved but not reachable from config.
-    embeddingProvider = "none";
-    embeddingEnabled = false;
-    embeddingConfigError =
-      "Local embedding provider is not available in user config. " +
-      "Please configure a remote embedding provider (e.g. openai, deepseek). Embedding has been disabled.";
+    // Local embedding provider (node-llama-cpp): enabled, no API key required
+    // startWarmup() will be called during core/gateway initialization
+    embeddingProvider = "local";
+    embeddingEnabled = true;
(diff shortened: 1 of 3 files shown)


RerankerGuo:
Closing as a duplicate of the earlier #683. Both PRs make provider=local reachable in config, create the local embedding service without an API key, call startWarmup(), and add focused regression coverage. #683 predates this branch and also includes the related CHANGELOG entry, so keeping one canonical implementation avoids duplicate maintainer review. #682 remains complementary for the tool-handler warmup timing path.

Maxwell-Code07:

PR #695, opened 2026-08-03
fix(knowledge): release idle per-asset build queues

RerankerGuo:
## Description | 描述

`MemoryKnowledge` keeps one `SerialQueue` in `BuildQueue.queues` for every
wiki or code-graph asset key ever observed. Completed and failed jobs leave
their empty queue in the map permanently, so a long-lived service processing
new asset IDs grows this registry without a bound.

This PR releases a per-asset queue after its jobs settle and the queue reaches
a true idle state.

- Cleanup is scheduled for both fulfilled and rejected jobs.
- A queue-identity check prevents an old cleanup from deleting a replacement
  queue for the same key.
- A `WeakSet` coalesces cleanup scheduling to one idle waiter per queue
  instance, even when many same-key jobs are pending.
- Same-key jobs remain serialized; no public API or job error semantics change.

Regression tests cover 100 distinct asset keys, a rejected build job, and
same-key reuse while a second job is still running.

## Related Issue | 关联 Issue

Related to #694. That PR recovers a `SerialQueue` after synchronous task
throws; this PR addresses the independent retention of already-idle queues by
the owning `BuildQueue`.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证

- `vitest run src/store/build-queue.test.ts` - 3 tests passed
- `vitest run` - 3 tests passed
- strict changed-file TypeScript check - passed
- `tsdown` - 11 output files built
- `git diff --check` - passed

## Additional Notes | 其他说明

- No API, configuration, storage schema, or dependency changes.
- Validation reused the existing workspace dependencies without modifying any
  package manifest or lockfile.

### DIFF
--- a/MemoryKnowledge/src/store/build-queue.test.ts
+++ b/MemoryKnowledge/src/store/build-queue.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, it, vi } from "vitest";
+
+import { BuildQueue } from "./build-queue.js";
+
+function activeQueueCount(queue: BuildQueue): number {
+  return (
+    queue as unknown as {
+      queues: Map<string, unknown>;
+    }
+  ).queues.size;
+}
+
+async function expectQueuesReleased(queue: BuildQueue): Promise<void> {
+  await vi.waitFor(() => {
+    expect(activeQueueCount(queue)).toBe(0);
+  });
+}
+
+describe("BuildQueue", () => {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #698, opened 2026-08-03
fix(config): prevent stale VDB cache resurrection

RerankerGuo:
## Description | 描述

`InstanceConfigProvider.evictVdb()` and `clear()` removed cached VDB configs but did not invalidate requests already awaiting the remote config source. When an older request completed, `fetchAndStoreVdb()` inserted the explicitly evicted instance back into the cache. Its unconditional `finally` cleanup could also delete a newer in-flight request registered for the same instance.

This PR binds a small invalidation token to each in-flight fetch:

- eviction and full cache clearing invalidate and detach the current request;
- stale requests still resolve for their original callers, but cannot repopulate the cache;
- completion only removes its own in-flight entry, so a newer request remains deduplicated;
- no per-instance generation registry is retained after requests settle.

Regression tests cover eviction during a fetch, a fresh fetch racing the stale request, and full-cache clearing.

## Related Issue | 关联 Issue

No existing issue. Discovered by the default-branch concurrency audit after all open Issue candidates were deduplicated.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证

- `vitest run` - 1 file, 3 tests passed
- `git diff --check` - passed
- remote author identity checked after PR creation

## Additional Notes | 其他说明

- No API, configuration schema, storage schema, or dependency changes.
- Eviction does not cancel the source request; it only prevents stale cache resurrection.
- Full changed-file and PR-body deduplication found no active implementation for `instance-config-provider.ts`.

### DIFF
--- a/MemoryCore/src/core/instance-config-provider.test.ts
+++ b/MemoryCore/src/core/instance-config-provider.test.ts
@@ -1,0 +1,92 @@
+import { describe, expect, it, vi } from "vitest";
+
+import type { IConfigSource } from "./abstractions/index.js";
+import {
+  InstanceConfigProvider,
+  type VdbConfig,
+} from "./instance-config-provider.js";
+
+const vdbConfig: VdbConfig = {
+  url: "https://vdb.example.test",
+  user: "test-user",
+  apiKey: "test-key",
+  database: "test-db",
+};
+
+const logger = {
+  debug: vi.fn(),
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #699, opened 2026-08-03
fix(metadata): dedupe concurrent store initialization

RerankerGuo:
## Description | 描述

Concurrent cold calls to `MetadataStorePool.getStore(instanceId)` all observed an empty cache because the store was cached only after asynchronous initialization. Each request therefore created and initialized a separate SQLite/MongoDB store; the last result replaced the cache entry while the other connections were leaked to callers.

This PR adds per-instance in-flight initialization deduplication:

- concurrent cold requests share one store creation Promise and receive the same store;
- failed initialization removes only its own pending entry and remains retryable;
- `purgeInstance()` waits for an already-started initialization before closing and dropping the database;
- `closeAll()` waits for pending stores before sweeping the cache, preventing post-shutdown connection resurrection.

A SQLite regression test proves that three concurrent cold calls now return the exact same store object.

## Related Issue | 关联 Issue

No existing issue. Discovered by the default-branch metadata lifecycle audit after all open Issue candidates were deduplicated.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证

- `vitest run` - 1 file, 1 test passed
- `git diff --check` - passed
- remote author identity checked after PR creation

## Additional Notes | 其他说明

- No public API, metadata schema, configuration, or dependency changes.
- Deduplication is scoped per instance; different instance databases still initialize concurrently.
- Full changed-file and PR-body deduplication found no active implementation for `metadata/store/factory.ts`.

### DIFF
--- a/MemoryCore/src/metadata/store/factory.test.ts
+++ b/MemoryCore/src/metadata/store/factory.test.ts
@@ -1,0 +1,35 @@
+import { mkdtemp, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+
+import { MetadataStorePool } from "./factory.js";
+
+const cleanup: Array<() => Promise<void>> = [];
+
+afterEach(async () => {
+  await Promise.allSettled(cleanup.splice(0).map((fn) => fn()));
+});
+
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #700, opened 2026-08-03
test(panel): lock KnowledgeTaskRegistry TTL and lifecycle

RerankerGuo:
## Description | 描述

Add first-class vitest coverage for `MemoryPanel/src/panel/state/knowledge-task-registry.ts`, the in-memory registry that Panel uses to hand off owner user keys between the code-graph/wiki create path and the S2S ready callback.

The module documents a TTL sweep contract but has no test file today, so future refactors could silently break the callback flow.

## Related Issue | 关联 Issue

None; this is `MemoryPanel` module-audit test coverage found during a default-branch review.

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [x] Code optimization | Test coverage

## Self-test Checklist | 自测清单

## What & Why

- **New test suite `MemoryPanel/tests/state/knowledge-task-registry.test.ts`**:
  - `record`/`peek` round-trip preserves multiple task types.
  - Same-key re-record replaces the previous owner tuple.
  - `take` returns and removes the task; second `take` yields `undefined`.
  - Explicit `sweep(now)` respects the configured TTL.
  - `record` implicitly sweeps stale entries when a new one is added.
- Uses `vi.useFakeTimers` around every scenario so `record()`/`peek()`'s
  implicit `sweep(Date.now())` calls observe deterministic ages regardless of
  wall-clock time. Without this, tests would occasionally see `undefined`
  because the real clock is far larger than the fake `created_at` timestamps.

### Boundaries

- Does not change any source file — pure additive test coverage.
- Does not add any new runtime dependency; only uses existing vitest.

## Verification | 验证

- `npx vitest run tests/state/knowledge-task-registry.test.ts` — 5/5 passed.
- `npm test` — full suite, 5/5 passed (no other tests exist yet in this scope).
- `npm run typecheck` — clean.
- `git diff --check origin/feat/server_team...HEAD` — clean.

### DIFF
--- a/MemoryPanel/tests/state/knowledge-task-registry.test.ts
+++ b/MemoryPanel/tests/state/knowledge-task-registry.test.ts
@@ -1,0 +1,120 @@
+import { describe, expect, it, vi } from 'vitest';
+
+import { KnowledgeTaskRegistry } from '../../src/panel/state/knowledge-task-registry.js';
+import type { KnowledgeTask } from '../../src/panel/state/knowledge-task-registry.js';
+
+const BASE_TIME = 1_700_000_000_000; // 2023-11-14 in ms epoch
+
+function task(overrides: Partial<KnowledgeTask> = {}): KnowledgeTask {
+  return {
+    knowledge_id: 'wiki-1',
+    type: 'wiki',
+    team_id: 'team-1',
+    owner_user_id: 'user-1',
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:

PR #701, opened 2026-08-03
test(panel): lock chat-memory governance rules

RerankerGuo:
## Description | 描述

Add vitest coverage for `MemoryPanel/src/panel/domain/chat-memory-governance.ts`, the chat-memory ownership / borrow-cap module that both frontend and backend rely on.

The module documents a strict validation contract (borrow cap, same-team peers, no self-borrow, no duplicates) and a `metadata_json` namespace round-trip, but has no test file today. Test coverage is critical because the shape is being migrated from `metadata_json` into a dedicated schema field later, so any silent regression would corrupt every agent's `chat_memory` relation.

## Related Issue | 关联 Issue

None; this is `MemoryPanel` module-audit test coverage found during a default-branch review.

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [x] Code optimization | Test coverage

## Self-test Checklist | 自测清单

## What & Why

- **New test suite `MemoryPanel/tests/domain/chat-memory-governance.test.ts`**:
  - `validateImportedAgents` accepts up to `MAX_IMPORTED_AGENTS` distinct
    same-team peers.
  - Rejects arrays exceeding the borrow limit with the documented reason.
  - Rejects non-array inputs, empty-string ids, self-borrow, cross-team ids,
    and duplicates — each with the exact reason string exposed to the UI.
  - `readChatMemoryRel` returns `DEFAULT_CHAT_MEMORY_REL` when metadata is
    missing or unparseable.
  - `writeChatMemoryRel` preserves foreign `metadata_json` keys and
    round-trips the `chat_memory` namespace.
  - Malformed inputs to `writeChatMemoryRel` are normalized: booleans default
    to shared, ids are de-duplicated, non-string entries are dropped, and the
    borrow cap is enforced on write.

### Boundaries

- Does not change any source file — pure additive test coverage.
- Does not add any new runtime dependency; only uses existing vitest.

## Verification | 验证

- `npx vitest run tests/domain/chat-memory-governance.test.ts` — 7/7 passed.
- `npm test` — full suite, 7/7 passed.
- `npm run typecheck` — clean.
- `git diff --check origin/feat/server_team...HEAD` — clean.

### DIFF
--- a/MemoryPanel/tests/domain/chat-memory-governance.test.ts
+++ b/MemoryPanel/tests/domain/chat-memory-governance.test.ts
@@ -1,0 +1,98 @@
+import { describe, expect, it } from 'vitest';
+
+import {
+  DEFAULT_CHAT_MEMORY_REL,
+  MAX_IMPORTED_AGENTS,
+  readChatMemoryRel,
+  validateImportedAgents,
+  writeChatMemoryRel,
+} from '../../src/panel/domain/chat-memory-governance.js';
+
+const teamAgents = [
+  { agent_id: 'a1' },
+  { agent_id: 'a2' },
+  { agent_id: 'a3' },
+];
+
+describe('validateImportedAgents', () => {
+  it('accepts up to MAX_IMPORTED_AGENTS distinct same-team peers', () => {
+    const first = MAX_IMPORTED_AGENTS >= 1 ? ['a2'] : [];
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:

PR #708, opened 2026-08-03
fix(proxy): clear failed connectivity probe timers

RerankerGuo:
## Description | 描述

Fix a startup resource leak in the MemoryProxy HTTP connectivity probe.

Each HTTP probe creates a five-second abort deadline. The success path cleared
that timer, but an immediate DNS or connection failure returned from `catch`
without cleanup. A startup check with several quickly failing dependencies
therefore retained live timers and `AbortController` closures until every
deadline expired, which can delay short-lived process exit and waste resources.

### Changes

- Create the probe deadline before entering the request `try` block.
- Clear it in `finally`, covering success, HTTP/body errors, and immediate
  transport failures.
- Add a fake-timer regression test proving that a rejected `fetch()` leaves
  zero pending timers after `checkConnectivity()` returns.

### Boundaries

- No timeout duration or connectivity result semantics change.
- Redis probing and dependency configuration are unchanged.
- This PR only fixes HTTP probe deadline ownership.

## Related Issue | 关联 Issue

Discovered during the required default-branch resource-lifecycle audit.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证

- `npx vitest run src/__tests__/connectivity.test.ts`
  - 1/1 passed.
- `npm test`
  - 1/1 passed (the default branch currently contains this one collected
    MemoryProxy test).
- `git diff --check`
  - passed.
- `npx tsc --noEmit`
  - reaches the existing default-branch blocker:
    `Cannot find module '@context-proxy/cost-guard'`; the optional private
    package is not present in this checkout and this PR does not touch it.

## Additional Notes | 其他说明

Semantic deduplication was run before implementation and immediately before
PR creation. No open PR references this timer leak or changes
`MemoryProxy/src/connectivity.ts`.

### DIFF
--- a/MemoryProxy/src/__tests__/connectivity.test.ts
+++ b/MemoryProxy/src/__tests__/connectivity.test.ts
@@ -1,0 +1,39 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("../report/log.js", () => ({
+  log: {
+    info: vi.fn(),
+    warn: vi.fn(),
+  },
+}));
+
+import { DEFAULT_CONFIG } from "../config.js";
+import { checkConnectivity } from "../connectivity.js";
+
+describe("connectivity HTTP probe cleanup", () => {
+  afterEach(() => {
+    vi.unstubAllGlobals();
+    vi.useRealTimers();
+  });
+
+  it("clears the abort deadline when fetch fails before the timeout", async () => {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #710, opened 2026-08-03
fix(deploy): brace-wrap variables followed by full-width CJK characters

sK3n0b1:
### Description
Under a non-UTF-8 locale (bash launched without `LC_ALL`/`LANG`, common in CI shells and automation on macOS), bash includes the bytes of a full-width CJK character that immediately follows a bare `$VAR` into the variable name. With `set -u` this aborts the script with `unbound variable`.

Reproduced with `start-memory-core.sh:174` — `$ADMIN_KEY_FILE）` fails the whole first boot (admin key never created, hub/proxy never start).

The repo already uses the `${VAR}` style before CJK punctuation in 22 places; this PR fixes the four remaining bare references (1 in `start-memory-core.sh`, 3 in `deploy/panel-knowledge-combined/build.sh`).

### Change Type
- [x] Bug fix

### Self-test Checklist
- [x] `bash -n` clean on both files
- [x] Full `./start-all.sh` boot verified on macOS under C locale

### Additional Notes
If you'd like a regression gate, a tiny shell-lint check over `deploy/**/*.sh` could be added to CI — happy to follow up; kept out of this diff to stay minimal.

### DIFF
--- a/deploy/global-images/start-memory-core.sh
+++ b/deploy/global-images/start-memory-core.sh
@@ -171,7 +171,7 @@ verify_user_key() {
   [[ "$code" == "200" ]]
 }
 
-info "初始化 admin user（username=${MEMORY_CORE_ADMIN_USERNAME}, key 持久化 → $ADMIN_KEY_FILE）..."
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

Issue #709, opened 2026-08-03
MEMORY_LLM_PROTOCOL=anthropic passes verify.sh but is silently ignored by memory-core (L1 extraction fails with 404 "Not Found")


sK3n0b1:
### Summary
Setting `MEMORY_LLM_PROTOCOL=anthropic` in `deploy/global-images/.env` produces a green `verify.sh` run, but memory-core's background pipeline then fails every L1 extraction with `LLM extraction failed: Not Found`. The variable is honored by verify.sh and memory-hub, but never reaches the memory-core runtime.

### Reproduction
1. `.env`: `MEMORY_LLM_BASE_URL=https://api.anthropic.com`, a valid key, `MEMORY_LLM_MODEL=claude-haiku-4-5-20251001`, `MEMORY_LLM_PROTOCOL=anthropic`
2. `./verify.sh` → `[ok] ... Anthropic 协议通路 OK（模型 ... 已应答）`
3. `./start-all.sh`, then add a conversation via `POST /v2/conversation/add`
4. memory-core log: `ERROR ... [l1-extractor] LLM extraction failed: Not Found`

### Root cause chain
| Component | Behavior |
|---|---|
| `verify.sh:86-129` | probes `POST {base}/v1/messages` with `x-api-key` + `anthropic-version` → 200, green |
| `start-memory-core.sh:66-71` | generates the gateway yaml **without any protocol field**; `MEMORY_LLM_PROTOCOL` is not referenced anywhere in the script |
| `MemoryCore/src/config.ts:202-228` | `StandaloneLLMOverrideConfig` has no protocol field; `provider` means `"openai" \| "proxy"` (access mode, not wire protocol) |
| `MemoryCore/src/config.ts:618-637` | unknown provider values silently coerce to `"openai"` |
| `MemoryCore/src/adapters/standalone/llm-runner.ts:283-289` | always `createOpenAI({compatibility:"compatible"})` → `POST {base}/chat/completions` with `Authorization: Bearer` |

So the runtime calls `https://api.anthropic.com/chat/completions`, which does not exist → 404. (The terse "Not Found" log line is a separate problem — a PR follows for that too.)

### Workaround
Use Anthropic's OpenAI-compatible endpoint:
`MEMORY_LLM_BASE_URL=https://api.anthropic.com/v1` + `MEMORY_LLM_PROTOCOL=openai`. Works end-to-end (verified: extraction produces atoms).

### Proposal
[...]

PR #711, opened 2026-08-03
fix(memory-core): include status, url and response body in LLM failure logs

sK3n0b1:
### Description
When an LLM call fails, `llm-runner.ts` logs only `err.message`. For AI SDK's `APICallError` that can be as terse as `Not Found` — the failing URL, HTTP status and response body exist on the error object but are dropped, making misconfiguration (e.g. a wrong base URL) undebuggable from logs. The same bare message goes into `report("llm_call")` telemetry, so the context is lost there too.

This PR enriches the single top-level catch in `run()` (through which all standalone LLM calls pass): when `APICallError.isInstance(err)`, the log line and telemetry now carry `status=`, `url=` and a 2048-char response body preview with a remainder counter — the same preview convention already used in `l1-extractor.ts` (NO_JSON diagnostics).

Before: `run() failed after 546ms: Not Found`
After: `run() failed after 546ms: Not Found (status=404, url=https://api.anthropic.com/chat/completions, body={...})`

### Related Issue
#709

### Change Type
- [x] Bug fix

### Self-test Checklist
- [x] `npm run build` — build:plugin green (note: build:scripts fails on a pre-existing missing `scripts/seed-v2/`, reported separately)
- [x] `scripts/ci/check-skill-queue-isolation.sh` PASS

### DIFF
--- a/MemoryCore/src/adapters/standalone/llm-runner.ts
+++ b/MemoryCore/src/adapters/standalone/llm-runner.ts
@@ -18,7 +18,7 @@
 
 import fsPromises from "node:fs/promises";
 import path from "node:path";
-import { generateText, tool, stepCountIs, jsonSchema } from "ai";
+import { APICallError, generateText, tool, stepCountIs, jsonSchema } from "ai";
 import { createOpenAI } from "@ai-sdk/openai";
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:

PR #712, opened 2026-08-03
fix(deploy): warn when MEMORY_LLM_PROTOCOL=anthropic reaches memory-core; document the limitation

sK3n0b1:
### Description
`MEMORY_LLM_PROTOCOL` is honored by verify.sh and memory-hub, but memory-core's standalone runner always speaks OpenAI-compatible `/chat/completions` (see #709 for the full chain). With `anthropic` configured, verify.sh is green while every L1 extraction fails with 404 — a silent contradiction.

This PR makes the failure loud and documented, without breaking hub-only setups where the value is legitimate:
- `start-memory-core.sh`: prints a prominent `warn` when `MEMORY_LLM_PROTOCOL=anthropic`, pointing to the working configuration (`MEMORY_LLM_BASE_URL=https://api.anthropic.com/v1` + `MEMORY_LLM_PROTOCOL=openai`);
- `.env.example` and `deploy/global-images/README.md`: annotate that `anthropic` applies to hub/knowledge only; memory-core requires an OpenAI-compatible endpoint.

Real Anthropic support is proposed separately in #709.

### Related Issue
#709

### Change Type
- [x] Bug fix

### Self-test Checklist
- [x] `bash -n` clean
- [x] Boot with `protocol=anthropic` shows the warning and continues; boot with `openai` unchanged

### DIFF
--- a/deploy/global-images/.env.example
+++ b/deploy/global-images/.env.example
@@ -34,7 +34,7 @@ PROXY_IMAGE=agentmemory/memory-proxy:latest
 MEMORY_LLM_BASE_URL=REPLACE_ME              # 例：https://api.deepseek.com/v1
 MEMORY_LLM_API_KEY=REPLACE_ME               # 例：sk-xxxxxxxx
 MEMORY_LLM_MODEL=REPLACE_ME                 # 例：deepseek-chat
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

PR #713, opened 2026-08-03
fix(deploy): restrict generated gateway config to 0600 (contains the LLM API key)

sK3n0b1:
### Description
`start-memory-core.sh` writes `MEMORY_LLM_API_KEY` in plaintext into `.memory-core-config/tdai-gateway.yaml` with default 0644 permissions — readable by any local process. This PR tightens the generated file to owner-only right after generation.

### Change Type
- [x] Bug fix (security hygiene)

### Self-test Checklist
- [x] Boot unchanged; generated file mode is 0600


Maxwell-Code07:

PR #717, opened 2026-08-03
fix(proxy): clear pending-write deadline timer

RerankerGuo:
## Description | 描述

`flushPendingWrites()` is used by the MemoryProxy SIGTERM handler to wait for
fire-and-forget L0 writes. It races the writes against a deadline timeout, but
the timeout remains scheduled when the writes settle first.

Because the timer is referenced, a fully drained shutdown can still keep the
Node process alive until the complete deadline (10 seconds in the production
caller). This change clears that timer as soon as the race finishes.

## Related Issue | 关联 Issue

No linked issue. Found during the default-branch MemoryProxy shutdown and
resource-lifecycle audit.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## What & Why

- Keep the deadline handle and clear it from a `finally` block after
  `Promise.race()` settles.
- Preserve the timeout behavior when writes do not drain before the deadline.
- Add fake-timer tests for both early drain and deadline expiry.

The change does not alter write tracking, retry behavior, the default deadline,
or the reported `{ drained, remaining }` contract.

## Verification | 验证

- `pnpm vitest run src/tdai/__tests__/pending-writes.test.ts`
  - 1 file, 2 tests passed
- `pnpm test`
  - 1 file, 2 tests passed
- Strict changed-file TypeScript check
- `git diff --check`

The early-drain regression verifies that the active timer count returns to
zero immediately. The deadline case verifies the existing non-drained result
and cleanup after timeout.

## Additional Notes | 其他说明

- Scope is limited to one shutdown helper and its focused tests.
- The target branch's existing `pnpm-lock.yaml` disagrees with
  `package.json` on `cos-nodejs-sdk-v5`; validation used the current manifest
  with lockfile writes disabled, and this PR contains no dependency changes.
- AI assistance: TRAE was used to audit the lifecycle, reproduce the lingering
  timer, implement the cleanup, and run validation. I reviewed and take
  responsibility for the final change.

### DIFF
--- a/MemoryProxy/src/tdai/__tests__/pending-writes.test.ts
+++ b/MemoryProxy/src/tdai/__tests__/pending-writes.test.ts
@@ -1,0 +1,51 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import {
+  flushPendingWrites,
+  pendingWriteCount,
+  trackWrite,
+} from "../pending-writes.js";
+
+afterEach(() => {
+  vi.useRealTimers();
+});
+
+describe("flushPendingWrites", () => {
+  it("clears the deadline timer when writes drain early", async () => {
+    vi.useFakeTimers();
+    let resolveWrite!: () => void;
+    const write = new Promise<void>((resolve) => {
+      resolveWrite = resolve;
+    });
+    trackWrite(write);
+
+    const flushing = flushPendingWrites(10_000);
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #718, opened 2026-08-03
fix(knowledge): dedupe code-graph instance loads

RerankerGuo:
## Description | 描述

Code-graph query routes lazily reopen a persisted index through
`instancePool.loadIfMissing()`. The current pool only checks the completed
instance map, so concurrent cache misses each call `openIndex(dir)`.

That creates multiple handles for the same index: callers can receive different
instances, the last map write wins, and the other handle is never closed. A
delete during a lazy load can also let the stale completion repopulate the
pool after cleanup.

This change gives the pool explicit in-flight and invalidation lifecycle
management.

## Related Issue | 关联 Issue

No linked issue. Found during the default-branch MemoryKnowledge concurrency
and resource-lifecycle audit.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## What & Why

- Extract instance-pool lifecycle logic into a small injected helper.
- Share one in-flight load for concurrent callers of the same code-graph ID.
- Invalidate pending loads when an instance is deleted or explicitly replaced.
- Close a handle that completes after invalidation and prevent it from
  repopulating the pool.
- Remove failed loads from the in-flight map so a later query can retry.
- Preserve the original optional `loadIfMissing` interface contract for route
  mocks and external callers.

The change does not alter query endpoints, index format, startup restoration,
or the existing synchronous close behavior for completed instances.

## Verification | 验证

- `pnpm vitest run src/code-graph-instance-pool.test.ts`
  - 1 file, 4 tests passed
- `pnpm test`
  - 1 file, 4 tests passed
- Changed-file strict TypeScript check
- `pnpm build`
  - 11 output files built
- `git diff --check`

The tests deterministically cover concurrent deduplication, delete-during-load
invalidation/close, and retry after rejected or synchronously thrown loads.

## Additional Notes | 其他说明

- Full-package `pnpm typecheck` reaches an existing default-branch error at
  `src/middleware/response-envelope.ts:47` (`Promise<string>` assigned to
  `string`). The changed files pass an isolated strict TypeScript check and
  the production build succeeds.
- Node 26 cannot compile the target branch's `better-sqlite3@11.10.0`; tooling
  installation used `--ignore-scripts`. These tests are pure TypeScript and do
  not load SQLite.
- AI assistance: TRAE was used to audit the concurrency path, design the
  invalidation ownership, implement tests, and run validation. I reviewed and
  take responsibility for the final change.

### DIFF
--- a/MemoryKnowledge/src/code-graph-instance-pool.test.ts
+++ b/MemoryKnowledge/src/code-graph-instance-pool.test.ts
@@ -1,0 +1,113 @@
+import { describe, expect, it, vi } from "vitest";
+
+import {
+  createCodeGraphInstancePool,
+  type CodeGraphInstancePoolOptions,
+} from "./code-graph-instance-pool.js";
+import type { CodeGraphInstance } from "./engines/code/index.js";
+
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  let reject!: (reason?: unknown) => void;
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+  return { promise, resolve, reject };
+}
+
+function instance(name: string): CodeGraphInstance {
+  return { name } as unknown as CodeGraphInstance;
+}
+
+function createPool(
+  loadIndex: CodeGraphInstancePoolOptions["loadIndex"],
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

PR #721, opened 2026-08-03
fix(panel): reject failed knowledge binding responses

RerankerGuo:
## Description | 描述

The Panel startup binding client previously accepted any parseable JSON response unless it contained an explicit non-zero envelope code. An HTTP 500 response such as `{}` was therefore treated as success, and startup logged that the Knowledge Service binding had been refreshed even though the request failed.

This change also requires a successful HTTP status before accepting the response. Legacy 2xx responses without an envelope code remain supported. Focused tests cover both paths.

## Related Issue | 关联 Issue

Maintenance fix found during default-branch startup transport audit; no linked issue.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Verification | 验证方式

- `npx vitest run tests/startup/ensure-knowledge-llm-binding.test.ts` (2/2)
- `npm test` (2/2)
- `npm run typecheck`
- `npm run build`
- `git diff --check`

## Impact | 影响范围

Only Panel startup calls to the Knowledge Service LLM-binding endpoints change. HTTP failures now enter the existing best-effort warning/recovery path. Successful legacy responses remain compatible. No API or configuration changes.

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

The startup flow remains best-effort and does not block Panel startup.

### DIFF
--- a/MemoryPanel/src/panel/startup/ensure-knowledge-llm-binding.ts
+++ b/MemoryPanel/src/panel/startup/ensure-knowledge-llm-binding.ts
@@ -74,7 +74,7 @@ async function ksPost<T>(
       signal: ctrl.signal,
     });
     const json = (await resp.json().catch(() => null)) as KsEnvelope<T> | null;
-    if (!json || (json.code !== undefined && json.code !== 0)) {
+    if (!resp.ok || !json || (json.code !== undefined && json.code !== 0)) {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #723, opened 2026-08-03
fix(offload): clear client request deadlines

RerankerGuo:
## Description | 描述

`OffloadApiClient` created an abort deadline for each health, ingest, L1.5, and compaction request, but cleared it only after `fetch()` resolved. An immediate connection or transport rejection returned through `catch` while leaving the timer active for another 5-30 seconds.

This PR:
- centralizes the repeated timeout setup in a private request helper;
- clears every deadline in `finally`, covering both resolve and reject paths;
- adds fake-timer regression coverage for all four public request paths.

Timeout values, response handling, logging, and Offload Server API contracts are unchanged.

## Related Issue | 关联 Issue

L3 resource-lifecycle audit finding; no existing issue.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证方式

- `npm test` (4/4)
- `npm run build:plugin`
- `git diff --check`

## Additional Notes | 其他说明

The test rejects `fetch()` immediately and verifies that no pending timer remains after each API method returns.

### DIFF
--- a/MemoryCore/src/offload-client/offload-api-client.test.ts
+++ b/MemoryCore/src/offload-client/offload-api-client.test.ts
@@ -1,0 +1,50 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { OffloadApiClient } from "./offload-api-client.js";
+import type { Logger, OffloadClientConfig } from "./types.js";
+
+const config: OffloadClientConfig = {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #724, opened 2026-08-03
fix(telemetry): drain ClickHouse inserts on shutdown

RerankerGuo:
## Description | 描述

`ClickHouseDirectExporter` starts inserts in the background when a buffer reaches `batchSize`. `shutdown()` flushed only the rows still in memory and then closed the shared client immediately, without waiting for those threshold-triggered inserts. A delayed request could therefore finish after client close, and a failed request could requeue rows after the final flush.

This PR:
- tracks every in-flight insert promise;
- waits for existing inserts during shutdown;
- performs a second flush for rows requeued by failed inserts before closing;
- makes shutdown idempotent because the OTel span and log adapters share one exporter;
- adds deterministic tests for delayed and failed in-flight inserts.

Batch sizing, buffer limits, ClickHouse schema, and normal runtime export behavior are unchanged.

## Related Issue | 关联 Issue

L3 telemetry lifecycle audit finding; no existing issue.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证方式

- `npm test` (2/2)
- `npm run build:plugin`
- `git diff --check`

## Additional Notes | 其他说明

The tests hold an insert promise pending and verify that client close cannot run early. They also reject an in-flight insert and verify its requeued row is flushed before the single client close.

### DIFF
--- a/MemoryCore/src/core/report/clickhouse-exporter.test.ts
+++ b/MemoryCore/src/core/report/clickhouse-exporter.test.ts
@@ -1,0 +1,83 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const mocks = vi.hoisted(() => ({
+  close: vi.fn(),
+  createClient: vi.fn(),
+  insert: vi.fn(),
+}));
+
+vi.mock("@clickhouse/client", () => ({
+  createClient: mocks.createClient,
+}));
+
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #726, opened 2026-08-03
fix(core): retry failed store initialization

RerankerGuo:
## Description | 描述

`initStores()` caches the first initialization Promise for each data directory. When a transient backend failure is converted into a degraded result (`vectorStore: undefined`), that resolved Promise currently remains cached indefinitely. Every later caller receives the same unusable result until shutdown explicitly calls `resetStores()`.

This PR keeps successful initialization cached, but evicts failed or degraded results after they settle so a later request can retry. Cache deletion is guarded by Promise identity, preventing an older completion from deleting a newer entry after an explicit reset. Concurrent callers still share one in-flight initialization.

Regression tests cover both lifecycle guarantees:

- concurrent callers share one failed initialization, followed by a successful retry;
- concurrent callers continue to share one successful initialization.

## Related Issue | 关联 Issue

No existing issue. Discovered during a default-branch MemoryCore lifecycle audit.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证

- `pnpm vitest run src/utils/pipeline-factory-init.test.ts` - 1 file, 2 tests passed
- `pnpm test` - 1 file, 2 tests passed
- `npm run build:migrate-sqlite-to-vdb` - passed
- `npm run build:export-tencent-vdb` - passed
- `npm run build:read-local-memory` - passed
- `git diff --check` - passed

The aggregate `pnpm build` reaches the existing `build:seed-v2` step and then fails because `scripts/seed-v2/tsconfig.json` is absent on the base branch. All preceding production and script build steps pass.

## Additional Notes | 其他说明

- No public API, configuration, storage schema, or dependency changes.
- Successful store initialization retains the existing once-per-data-directory behavior.
- Full Open PR title/body and changed-file checks found no active implementation of failed `initStores()` retry behavior on `feat/server_team`.

### DIFF
--- a/MemoryCore/src/utils/pipeline-factory-init.test.ts
+++ b/MemoryCore/src/utils/pipeline-factory-init.test.ts
@@ -1,0 +1,121 @@
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import type { MemoryTdaiConfig } from "../config.js";
+import type { IMemoryStore } from "../core/store/types.js";
+
+const mocks = vi.hoisted(() => ({
+  createStoreBundle: vi.fn(),
+}));
+
+vi.mock("../core/store/factory.js", () => ({
+  createStoreBundle: mocks.createStoreBundle,
+}));
+
+import {
+  initStores,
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #727, opened 2026-08-03
fix(knowledge): terminate after telemetry flush

RerankerGuo:
## Description | 描述

When Langfuse telemetry is enabled, Knowledge Service installs a `SIGTERM` listener to flush pending spans. Installing that listener suppresses Node's default SIGTERM termination, but the existing handler only started `sdk.shutdown()` and never exited or restored the signal behavior. The HTTP service could therefore remain alive after telemetry flush.

This change installs a one-shot handler that flushes telemetry once, logs shutdown failures, and then re-sends SIGTERM after the listener has been removed so Node resumes its default termination behavior. Focused tests verify ordering, duplicate-signal suppression, and failure handling.

## Related Issue | 关联 Issue

Maintenance fix found during default-branch telemetry lifecycle audit; no linked issue.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Verification | 验证方式

- `npx vitest run src/telemetry.test.ts` (2/2)
- `npm test` (2/2)
- `npm run build` (11 files)
- `git diff --check`

## Impact | 影响范围

Only SIGTERM handling when Knowledge Langfuse telemetry is enabled changes. Telemetry still receives a graceful flush opportunity, and the process now terminates afterward even if flush fails. No API or configuration changes.

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

The repository-wide Knowledge typecheck still fails on the unchanged default-branch error in `src/middleware/response-envelope.ts:47`; the changed files pass strict TypeScript checking independently.

### DIFF
--- a/MemoryKnowledge/src/telemetry.test.ts
+++ b/MemoryKnowledge/src/telemetry.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it, vi } from "vitest";
+
+import { createTelemetrySigtermHandler } from "./telemetry.js";
+
+function deferred(): {
+  promise: Promise<void>;
+  resolve: () => void;
+  reject: (error: Error) => void;
+} {
+  let resolve!: () => void;
+  let reject!: (error: Error) => void;
+  const promise = new Promise<void>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #735, opened 2026-08-03
fix(proxy): make logger initialization idempotent

RerankerGuo:
## Description | 描述

`initLogger()` documents that initialization is idempotent, but repeated calls
currently replace the active backend and FileLogger references. Any previous
backend resources, file stream, and flush timer then become unreachable and
are omitted from the later shutdown.

This change:

- ignores initialization attempts after the first successful lifecycle entry;
- preserves the original log level, backend, and FileLogger ownership;
- adds a deterministic lifecycle test proving that logs and shutdown remain
  attached to the original backend.

Log formatting, filtering, backend behavior, and normal single-call startup
are unchanged.

## Related Issue | 关联 Issue

L3 MemoryProxy logging lifecycle audit finding; no existing issue.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证方式

- `npm test -- --run src/__tests__/logger-lifecycle.test.ts` (1/1)
- `npm test` (1/1 collected Proxy test)
- `git diff --check`

## Additional Notes | 其他说明

The repository-wide Proxy typecheck retains unchanged default-branch errors in
`RawYamlConfig.memCommand`, two session `isDefault` fields, and the absent
optional `@context-proxy/cost-guard` package. This PR does not change those
modules, dependencies, or lockfiles.

### DIFF
--- a/MemoryProxy/src/__tests__/logger-lifecycle.test.ts
+++ b/MemoryProxy/src/__tests__/logger-lifecycle.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, it, vi } from "vitest";
+import type { ILogBackend, LogConfig } from "../report/types.js";
+
+function createBackend(type: string): ILogBackend {
+  return {
+    type,
+    debug: vi.fn(),
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #685, opened 2026-08-02
security: fix auth bypass and SSRF in MemoryProxy & MemoryKnowledge

Schrotti77:
## Summary

Fixes #672 — all 4 vulnerabilities from the security advisory:

### CRITICAL: MemoryProxy Auth Bypass
1. **rate-limits.ts** — `GET/PUT/DELETE /v3/admin/rate-limits` had no auth check; unauthenticated callers could query/modify/delete tenant rate limits (DoS via qpm=1). Now wrapped with `checkAdminAuth`.
2. **admin-auth.ts** — `if (!expected) return "ok"` meant an unconfigured `admin.apiKey` granted full admin access (including `POST /v3/instance/proxy-destroy` = remote data purge). Now fail-closed: unconfigured key → "missing" (401).

### HIGH: MemoryKnowledge SSRF + Argument Injection
3. **SSRF (git-fetcher.ts)** — `isPrivateAddress()` only regex-matched the hostname string; DNS-rebinding / cloud-metadata hosts (e.g. resolving to 169.254.169.254) bypassed it. Now also resolves DNS via `node:dns/promises` and checks the resolved address; unresolvable hosts fail closed.
4. **Argument Injection (git-fetcher.ts)** — `sourceUrl` was passed unsanitized to `simpleGit().clone()`; git option separators (`--upload-pack=...`) or shell metacharacters could inject options. Now rejected with a clear error.

## Changes

- `MemoryProxy/src/routes/rate-limits.ts` — auth guard on get/put/delete
- `MemoryProxy/src/routes/admin-auth.ts` — fail-closed on empty apiKey
- `MemoryKnowledge/src/source-fetcher/git-fetcher.ts` — DNS-resolved SSRF check + injection-safe URL validation
- Regression tests: `security-672-auth.test.ts` (4), `security-672-ssrf.test.ts` (6)
- CHANGELOG entry (Unreleased)

## Test Plan

- `npx vitest run src/routes/__tests__/security-672-auth.test.ts` — 4/4 passed
- `npx vitest run src/source-fetcher/__tests__/security-672-ssrf.test.ts` — 6/6 passed
- `npx tsc --noEmit` — only pre-existing errors (unrelated to this PR)

Note: advisory paths referenced the old monorepo layout (`MemoryProxy/src/...`); verified and fixed against current `feat/server_team`.

PR #673, opened 2026-08-01
fix(security): enforce authentication on rate limits, deny unconfigured admin access, and sanitize git fetcher args

tarun1790:
## Summary of Changes

This Pull Request resolves critical authentication bypasses and SSRF argument injection vulnerabilities across `MemoryProxy` and `MemoryKnowledge`:

1. **Enforce Authentication on Admin Rate Limit Endpoints (`MemoryProxy/src/routes/rate-limits.ts`)**:
   - Wrapped `get`, `put`, and `delete` handlers in `createRateLimitHandlers` with `checkAdminAuth()` validation to prevent unauthenticated query, override, or deletion of tenant rate limits.

2. **Deny Access on Unconfigured Admin Key (`MemoryProxy/src/routes/admin-auth.ts`)**:
   - Updated `checkAdminAuth()` to return `"missing"` instead of `"ok"` when `config.admin.apiKey` is empty/unconfigured, protecting operational administrative routes (such as `/v3/instance/proxy-destroy`) from unauthenticated exposure.

3. **Sanitize Input Parameters in Git Source Fetcher (`MemoryKnowledge/src/source-fetcher/git-fetcher.ts`)**:
   - Added validation against whitespace and trailing CLI option flags (`-c`, `--`) in `sourceUrl` before invoking `simpleGit().clone()` to prevent argument injection attacks.

## Verification
- Verified routes and type correctness across `MemoryProxy` and `MemoryKnowledge`.

PR #736, opened 2026-08-03
fix(proxy): authenticate session management routes

RerankerGuo:
## Description | 描述

The v2.0.0 release adds two Proxy session-management endpoints:

- `POST /v3/session/refresh-cache`
- `POST /v3/session/force-archive-skill`

The refresh handler explicitly documents that it uses the existing admin authentication contract, but neither handler currently calls `checkAdminAuth()`. As a result, configuring `admin.apiKey` protects the existing instance-destroy endpoint while unauthenticated callers can still reach session lookup, cache refresh, and force-archive logic.

This PR applies the shared constant-time Bearer-token guard at the start of both handlers, before JSON parsing or session operations. Focused tests cover missing, invalid, and valid tokens for both routes, plus the existing empty-key compatibility behavior.

## Related Issue | 关联 Issue

No dedicated issue. Discovered during the v2.0.0 default-branch Proxy route audit.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证

- `pnpm vitest run src/routes/__tests__/session-admin-auth.test.ts` - 1 file, 8 tests passed
- `pnpm test` - 1 file, 8 tests passed
- `git diff --check` - passed

The repository-wide Proxy typecheck reaches existing v2.0.0 baseline errors in the raw config model, session task-detail types, and private `@context-proxy/cost-guard` package resolution. None reference the changed files.

## Additional Notes | 其他说明

- No session payload, response envelope, archive, or refresh behavior changes after authentication succeeds.
- An empty `admin.apiKey` retains the current shared-helper behavior (authentication disabled). This PR intentionally does not change the fail-open/fail-closed policy under discussion in #673 and #685.
- #673/#685 modify the shared helper and rate-limit routes; full changed-file scans found no active PR touching either v2.0.0 session route.

### DIFF
--- a/MemoryProxy/src/routes/__tests__/session-admin-auth.test.ts
+++ b/MemoryProxy/src/routes/__tests__/session-admin-auth.test.ts
@@ -1,0 +1,97 @@
+import { Hono, type Context } from "hono";
+import { describe, expect, it } from "vitest";
+
+import type { ProxyConfig } from "../../types.js";
+import { createSessionForceArchiveHandler } from "../session-force-archive.js";
+import { createSessionRefreshHandler } from "../session-refresh.js";
+
+const routes = [
+  {
+    path: "/v3/session/refresh-cache",
+    createHandler: createSessionRefreshHandler,
+  },
+  {
+    path: "/v3/session/force-archive-skill",
+    createHandler: createSessionForceArchiveHandler,
+  },
+] as const;
+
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

PR #738, opened 2026-08-03
fix(skill): release recovered transient failure state

RerankerGuo:
## Description | 描述

`SkillConversationExtractWorker` keeps a per-task transient failure counter so
repeated network/timeout failures can be log-sampled. The counter was removed
when a later failure became permanent, but not when the task recovered
successfully or was terminally discarded as a ghost.

As a result, a long-lived worker retained one map entry for every recovered
task that had ever failed transiently. A reused task id also inherited the old
sampling count and no longer emitted the expected first-failure error.

This change:

- clears the transient streak only after success/ghost handling has removed the
  task under the existing tasks mutex;
- preserves the streak when extraction or terminal task removal fails and the
  task may still be retried;
- adds a regression test covering `transient failure -> success -> reused id
  transient failure`, proving that the final failure starts a fresh streak.

The retry classification, retry delay, permanent retry count, DLQ policy, and
queue locking behavior are unchanged.

## Related Issue | 关联 Issue

No linked issue. This was found while auditing the v2.0.0 long-lived skill
worker lifecycle on `feat/server_team`.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification

- `pnpm exec vitest run src/core/skill/conversation-add/extract-worker-transient-state.test.ts`
  - 1 test passed
- `pnpm test`
  - 1 test passed
- `pnpm run build:plugin`
  - plugin bundle completed successfully
- `pnpm run lint:skill-isolation`
  - passed
- `git diff --check`
  - passed

The focused test was also run before the implementation change and failed with
one error log instead of two, directly reproducing the stale streak.

## Additional Notes | 其他说明

- Breaking change: none.
- A strict standalone `tsc` invocation still reaches existing default-branch
  errors for the excluded observability integration, the OpenTelemetry v2
  `Resource` API, and an optional worker log field. The production plugin build
  succeeds and none of those diagnostics are introduced or modified here.
- Scope is limited to two files: the worker state cleanup and its regression
  test.

### DIFF
--- a/MemoryCore/src/core/skill/conversation-add/extract-worker-transient-state.test.ts
+++ b/MemoryCore/src/core/skill/conversation-add/extract-worker-transient-state.test.ts
@@ -1,0 +1,114 @@
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock("../../report/otel-context.js", () => ({
+  runInRootContext: <T>(fn: () => T): T => fn(),
+}));
+vi.mock("../../report/obs-logger.js", () => ({
+  obsLogger: {
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+  },
+}));
+vi.mock("../../report/trace.js", () => ({
+  trace: { report: vi.fn() },
+}));
+
+import type {
+  AgentTuple,
+  ISkillAgentTaskQueue,
+} from "./agent-task-queue.js";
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #739, opened 2026-08-03
fix(panel): preserve kernel task pagination

RerankerGuo:
## Background

`POST /task/list-with-agents` aggregates task and agent data from the metadata kernel. The route previously rebuilt `limit` and `offset` from the request, even though the kernel may normalize or clamp those values. This could make the response metadata disagree with the returned task page.

## Changes

- Return the pagination metadata supplied by `task/list`.
- Keep the existing request-derived values only as a compatibility fallback when older kernel responses omit pagination metadata.
- Add a route-level regression test covering a clamped limit and kernel-provided offset.

The change is limited to the Panel task aggregation response. It does not modify kernel pagination behavior or other task routes.

## Verification

- [x] `npm test -- --run tests/http/task-pagination.test.ts` (1 test passed)
- [x] `npm test` (1 test passed)
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `git diff --check`

## Impact

- Breaking change: No
- Affected module: `MemoryPanel`
- Performance impact: None

## Checklist

- [x] The change follows the existing route and test style.
- [x] A regression test proves the corrected behavior.

### DIFF
--- a/MemoryPanel/src/panel/http/routes/task.ts
+++ b/MemoryPanel/src/panel/http/routes/task.ts
@@ -153,8 +153,8 @@ export function registerTaskRoutes(api: Hono, deps: PanelDeps): void {
         okEnvelope(c, {
           items,
           total,
-          limit: (listPayload.limit as number) ?? 50,
-          offset: (listPayload.offset as number) ?? 0,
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #740, opened 2026-08-03
fix(sdk): validate v3 response envelopes

RerankerGuo:
## Background

The TypeScript v3 HTTP transport treated malformed successful responses inconsistently. A success envelope without a numeric `code` could surface as `TDAMError(code=200)`, while a primitive `data` value could be returned as though it were a valid SDK result. The Python v3 transport already rejects these protocol-invalid shapes.

## Changes

- Parse response JSON as `unknown` and require the top-level envelope to be an object.
- Treat a missing or invalid business code on a successful HTTP response as protocol error `-1`.
- Require successful `data` to be an object before returning it.
- Preserve the current behavior for valid object data, business errors, request IDs, and trace IDs.
- Add focused tests for invalid envelopes, missing codes, primitive data, and valid trace propagation.

The change is limited to v3 HTTP response validation. It does not alter request serialization or public SDK method signatures.

## Verification

- [x] `npm test -- --run tests/v3-http-envelope.test.ts` (4 tests passed)
- [x] `npm test` (4 tests passed)
- [x] `npm run build`
- [x] `git diff --check`

## Impact

- Breaking change: No for protocol-valid server responses
- Affected module: `sdk/memory-core/typescript`
- Performance impact: None

## Checklist

- [x] The change follows the existing transport and test style.
- [x] Regression tests cover malformed and valid envelopes.

### DIFF
--- a/sdk/memory-core/typescript/src/v3/http.ts
+++ b/sdk/memory-core/typescript/src/v3/http.ts
@@ -62,9 +62,9 @@ export class V3HttpTransport {
         response.headers.get("x-trace-id") ??
         "";
 
-      let envelope: ApiResponseEnvelope<T>;
+      let parsed: unknown;
       try {
-        envelope = JSON.parse(responseText) as ApiResponseEnvelope<T>;
+        parsed = JSON.parse(responseText) as unknown;
       } catch {
         throw new TDAMError(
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #741, opened 2026-08-03
fix(proxy): restore v2 typecheck

RerankerGuo:
## Description | 描述

The v2.0.0 Proxy package does not pass its own TypeScript check. The release
runtime already supports `memCommand` configuration and virtual default task
entries, but the corresponding raw-config and inferred task-list types were
not updated. Open-source checkouts also lack a declaration for the optional
private cost-guard submodule.

This change:

- adds the existing `memCommand` YAML shape to `RawYamlConfig`;
- types both session-init task arrays with the shared `TeamOption` contract so
  the existing `isDefault` marker remains valid;
- declares the optional cost-guard export without requiring the submodule in
  an open-source checkout;
- adds focused coverage for loading and filtering `memCommand` configuration.

Runtime configuration, session selection, and optional-module loading behavior
are unchanged.

## Related Issue | 关联 Issue

L4 v2.0.0 Proxy build/typecheck audit finding; no existing issue.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [ ] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

## Verification | 验证方式

- `npm test -- --run src/__tests__/config-mem-command.test.ts` (1/1)
- `npm test` (1/1 collected Proxy test)
- `npx tsc --noEmit`
- `npm pack --dry-run --json` (149 files)
- `git diff --check`

## Additional Notes | 其他说明

The default branch reported six errors across these boundaries before this
change. The full package typecheck now passes. No dependency or lockfile is
changed.

PR #742, opened 2026-08-03
fix(proxy): avoid re-recording stale user turns

RerankerGuo:
## Description | 描述

TDAI L0 recording should consider the newest user-role message in the current
request. When that message contains only coding-agent harness metadata,
`extractLatestUserMessage()` currently continues scanning backward and returns
an older real question. The current turn then records that previous question a
second time.

This change:

- stops after evaluating the newest user-role message;
- returns `null` when that message contains no human input;
- preserves extraction of the newest real user message when assistant messages
  follow it;
- adds focused regression tests for both cases.

The shared user-query cleanup rules and TDAI write path are unchanged.

## Related Issue | 关联 Issue

L3 MemoryProxy TDAI data-correctness audit finding; no existing issue.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证方式

- `npm test -- --run src/__tests__/tdai-recorder.test.ts` (2/2)
- `npm test` (2/2 collected Proxy tests)
- `git diff --check`

## Additional Notes | 其他说明

The default-branch full Proxy typecheck errors are repaired separately by
#741. This PR changes only TDAI user-turn selection and its regression test; it
does not change configuration, dependencies, or lockfiles.

### DIFF
--- a/MemoryProxy/src/__tests__/tdai-recorder.test.ts
+++ b/MemoryProxy/src/__tests__/tdai-recorder.test.ts
@@ -1,0 +1,27 @@
+import { describe, expect, it } from "vitest";
+import { extractLatestUserMessage } from "../tdai/recorder.js";
+
+describe("extractLatestUserMessage", () => {
+  it("does not fall back to an older turn when the latest user message is harness noise", () => {
+    const result = extractLatestUserMessage([
+      { role: "user", content: "earlier real question" },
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #743, opened 2026-08-03
fix(panel): scope user-key request deduplication

RerankerGuo:
## Description | 描述

The Panel deduplicates concurrent read requests to avoid duplicate network
calls from React StrictMode and concurrent component mounts. The user-key list
used one global in-flight key, `user-key/list`, independent of the active
Panel session.

If user A's request was still pending while the local session switched to user
B, B's call reused A's Promise instead of issuing a request with B's
credentials. This could render user A's key metadata in user B's session until
the view refreshed.

This change scopes only the user-key list dedupe key to:

- the selected memory instance; and
- the authenticated `user_id` (falling back to the current user key only for
  legacy/incomplete cached sessions without a user object).

Concurrent reads within one authenticated session still share one Promise.
Different users or instances now use independent pending requests. The
response format, backend API, authentication headers, and create/revoke paths
are unchanged.

## Related Issue | 关联 Issue

No linked issue. This was found during a v2.0.0 Panel authentication and
session-boundary audit.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification

- `pnpm exec vitest run tests/web-user-key-dedupe.test.ts`
  - 2 tests passed
- `pnpm test`
  - 2 tests passed
- `pnpm run typecheck`
  - passed
- `pnpm run build`
  - Panel backend build passed
- `npm run build` in `MemoryPanel/web`
  - TypeScript and Vite production build passed
- `npx eslint src/lib/api/users.ts`
  - passed
- `npm pack --dry-run --json --ignore-scripts`
  - 197 files verified
- `git diff --check`
  - passed

The cross-user regression test was first run against the unchanged
implementation and failed because two authenticated users produced only one
fetch. A second test locks the intended same-user coalescing behavior.

## Additional Notes | 其他说明

- Breaking change: none.
- A full Web lint run still reports the existing `AgentGrid.tsx` empty-block
  error and 97 unrelated warnings. The changed production file passes ESLint.
- The source change is limited to the user-key list request key. The generic
  dedupe helper and public instance-list deduplication remain unchanged.

### DIFF
--- a/MemoryPanel/tests/web-user-key-dedupe.test.ts
+++ b/MemoryPanel/tests/web-user-key-dedupe.test.ts
@@ -1,0 +1,127 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("../web/src/lib/error-message.ts", () => ({
+  formatApiErrorMessage: ({ message }: { message?: string }) =>
+    message ?? "API error",
+}));
+
+import { userKeysApi } from "../web/src/lib/api/users.js";
+import {
+  clearPanelSession,
+  setPanelSession,
+  type PanelSession,
+} from "../web/src/lib/panelSession.js";
+
+interface PendingRequest {
+  userKey: string;
+  resolve: (response: Response) => void;
+}
+
+function session(userId: string, userKey: string): PanelSession {
+  return {
+    instanceId: "instance-1",
+    userKey,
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #730, opened 2026-08-03
fix(knowledge): restore middleware typecheck

RerankerGuo:
## Description | 描述

`MemoryKnowledge` currently fails its own `pnpm typecheck` because the access-log middleware assigns `Promise<string>` and `Promise<unknown>` values directly to Hono's publicly typed `bodyCache` fields. Those writes are also unnecessary: `c.req.text()` already populates Hono's internal cache, and later `c.req.json()`/`c.req.text()` reads derive from that cached body.

This PR removes the direct internal-cache mutation and relies on Hono's request reader to own the cache lifecycle. Focused middleware tests prove that:

- a JSON request body remains readable by the downstream route after access logging inspects it;
- reading an error response for logging does not consume the response returned to the client.

## Related Issue | 关联 Issue

No existing issue. Discovered during a default-branch Knowledge build/typecheck audit.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [ ] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

## Verification | 验证

- `pnpm vitest run src/middleware/response-envelope.test.ts` - 1 file, 2 tests passed
- `pnpm test` - 1 file, 2 tests passed
- `pnpm typecheck` - passed
- strict changed-file TypeScript check - passed
- `pnpm build` - 11 output files built
- `git diff --check` - passed

## Additional Notes | 其他说明

- No route, envelope, logging format, dependency, or public API changes.
- The request body is still inspected once for error-log fields; Hono retains ownership of the cached representation.
- Full Open PR title/body and changed-file scans found no active implementation touching this middleware.

Signed-off-by: RerankerGuo <121015044+RerankerGuo@users.noreply.github.com>

PR #746, opened 2026-08-03
fix(knowledge): validate list pagination

RerankerGuo:
## Background

`POST /v3/wiki/list` and `POST /v3/code-graph/list` previously accepted any JavaScript number for `limit` and `offset`. Fractional, negative, or non-finite values could therefore reach Drizzle/SQLite instead of producing a deterministic client error.

## Changes

- Add one shared parser for Knowledge list pagination.
- Require `limit` to be a positive integer and `offset` to be a non-negative integer.
- Reject malformed pagination with an HTTP 400 envelope before calling either store.
- Document the same bounds in the Wiki and Code-Graph OpenAPI schemas.
- Add route-level coverage for both resource types and valid/invalid values.

The change is limited to pagination validation. Filtering, ordering, defaults, store queries, and response shapes remain unchanged.

## Verification

- [x] `npm test -- src/routes/list-pagination.test.ts` (16 tests passed)
- [x] `npm test` (16 tests passed)
- [x] Strict changed-file TypeScript compilation
- [x] `npm run build`
- [x] OpenAPI YAML parse
- [x] `npm pack --dry-run --ignore-scripts` (80 files)
- [x] `git diff --check`

Full-package `npm run typecheck` still reaches the default-branch error at `src/middleware/response-envelope.ts:47`, which is independently fixed by open PR #730. The strict compilation covering every file changed by this PR passes.

## Impact

- Breaking change: No for contract-valid requests
- Affected module: `MemoryKnowledge`
- Performance impact: None

## Checklist

- [x] The change follows the existing route validation style.
- [x] Regression tests cover both affected endpoints.

### DIFF
--- a/MemoryKnowledge/openapi.yaml
+++ b/MemoryKnowledge/openapi.yaml
@@ -833,9 +833,11 @@ components:
               enum: [draft, pending, processing, ready, failed]
             limit:
               type: integer
+              minimum: 1
               default: 20
             offset:
               type: integer
+              minimum: 0
               default: 0
 
     WikiListData:
@@ -1072,9 +1074,11 @@ components:
               enum: [pending, processing, ready, failed]
             limit:
               type: integer
+              minimum: 1
               default: 20
             offset:
               type: integer
+              minimum: 0
               default: 0
 
     CodeGraphListData:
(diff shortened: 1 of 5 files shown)


Maxwell-Code07:

PR #747, opened 2026-08-03
docs(sdk): clarify v3 session isolation

RerankerGuo:
## Background

The Python SDK implementation requires `team_id`, `agent_id`, and `user_id` when constructing a v3 `MemoryClient`. `session_id` is required only for `add_conversation`; query, search, count, and delete operations may omit it to aggregate across sessions.

Several package docstrings and documentation pages instead said that all four values were constructor requirements or that every L0/L1 call required a session. That guidance contradicted the implemented and documented client behavior.

## Changes

- Clarify the v3 constructor requirements in package and module docstrings.
- Document that L0 conversation writes require a session while reads may omit it.
- Correct the Chinese v2/v3 comparison table.
- Replace misleading "isolation quad" wording in MetadataClient documentation.

This PR changes documentation and docstrings only. It does not alter SDK behavior, request bodies, validation, or public signatures.

## Verification

- [x] Runtime contract check: construct v3 `MemoryClient` without a session and query across sessions
- [x] Runtime contract check: `add_conversation` without a session still fails
- [x] `python3 -m compileall -q sdk/memory-core/python/tencentdb_agent_memory`
- [x] `python3 -m pip wheel sdk/memory-core/python --no-deps`
- [x] `git diff --check`

## Impact

- Breaking change: No
- Affected module: Python SDK documentation
- Runtime impact: None

## Checklist

- [x] The text matches the current synchronous and asynchronous client contract.
- [x] English and Chinese management-plane wording is consistent.
- [x] The Python package compiles and builds as a wheel.
- [x] The PR contains one isolated documentation correction.

### DIFF
--- a/sdk/memory-core/python/AGENT_GUIDE.python.zh-CN.md
+++ b/sdk/memory-core/python/AGENT_GUIDE.python.zh-CN.md
@@ -222,7 +222,7 @@ except TDAMError as e:
 
 ## 7. 管理面：Knowledge / 元数据
 
-上面讲的都是 `MemoryClient`（数据面：读写记忆）。如果你还需要**管理** Knowledge 知识源（wiki / code-graph）的元数据，用 `MetadataClient`（v3 管理面，不需要 isolation 四元组）：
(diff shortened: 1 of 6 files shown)


Maxwell-Code07:

PR #748, opened 2026-08-03
fix(skill): strengthen archive task identifiers

RerankerGuo:
## Description | 描述

`SkillTriggerService.archive()` generated internal task IDs from only the
first eight hexadecimal characters of `randomUUID()`. That leaves a 32-bit
identity space:

- collision probability reaches roughly 1% around 9,300 generated IDs;
- it reaches roughly 50% around 77,000 generated IDs.

This is not only an observability concern. The worker completes a task by
filtering `_tasks.json` on `task_id`, so two same-agent tasks with a colliding
ID can be removed together even though only one was processed.

This change replaces the truncated UUID suffix with the existing
`randomBase62(12)` helper:

- approximately 71 bits of unbiased CSPRNG entropy;
- URL, storage-key, and log friendly alphanumeric output;
- the existing `skill-extract-task-` prefix is preserved.

A deterministic service-level test supplies two legacy UUIDs with the same
first eight characters and verifies that the queued task IDs remain distinct
and match the 12-character base62 format.

Queue ordering, archive keys, task payload fields, retry behavior, and worker
processing are unchanged.

## Related Issue | 关联 Issue

No linked issue. This was found during a v2.0.0 skill queue identity audit.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification

- `pnpm exec vitest run src/core/skill/conversation-add/trigger-service-task-id.test.ts`
  - 1 test passed
- `pnpm test`
  - 1 test passed
- `pnpm run build:plugin`
  - plugin bundle completed successfully
- `pnpm run lint:skill-isolation`
  - passed
- `npm pack --dry-run --json --ignore-scripts`
  - 293 files verified
- `git diff --check`
  - passed

The focused regression test was first run against the unchanged
implementation and failed with
`skill-extract-task-deadbeef`, directly reproducing the prefix collision.

## Additional Notes | 其他说明

- Breaking change: none. Task IDs are opaque strings; their documented prefix
  remains unchanged.
- Callers that incorrectly validate the old eight-hex-character suffix will
  need to stop parsing this internal identifier.
- Scope is limited to the task ID generator call and its regression test.

### DIFF
--- a/MemoryCore/src/core/skill/conversation-add/trigger-service-task-id.test.ts
+++ b/MemoryCore/src/core/skill/conversation-add/trigger-service-task-id.test.ts
@@ -1,0 +1,93 @@
+import { describe, expect, it, vi } from "vitest";
+
+const cryptoMocks = vi.hoisted(() => ({
+  randomUUID: vi.fn(),
+  randomBytes: vi.fn(),
+}));
+
+vi.mock("node:crypto", () => ({
+  randomUUID: cryptoMocks.randomUUID,
+  randomBytes: cryptoMocks.randomBytes,
+}));
+vi.mock("../../report/obs-logger.js", () => ({
+  obsLogger: { info: vi.fn() },
+}));
+
+import type { ISkillAgentTaskQueue } from "./agent-task-queue.js";
+import type {
+  AgentTasksDoc,
+  SessionKey,
+  SkillBufferStorage,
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #749, opened 2026-08-03
fix(sdk): allow clearing Python skill client defaults

RerankerGuo:
## Description | 描述

`SkillClient.with_defaults()` and `AsyncSkillClient.with_defaults()` currently
use `None` as the default value for every keyword argument. As a result, the
clone operation cannot distinguish an omitted argument from an explicit
request to clear an optional default:

```python
client.with_defaults(task_id=None)
```

retains the existing `task_id` instead of removing it from subsequent request
bodies. This can keep stale optional isolation fields attached when an
application reuses a client for a broader scope.

This change:

- introduces an internal sentinel for omitted `with_defaults()` arguments;
- preserves existing defaults only when the corresponding argument is omitted;
- treats explicit `None` as a request to clear that default;
- applies the same behavior to synchronous and asynchronous skill clients;
- documents the distinction on both public clone methods.

The approach matches the sentinel pattern already used by the Python v3 data
client for clearable isolation fields.

## Related Issue | 关联 Issue

L3/L4 Python SDK behavior audit finding; no existing issue or active PR modifies
the affected v3 Skill client files.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证方式

- `PYTHONPATH=. /usr/bin/python3 -m pytest -q tests/test_v3_skill_defaults.py`
  (2/2)
- `PYTHONPATH=. /usr/bin/python3 -m pytest -q` (2/2 collected tests)
- `/usr/bin/python3 -m compileall -q tencentdb_agent_memory tests`
- `/usr/bin/python3 -m pip check`
- `/usr/bin/python3 -m pip wheel --no-deps .` (wheel built and inspected)
- `git diff --check`

The regression tests exercise both sync and async clients. They verify that an
argument omitted from `with_defaults()` is inherited, while an explicitly
cleared value is absent from the next serialized `/v3/skill/list` request.

## Additional Notes | 其他说明

This PR changes only clone-time handling of optional Skill client defaults.
Constructor behavior, per-request override behavior, transport sharing,
dependencies, and package metadata are unchanged. Existing calls that omit
arguments retain their current behavior; explicit `None` now performs the
clear operation that the public optional-value API could not previously
express.

### DIFF
--- a/sdk/memory-core/python/tencentdb_agent_memory/v3/skill_client.py
+++ b/sdk/memory-core/python/tencentdb_agent_memory/v3/skill_client.py
@@ -33,6 +33,7 @@ from .._v3_http import AsyncHttpStub, HttpStub
 from ..errors import ParamError
 
 _V3 = "/v3/skill"
+_UNSET = object()
 
 # ── Numeric error codes returned in envelope.code for /v3/skill/*. ──
 SKILL_ERROR_CODE: Dict[str, int] = {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #750, opened 2026-08-03
fix(core): serialize and drain timer scanner runs

RerankerGuo:
## Description | 描述

`TimerScanner` currently starts its asynchronous `scan()` method directly from
`setInterval`. If one shard claim takes longer than the configured interval,
later ticks start additional scans against the same backend. Although the
Redis claim itself is atomic, overlapping scans multiply backend pressure and
make lifecycle accounting non-deterministic.

The same ownership gap affects shutdown: `stop()` clears the interval but does
not wait for a scan that has already started. It can therefore report that the
scanner stopped while the owned scan is still claiming timers and enqueueing
tasks.

This change:

- tracks the scanner's single owned in-flight scan;
- coalesces interval ticks while that scan is still running;
- makes `stop()` wait for the active scan after disabling new ticks;
- preserves completion of already-claimed timers before shutdown returns.

Completing the owned scan is intentional. Returning early or abandoning its
results after an atomic Redis claim would remove timers without enqueueing
their tasks.

## Related Issue | 关联 Issue

L3 MemoryCore service-lifecycle audit finding; no existing issue or active PR
modifies the TimerScanner implementation or tests.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证方式

- `npx vitest run src/services/timer-scanner.test.ts` (2/2)
- `npm test` (2/2 collected tests)
- `npm run build:plugin`
- `npm pack --dry-run --ignore-scripts` (293 files)
- `git diff --check`

The regression tests use a deferred shard claim to prove that:

1. multiple interval ticks issue only one claim while a scan is pending; and
2. `stop()` remains pending until the active scan enqueues an already-claimed
   timer.

## Additional Notes | 其他说明

This PR changes only TimerScanner scheduling and shutdown ownership. Timer
shard formats, claim batch size, task payload construction, backend APIs, and
the local-state fallback are unchanged.

### DIFF
--- a/MemoryCore/src/services/timer-scanner.test.ts
+++ b/MemoryCore/src/services/timer-scanner.test.ts
@@ -1,0 +1,82 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import type { IStateBackend, TimerEntry } from "../core/state/types.js";
+import { TimerScanner } from "./timer-scanner.js";
+
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((res) => {
+    resolve = res;
+  });
+  return { promise, resolve };
+}
+
+function createBackend(
+  claimExpiredFromShard: () => Promise<TimerEntry[]>,
+): IStateBackend {
+  return {
+    timerShardCount: 1,
+    claimExpiredFromShard,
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #751, opened 2026-08-03
fix(sdk): support internal COS URLs in Python

RerankerGuo:
## Description | 描述

The Python memory-file reader accepts only public COS URLs of the form:

```text
bucket.cos.<region>.myqcloud.com
```

However, STS can return the VPC/internal form used by MemoryCore and already
supported by the TypeScript SDK:

```text
bucket.cos-internal.<region>.tencentcos.cn
```

The Python `StsCredential` currently rejects that valid `CosUrl` before any
file request is attempted. It also reconstructs a public host from the parsed
bucket and region, which would be incorrect for signing an internal request.

This change:

- parses both public and internal COS hostname forms;
- preserves the exact normalized host returned by STS;
- uses that preserved host for the COS request URL and signed `Host` header;
- keeps existing bucket, region, prefix, expiry, and credential APIs intact.

## Related Issue | 关联 Issue

L3/L4 Python/TypeScript SDK compatibility audit finding; no existing issue or
active PR modifies the Python COS reader or its tests.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification | 验证方式

- `PYTHONPATH=. /usr/bin/python3 -m pytest -q tests/test_cos_internal_url.py`
  (3/3)
- `PYTHONPATH=. /usr/bin/python3 -m pytest -q` (3/3 collected tests)
- `/usr/bin/python3 -m compileall -q tencentdb_agent_memory tests`
- `/usr/bin/python3 -m pip check`
- `/usr/bin/python3 -m pip wheel --no-deps .` (wheel built and inspected)
- `git diff --check`

The regression tests verify:

1. public COS parsing remains unchanged;
2. the synchronous reader sends an internal URL and matching `Host` header;
3. the asynchronous reader preserves the same internal-host behavior.

## Additional Notes | 其他说明

This PR does not change credential refresh/coalescing, 403 retry behavior, COS
V5 signing, request timeouts, package dependencies, or server contracts. It
only brings the Python hostname boundary in line with the existing MemoryCore
and TypeScript SDK behavior.

### DIFF
--- a/sdk/memory-core/python/tencentdb_agent_memory/cos.py
+++ b/sdk/memory-core/python/tencentdb_agent_memory/cos.py
@@ -34,16 +34,24 @@ logger = logging.getLogger(__name__)
 # COS URL parser
 # ---------------------------------------------------------------------------
 
-def _parse_cos_url(cos_url: str) -> tuple[str, str]:
-    """Parse CosUrl like ``https://bucket.cos.region.myqcloud.com`` → (bucket, region)."""
+def _parse_cos_url(cos_url: str) -> tuple[str, str, str]:
+    """Parse a public or internal COS URL into ``(bucket, region, host)``."""
     host = urlparse(cos_url).hostname or ""
-    # Pattern: {bucket}.cos.{region}.myqcloud.com
-    m = re.match(r"^(.+?)\.cos\.(.+?)\.myqcloud\.com$", host)
-    if m:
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #752, opened 2026-08-03
fix(ts-sdk): avoid process-wide TLS disablement

RerankerGuo:
## Description | 描述

The TypeScript v2 transport currently calls `require("undici")` from an ESM
package. Because `require` is unavailable in ESM, construction falls into the
catch path and sets `NODE_TLS_REJECT_UNAUTHORIZED=0`, disabling TLS certificate
verification for every HTTPS request in the process.

This change:

- imports `Agent` from the SDK's existing `undici` runtime dependency;
- keeps the self-signed-certificate behavior scoped to this transport's
  dispatcher;
- adds regression tests for the default and `rejectUnauthorized: true` paths.

The public API and the existing v2 default (`rejectUnauthorized: false`) remain
unchanged.

## Related Issue | 关联 Issue

N/A - found while auditing the default branch's TypeScript v2 transport.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Verification:

- `npx vitest run tests/http-tls.test.ts` (2 passed)
- `npm test` (2 passed)
- `npm run build`
- `npm pack --dry-run` (44 files)
- `git diff --check`

No process environment variables are changed by the transport after this fix.

### DIFF
--- a/sdk/memory-core/typescript/src/http.ts
+++ b/sdk/memory-core/typescript/src/http.ts
@@ -4,10 +4,11 @@
  * - Auth: `Authorization: Bearer {apiKey}` + `x-tdai-service-id` + optional `x-tdai-user-key`
  * - Envelope unwrap: `code === 0` → return `data`; else throw `TDAMError`
  * - trace_id: extracted from `x-trace-id` response header
- * - Zero runtime dependencies — uses native `fetch`.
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #753, opened 2026-08-03
fix(python-sdk): validate v2 response envelopes

RerankerGuo:
## Description | 描述

The Python v2 sync and async transports duplicate response-envelope decoding
and assume both the top-level response and successful `data` value are
mappings. A successful response such as `{"code": 0, "data": []}` therefore
returns a list even though both transport methods declare an object result.

This change:

- shares successful-response decoding across the sync and async transports;
- requires an object envelope, an integer business code, and object success
  data while retaining `null` data as an empty object;
- preserves trace IDs, business error details, and request IDs;
- adds focused helper, sync transport, and async transport regression tests.

Non-2xx responses continue to use the existing `raise_for_status()` behavior.
Request serialization and public APIs are unchanged.

## Related Issue | 关联 Issue

N/A - found while auditing the default branch's Python v2 transport.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Verification:

- `python3 -m pytest -q` (13 passed)
- `python3 -m compileall -q tencentdb_agent_memory tests`
- `python -m build --wheel` (wheel built successfully)
- `git diff --check`

### DIFF
--- a/sdk/memory-core/python/tencentdb_agent_memory/_http.py
+++ b/sdk/memory-core/python/tencentdb_agent_memory/_http.py
@@ -34,6 +34,45 @@ class Stub(ABC):
         ...
 
 
+def _decode_success_response(resp: httpx.Response) -> dict:
+    request_id = (
+        resp.headers.get("x-qcloud-transaction-id")
+        or resp.headers.get("x-trace-id")
+        or ""
+    )
+    try:
+        envelope = resp.json()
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #175, opened 2026-06-10
fix(security): enable L1 prompt injection filtering

RerankerGuo:
## Related Issue | 关联 Issue
  Fix #158
## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [ ] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

PR #758, opened 2026-08-04
Fix variable interpolation in info messages

Desperado-Jia:
## Description | 描述

fix(scripts): wrap ADMIN_KEY_FILE variable in braces for bash 3.2 compatibility

Fix an "unbound variable" error when running under bash 3.2.57 (default on macOS).

In bash 3.2, variable parsing is not multibyte-safe. Placing a full-width  closing parenthesis '）' (UTF-8: `EF BC 89`) immediately after a unbraced variable  causes the parser to consume the leading byte (`0xEF`) as part of the variable  identifier (e.g., `ADMIN_KEY_FILE\xEF`). Under `set -u`, this triggers an unbound  variable fatal error.

Explicitly scoping the identifier as `${ADMIN_KEY_FILE}` resolves the parsing ambiguousness.

Fixes: #175 in start-memory-core.sh

## Change Type | 修改类型
- [✅] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

### DIFF
--- a/deploy/global-images/start-memory-core.sh
+++ b/deploy/global-images/start-memory-core.sh
@@ -121,7 +121,7 @@ skill:
     maxResourceSizeBytes: 5000000
 YAML
 
-info "启动 memory-core (image=$MEMORY_CORE_IMAGE, port=$MEMORY_CORE_PORT)"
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:

PR #773, opened 2026-08-04
fix(mcp): send x-tdai-service-id header + log to stderr

### LINKED ISSUE #766: MCP server cannot call its own HTTP API: http-client never sends the x-tdai-service-id header that every /v3 endpoint requires
The bundled MCP server cannot call the HTTP API that ships in the same package: `http-client.ts` never sends the `x-tdai-service-id` header, and every `/v3` endpoint requires it. As a result every `tools/call` fails.

**Version:** commit `f3df793` (2026-07-29)

## The API requires the header

`MemoryKnowledge/src/api-helpers.ts`:

```ts
12: export const SERVICE_ID_HEADER = "x-tdai-service-id";
```

`MemoryKnowledge/src/routes/code-graph.ts`:

```ts
 10: * 多租户（001）：`service_id` 每个端点必传于 `x-tdai-service-id` 请求头（与内核路由键统一）。
184:   const idFields = extractIdFields(c.req.header("x-tdai-service-id"), body);
```

## The MCP client never sends it

`MemoryKnowledge/src/mcp/http-client.ts`, `callApi()`:

```ts
const headers: Record<string, string> = { "Content-Type": "application/json" };
if (opts.token) headers["Authorization"] = `Bearer ${opts.token}`;
```

Only `Content-Type` and an optional bearer token. `x-tdai-service-id` is never set, and there is no option to supply it — so the MCP server as published can never satisfy its own API.

## Reproduction

1. Start the service and create a code graph via the HTTP API (works — you can pass the header yourself with `curl`).
2. Register the bundled MCP server and call any tool, e.g. `code_status`.
3. Every call fails with `x-tdai-service-id header is required`.

Adding `team_id` to the request body is not a workaround: the tool schemas are strict and reject it with `unexpected field: team_id`. Identity is only accepted via the header.

## Fix we are running locally

Read the service id from the MCP process environment and set the header when present, preserving current behaviour when it is absent:

```ts
const serviceId = process.env.KNOWLEDGE_SERVICE_ID;
if (serviceId) headers["x-tdai-service-id"] = serviceId;
```

This makes the published MCP server usable. Happy to send it as a PR if that shape is acceptable — or if you would rather thread the value through `HttpClientOptions` explicitly, that works too.

## Unrelated note that cost us time

The MCP server writes logs to **stdout**, which corrupts the JSON-RPC stream that the MCP transport reads from the same channel. Clients fail to parse the handshake until `LOG_LEVEL=error` is set. Logs should go to stderr for stdio-based MCP servers.

YOMXXX:
## Problem

The bundled MCP server can never call the knowledge API that ships in the same package: `MemoryKnowledge/src/mcp/http-client.ts` sends only `Content-Type` and an optional bearer token, while every `/v3` endpoint requires the `x-tdai-service-id` header (`api-helpers.ts` → `extractIdFields` returns null → 400). As a result **every MCP `tools/call` fails**.

A second, equally fatal issue surfaced while fixing the first: the logger writes `info`/`debug` to **stdout**, which is exactly the channel the MCP stdio transport reads JSON-RPC from — clients fail to parse the handshake unless `LOG_LEVEL=error`.

Fixes #766.

## Change

**Commit 1 — `fix(mcp): send x-tdai-service-id header from KNOWLEDGE_SERVICE_ID`**
- `HttpClientOptions` gains an optional `serviceId`, sent as `x-tdai-service-id` when present (absent keeps previous behaviour — backward compatible)
- MCP server startup reads `KNOWLEDGE_SERVICE_ID` from the environment
- documented in `.env.example` and `README.md`

**Commit 2 — `fix(mcp): write diagnostic logs to stderr for stdio safety`**
- `logger.ts`: `debug`/`info` now write to stderr; stdout carries only the MCP protocol. (HTTP-server logging is unaffected functionally — stderr is the standard diagnostic channel.)

## Tests

New vitest files (first tests for this package):

- `src/mcp/http-client.test.ts` — 7 tests: header sent when `serviceId` present, omitted when absent, bearer token forwarding, trailing-slash normalization, HTTP-error / non-zero-code / invalid-JSON handling.
- `src/logger.test.ts` — 2 tests: `info`/`debug` go to stderr, never stdout.

`npm test`: 2 files, 9 tests, all passing.

## Note

`npm run typecheck` reports one **pre-existing** error in `src/middleware/response-envelope.ts(47,9)` (`Promise<string>` not assignable to `string`) — unrelated to this change and present on `feat/server_team`. Happy to follow up separately if useful.

### DIFF
--- a/MemoryKnowledge/.env.example
+++ b/MemoryKnowledge/.env.example
@@ -32,6 +32,9 @@ LLM_BASE_URL=
 # MCP server 通过 HTTP 调用本服务 API，这里配置转发目标
 KNOWLEDGE_API_URL=http://localhost:8421
 # KNOWLEDGE_API_TOKEN=optional-bearer-token
+# service_id 租户身份：随 x-tdai-service-id header 发送，每个 /v3 端点必传。
+# 未设置时 MCP 工具的 HTTP 调用会被服务端以 400 拒绝。
+KNOWLEDGE_SERVICE_ID=
 
 # ═══ 与管控（team-memory-control）对接 ═══
 # TMC + KS 部署时以下两项必须显式配置。代码中的默认值均为空：
--- a/MemoryKnowledge/README.md
+++ b/MemoryKnowledge/README.md
@@ -96,6 +96,19 @@ pnpm test
 pnpm build        # tsdown → dist/
 ```
 
+## MCP stdio
+
+MCP server 通过 HTTP 转发工具调用到本服务，启动时需把租户身份 `service_id` 一并注入：
+
+```bash
+KNOWLEDGE_API_URL=http://localhost:8421 \
+KNOWLEDGE_SERVICE_ID=<tenant-id> \
+pnpm dev:mcp
+```
+
+`KNOWLEDGE_SERVICE_ID` 会以 `x-tdai-service-id` 请求头发送（每个 `/v3` 端点必传，缺失返回 400）。  
+MCP 是 stdio 协议，诊断日志统一走 stderr，stdout 仅承载 JSON-RPC。
+
 ## 可选：Langfuse
 
 配置 `LANGFUSE_PUBLIC_KEY` + `LANGFUSE_SECRET_KEY`（及可选 `LANGFUSE_BASE_URL`）即可上报 Wiki LLM 调用。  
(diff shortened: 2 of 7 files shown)


Maxwell-Code07:

Issue #760, opened 2026-08-04
v2.0.0 release tag but MemoryCore/package.json still reports 2.0.0-beta.1


ConradLu2740:
### Summary

The repo is tagged `v2.0.0` (commit 0aff21a, 2026-08-03) and CHANGELOG.md documents a `2.0.0` release, but `MemoryCore/package.json` still has:

```json
"version": "2.0.0-beta.1"
```

### Impact

Fresh installs print `@tencentdb-agent-memory/memory-tencentdb-v2@2.0.0-beta.1`; any future npm publish from this tag would ship the stable release under a beta SemVer, which is confusing for consumers.

### Suggested fix

Bump `MemoryCore/package.json` version to `2.0.0` in the release commit (and mirror it to the SDK packages if applicable).

Maxwell-Code07:
Thank you for submitting this issue. We have documented your feedback, and the team will schedule an internal review and assessment. If you need a temporary workaround, you may refer to our official documentation for self-testing. We will post updates here once we make progress.

Issue #759, opened 2026-08-04
npm run build fails on v2.0.0: build:seed-v2 references removed scripts/seed-v2 directory


ConradLu2740:
### Summary

On v2.0.0 (commit 0aff21a, feat/server_team), running `npm run build` in `MemoryCore/` fails:

1. `build:plugin` (tsdown) succeeds and produces `dist/index.mjs`.
2. Then `build:scripts` runs `build:seed-v2` → `tsc -p scripts/seed-v2/tsconfig.json`.
3. `scripts/seed-v2/` no longer exists in v2.0.0 (directory was removed), so tsc exits with:

```
error TS5058: The specified path does not exist: 'scripts/seed-v2/tsconfig.json'
```

### Impact

Any local build or CI that runs `npm run build` fails on the v2.0.0 tag; users must bypass with `npm run build:plugin`.

### Suggested fix

Remove `build:seed-v2` (and the now-unused `seed-v2` npm script) from `MemoryCore/package.json`, or restore the `scripts/seed-v2/` directory.

### Environment

Node v22.23.2, npm 10.9.2, fresh `npm install` on v2.0.0 tag.

Maxwell-Code07:
Thank you for submitting this issue. We have documented your feedback, and the team will schedule an internal review and assessment. If you need a temporary workaround, you may refer to our official documentation for self-testing. We will post updates here once we make progress.

PR #775, opened 2026-08-04
fix(build): Dockerfile missing files, dead seed-v2 build, version alignment

### LINKED ISSUE #765: docker build fails on a clean clone: Dockerfile COPYs three files that are not in the repository (MemoryKnowledge)
`docker build` for MemoryKnowledge fails on a clean clone: the Dockerfile `COPY`s three paths that do not exist in the repository.

**Version:** commit `f3df793` (2026-07-29)

## The three missing paths

`MemoryKnowledge/Dockerfile`:

```dockerfile
35: COPY docs/api/openapi.yaml ./docs/api/openapi.yaml
59: COPY docker/entrypoint.sh /usr/local/bin/docker-entrypoint.sh
60: COPY docker/smoke-test.sh /usr/local/bin/smoke-test.sh
```

None of them are tracked. `git ls-tree --name-only HEAD MemoryKnowledge/` returns:

```
MemoryKnowledge/.dockerignore
MemoryKnowledge/.env.example
MemoryKnowledge/.npmrc
MemoryKnowledge/Dockerfile
MemoryKnowledge/README.md
MemoryKnowledge/bin
MemoryKnowledge/docker-compose.yml
MemoryKnowledge/drizzle.config.ts
MemoryKnowledge/package.json
MemoryKnowledge/src
MemoryKnowledge/start.sh
MemoryKnowledge/tsconfig.json
MemoryKnowledge/tsdown.config.ts
```

There is no `docker/` directory and no `docs/` directory. The build therefore fails at line 35 for every clean clone.

`ENTRYPOINT` also depends on one of the missing files:

```dockerfile
75: ENTRYPOINT ["/usr/bin/tini", "--", "/usr/local/bin/docker-entrypoint.sh"]
```

so even if the `COPY` lines were skipped, the resulting image would have no entrypoint.

## Why this is blocking rather than cosmetic

The documented local build path is not a workaround: `npm install` fails on the `better-sqlite3` native module on a stock Windows toolchain, which is what pushed us to Docker in the first place. With the Docker path broken as well, there is no working install route out of the box.

## Two smaller issues found while working around this

1. The builder stage installs apt packages from `mirrors.tencent.com`, which is not reachable outside Tencent's network. A public default (e.g. `deb.debian.org`) would make the image buildable elsewhere.
2. `bin/` is not copied into the runtime stage, and the entrypoint refers to `dist/server.js` while the build actually emits `dist/server.mjs`.

## Suggested fix

Either commit the three missing files, or drop the `COPY` lines and the `tini`/entrypoint indirection and start the server directly (`ENTRYPOINT ["node"]`, `CMD ["/app/dist/server.mjs"]` works — that is what we run in production).

Happy to open a PR if a maintainer confirms which direction is preferred.

YOMXXX:
Fixes #765, #759, #760 — three build/release blockers on the v2.0.0 tag.

## #765 — `docker build` fails on a clean clone

`MemoryKnowledge/Dockerfile` COPYs `docker/entrypoint.sh` and `docker/smoke-test.sh`, which were not in the repository (the `openapi.yaml` COPY had already been fixed upstream). Added both:

- `docker/entrypoint.sh` — forwards the CMD (`["start"]`) / CLI args to the built server: `node dist/server.mjs` (the tsdown ESM output; `bin/*.mjs` launchers are for npm global installs and not part of the image)
- `docker/smoke-test.sh` — curls `/health` for image smoke testing

## #759 — `npm run build` fails: `build:seed-v2` references removed `scripts/seed-v2/`

Removed the dead `build:seed-v2` / `seed-v2` scripts (and their slot in `build:scripts`), deleted `bin/seed-v2.mjs`, and updated the README seed reference to the current `openclaw memory-tdai seed --input <file>` CLI command. Verified `npm run build` now passes.

## #760 — v2.0.0 tag but `package.json` still `2.0.0-beta.1`

Bumped `MemoryCore/package.json` to `2.0.0`.

## Verification

- `MemoryCore`: `npm run build` passes (no seed-v2 step)
- `MemoryKnowledge`: `tsdown` build passes; both scripts pass `sh -n`

## Additional findings (not changed, for maintainers)

- `MemoryKnowledge/package.json` `main: "./dist/server.js"` and `bin/server.mjs` (`import "../dist/server.js"`) point to a file that does not exist — `tsdown` emits `dist/server.mjs` (ESM + `fixedExtension`). npm consumers / `bin` entrypoints would fail; the Docker image is unaffected (entrypoint uses `dist/server.mjs` directly).
- `MemoryKnowledge/src/server.ts` reads config from env only; the CLI flags shown in `Dockerfile` comments / `docker-compose.yml` (`--public-url`, `--tmc-callback`, …) are not parsed by the server. That docker-compose invocation path looks broken independent of this PR.
- `pr-ci.yml` only triggers on PRs to `main`, not the current default branch `feat/server_team`.


Maxwell-Code07:

PR #777, opened 2026-08-04
fix(embedding): enable provider=local and warm up the local model

### LINKED ISSUE #678: Local embedding provider (provider=local) is unreachable from config and startWarmup() is never called

YOMXXX:
Fixes #678.

## Problem

`LocalEmbeddingService` (node-llama-cpp, embeddinggemma-300m, offline, 768-dim) was fully implemented but unreachable:

1. `config.ts` coerced `provider=local` to `provider=none` + `embeddingEnabled=false` ("Local embedding provider is not available in user config")
2. the sqlite store factories created an embedding service only when `enabled && provider !== "local" && apiKey`
3. nothing ever called `startWarmup()` — the GGUF model was never downloaded, and `embed()` threw `EmbeddingNotReadyError`, silently degrading recall to keyword-only

## Change

- **`config.ts`**: `provider=local` now keeps `embeddingEnabled=true` (offline, no apiKey needed). Model downloads lazily on `startWarmup()`; until ready, `embed()` throws `EmbeddingNotReadyError` and recall falls back to keyword-only.
- **`store/factory.ts` + `store-pool.ts`** (sqlite branches): create the embedding service for `provider=local` (remote providers still require apiKey; `none` stays disabled).
- **`pipeline-factory.ts` (`initStores`) + `store-pool.ts` (`createSqliteStore`)**: call `startWarmup()` at startup — triggers background model download/load, a safe no-op for remote providers.

## Tests

- `src/config.test.ts` (3): `provider=local` enables embedding; `none`/default stays disabled
- `src/core/store/factory.test.ts` (2): `createStoreBundle` builds an embedding service for `local`, skips `none`

`npm test`: 2 files, 5 tests, all passing. `npm run build:plugin` passes.

## Notes

- `node-llama-cpp` is an optional peerDependency; if absent, `startWarmup()` fails to load and recall degrades to keyword-only (existing behaviour).
- The deploy yaml (`start-memory-core.sh`) still emits `embedding.provider: none`; users can switch to offline embedding by setting `provider: local` in their gateway config.

### DIFF
--- a/MemoryCore/src/config.test.ts
+++ b/MemoryCore/src/config.test.ts
@@ -1,0 +1,27 @@
+/**
+ * Tests for #678 — parseConfig must accept provider="local" for fully-offline
+ * embedding (node-llama-cpp), instead of coercing it to "none" + disabled.
+ */
+
+import { describe, expect, it } from "vitest";
+import { parseConfig } from "./config.js";
+
+describe("parseConfig embedding provider (#678)", () => {
+  it("defaults to provider=none (embedding disabled)", () => {
+    const cfg = parseConfig({});
+    expect(cfg.embedding.provider).toBe("none");
+    expect(cfg.embedding.enabled).toBe(false);
+  });
+
+  it("keeps provider=none disabled when explicitly set", () => {
+    const cfg = parseConfig({ embedding: { provider: "none" } });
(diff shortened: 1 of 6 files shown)


Maxwell-Code07:

PR #780, opened 2026-08-04
docs(health): surface FTS5 availability and document the requirement

### LINKED ISSUE #679: Docs: default SQLite backend's keyword search requires FTS5, which is not present in every official Node build
### Summary

The default SQLite backend uses `node:sqlite` FTS5 virtual tables for BM25 keyword search (e.g. `bm25(l1_fts)`). However, official Node binaries are not guaranteed to include the FTS5 module:

- Node v22.13.1 (official darwin-arm64): `CREATE VIRTUAL TABLE ... USING fts5` → `no such module: fts5`. Keyword search silently unavailable (`strategy: embedding` only).
- Node v22.23.2 (official darwin-arm64): FTS5 available, `strategy: hybrid` works.

### Suggested improvement

Document the FTS5 requirement (and version/build caveat) in the README, plus what recall degrades to when FTS5 is absent (embedding-only). A startup log line already warns about this — consider also surfacing it in `/health`.

YOMXXX:
Fixes #679.

## Problem

BM25 keyword / hybrid recall depends on SQLite's **FTS5** module, but the official Node binaries are not guaranteed to include it (e.g. v22.13.1 → `no such module: fts5`; v22.23.2 → works). When absent, keyword search silently returns nothing and recall degrades to embedding-only, which is hard to diagnose.

## Change

- **`GET /health`** now reports `stores.ftsAvailable`, sourced from the store's existing `isFtsAvailable()` (which already tracks whether FTS5 compiled in).
- **README / README_CN** document the FTS5 requirement, the Node build caveat, and the embedding-only fallback.

`npm run build:plugin` passes.

### DIFF
--- a/MemoryCore/README.md
+++ b/MemoryCore/README.md
@@ -52,6 +52,7 @@ MemoryCore is distributed as a standalone runtime for local development, single-
 - Node.js `>= 22.16.0`
 - npm
 - An OpenAI-compatible LLM API. Read-only queries may not invoke an LLM, but memory extraction and aggregation require valid credentials.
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

PR #781, opened 2026-08-04
feat(examples): Claude Code hooks for subscription-auth users

### LINKED ISSUE #715: Feature: hook-based integration example for coding agents on subscription auth (no ANTHROPIC_BASE_URL override possible)
### Context
The documented Claude Code integration relies on rerouting `ANTHROPIC_BASE_URL` through memory-proxy. That works for API-key users, but Claude Code sessions authenticated via subscription (OAuth) cannot be pointed at a proxy — overriding the base URL breaks subscription auth entirely. Those users currently have no supported way to feed conversations into TencentDB Agent Memory.

### Working pattern
I integrated a subscription-authenticated Claude Code with the platform using two small shell hooks (no proxy involved, answers keep flowing through the subscription):
- **Stop hook**: extracts the last user/assistant exchange from the session transcript and `POST /v2/conversation/add` (Bearer = business user key). Fail-silent, `curl --max-time 4`.
- **SessionStart hook**: `POST /v2/core/read` + `/v2/scenario/ls`, prints a compact L3-profile block that Claude Code injects into session context.

The L0→L1 pipeline picks conversations up exactly as with the proxy path (verified: atoms extracted).

### Proposal
Ship this as a documented example (e.g. `examples/claude-code-hooks/`) for subscription-auth users. Happy to contribute my scripts as a starting point.

Related wish: a `Dockerfile` for memory-core so security-conscious users can build the full stack from source (proxy and hub already have one).

YOMXXX:
Closes #715.

## Problem

The documented Claude Code integration reroutes `ANTHROPIC_BASE_URL` through memory-proxy — that works for API-key users but **breaks subscription (OAuth) auth** entirely. Subscription-authenticated users had no supported way to feed conversations into TencentDB Agent Memory.

## Change

A documented hook example at `examples/claude-code-hooks/`:

- **`stop-hook.sh`** (`Stop` hook): reads the hook payload from stdin, extracts the last user/assistant exchange, `POST /v2/conversation/add` (Bearer = business user key). Verified extraction keeps the final user turn + its nearest assistant reply, honours optional team/agent/user isolation, and falls back to `session_id` → `default`.
- **`session-start-hook.sh`** (`SessionStart` hook): `POST /v2/core/read` + `/v2/scenario/ls`, prints a compact `<user-profile>` / `<known-scenes>` block that Claude Code injects into session context.
- **`README.md`**: env vars (`TD_MEMORY_URL` / `TD_MEMORY_KEY` / optional isolation), `settings.json` registration, verification steps.

Both hooks are **fail-silent** (`set +e`, `curl --max-time 4`) — never block, error, or slow down Claude Code.

## Verification

- `sh -n` passes for both scripts
- The JSON extraction (stop) and response parsing (session-start) logic verified against sample payloads (last-exchange extraction, isolation passthrough, empty-scene handling)

Related wish from the issue (a `memory-core` Dockerfile) is not part of this PR.

### DIFF
--- a/examples/claude-code-hooks/README.md
+++ b/examples/claude-code-hooks/README.md
@@ -1,0 +1,78 @@
+# Claude Code hooks — TencentDB Agent Memory for subscription-auth users
+
+> **适用人群**：Claude Code 会话通过 **订阅（OAuth）** 认证的用户。
+
+官方集成通过把 `ANTHROPIC_BASE_URL` 重定向到 memory-proxy 工作，但这会**破坏订阅鉴权**——OAuth 会话无法被指向代理。本示例用两个轻量 shell hooks 直连 Memory Gateway 的 v2 API，**对话仍然走订阅，记忆照常沉淀**。
+
+工作原理与 API-key 代理路径一致（已验证 L0→L1 抽取正常）：
+
+| Hook | 时机 | 作用 |
+|------|------|------|
+| `stop-hook.sh` | 每轮对话结束（`Stop`） | 从 session transcript 提取最后 user/assistant 一轮，`POST /v2/conversation/add` 写入 L0 |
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

Issue #784, opened 2026-08-04
MemoryKnowledge typecheck fails: response-envelope.ts Promise<string> not assignable to string


YOMXXX:
## Problem

`npm run typecheck` in `MemoryKnowledge/` fails with a pre-existing error:

```
src/middleware/response-envelope.ts(47,9): error TS2322:
  Type 'Promise<string>' is not assignable to type 'string'.
```

## Repro

```bash
cd MemoryKnowledge
npm run typecheck   # tsc --noEmit
```

Present on `feat/server_team` (v2.0.0) before any local changes.

## Impact

- `typecheck` is listed in `MemoryKnowledge/README.md` as a standard command but always fails
- Masks new type errors on that module (reviewers must diff the known failure)

## Suggested fix

Await or restructure the assignment at `src/middleware/response-envelope.ts:47`.

Maxwell-Code07:
Thank you for submitting this issue. We have documented your feedback, and the team will schedule an internal review and assessment. If you need a temporary workaround, you may refer to our official documentation for self-testing. We will post updates here once we make progress.

Maxwell-Code07:
Thank you for submitting this issue. We have documented your feedback, and the team will schedule an internal review and assessment. If you need a temporary workaround, you may refer to our official documentation for self-testing. We will post updates here once we make progress.

aliabbas6622:
Hi, I’d like to work on this issue. I’ll reproduce the typecheck failure, inspect the async flow in `response-envelope.ts`, and submit a focused fix with verification. Is this issue available for contribution?

Issue #783, opened 2026-08-04
MemoryKnowledge main/bin point at dist/server.js which tsdown never emits (it produces server.mjs)


YOMXXX:
## Problem

`MemoryKnowledge/package.json` declares `"main": "./dist/server.js"` and ships `bin/server.mjs` whose body is `import "../dist/server.js"`. But `tsdown` (with `format: "esm"` + `fixedExtension: true`) emits **`dist/server.mjs`** — `dist/server.js` does not exist.

Verified on `feat/server_team` (v2.0.0, commit `0aff21a`):

- `npm run build` (tsdown) → `dist/server.mjs` + `dist/mcp/server.mjs`, no `.js` entry
- `ls dist/server.js` → `No such file or directory`
- `node bin/server.mjs` → `Cannot find module` for `../dist/server.js`

## Impact

- npm consumers importing the package (`main`) resolve a missing file
- `knowledge-server` / `knowledge-mcp` bin launchers fail
- The Docker image is **unaffected** because `docker/entrypoint.sh` calls `node dist/server.mjs` directly

## Suggested fix

Point `main` at `./dist/server.mjs`, and fix `bin/server.mjs` / `bin/mcp.mjs` to `import "../dist/server.mjs"` (and `../dist/mcp/server.mjs` respectively). Alternatively drop `fixedExtension` in `tsdown.config.ts` so `.js` entries are emitted as declared.

Maxwell-Code07:
Thank you for submitting this issue. We have documented your feedback, and the team will schedule an internal review and assessment. If you need a temporary workaround, you may refer to our official documentation for self-testing. We will post updates here once we make progress.

PR #785, opened 2026-08-04
fix: address upstream CI/build issues (#782 #783 #784)

### LINKED ISSUE #782: CI never runs on PRs to feat/server_team (pr-ci.yml only triggers on main)
## Problem

`MemoryCore/.github/workflows/pr-ci.yml` only triggers on pull requests targeting `main`:

```yaml
on:
  pull_request:
    branches: [main]
```

But the repository's current **default branch is `feat/server_team`** (releases, including v2.0.0, are tagged from it). Every PR targeting `feat/server_team` therefore **skips CI entirely** — `gh pr checks` reports nothing, and the pack validation, manifest check, size guard, and skill-queue isolation guard never run.

## Impact

- No automated verification on the main contribution path
- Build-breaking regressions (e.g. #759, #765) can land uncaught
- Package size limits and skill-isolation rules are unenforced on PRs

## Suggested fix

Trigger on the default branch too, e.g.:

```yaml
on:
  pull_request:
    branches: [main, feat/server_team]
```

or switch to a wildcard that covers the release branch.

YOMXXX:
Fixes #782, #783, #784 — three upstream issues, plus the dead `build:seed-v2` that blocks CI on the release branch.

## #782 — CI never runs on PRs to `feat/server_team`

`pr-ci.yml` only triggered on PRs to `main`, but the default/release branch is `feat/server_team`. Trigger on both branches so pack validation, manifest checks, size guard, and skill-isolation guard actually run on PRs.

## #783 — `main` / bin launchers point at `dist/server.js` (tsdown emits `.mjs`)

- `package.json` `main` → `./dist/server.mjs`
- `bin/server.mjs` → `import "../dist/server.mjs"`
- `bin/mcp.mjs` → `import "../dist/mcp/server.mjs"`

npm consumers and `knowledge-server` / `knowledge-mcp` bins now resolve the real build output.

## #784 — `typecheck` fails in `response-envelope.ts`

`bodyCache.text` is typed `string` in Hono but holds a `Promise` at runtime. Cast the cached value (`as unknown as string`) to keep runtime behaviour and fix the TS error.

## (extra) — dead `build:seed-v2` breaks CI Pack

With CI now enabled (#782), the Pack job surfaced a pre-existing build break: `build:scripts` still runs `build:seed-v2`, which references the removed `scripts/seed-v2/` (`TS5058`). Dropped the dead script so this branch's CI is green (same change as #775, needed here because #782 now triggers CI on `feat/server_team`).

## Verification

- `MemoryKnowledge`: `npm run build` → emits `dist/server.mjs`; `npm run typecheck` → exit 0
- `node --check` on both bin launchers
- `pr-ci.yml` parses with `branches: ["main", "feat/server_team"]`
- `MemoryCore`: `npm run build:scripts` passes (seed-v2 removed)

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -2,7 +2,9 @@ name: CI
 
 on:
   pull_request:
-    branches: [main]
+    # The repo's default / release branch (feat/server_team) is where every
+    # PR lands — main is a legacy branch. Trigger CI on both (issue #782).
+    branches: [main, feat/server_team]
 
 # Cancel previous runs for the same PR / branch
 concurrency:
--- a/MemoryCore/package.json
+++ b/MemoryCore/package.json
@@ -18,7 +18,7 @@
   "scripts": {
     "build": "npm run build:plugin && npm run build:scripts",
     "build:plugin": "tsdown",
-    "build:scripts": "npm run build:migrate-sqlite-to-vdb && npm run build:export-tencent-vdb && npm run build:read-local-memory && npm run build:seed-v2",
+    "build:scripts": "npm run build:migrate-sqlite-to-vdb && npm run build:export-tencent-vdb && npm run build:read-local-memory",
     "prepack": "npm run build",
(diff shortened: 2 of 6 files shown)


Maxwell-Code07:

PR #807, opened 2026-08-05
 feat(proxy): support custom headers for upstream requests

IntegerAlex:
Some OpenAI-compatible APIs need more than just a URL and API key they require extra headers on every request. OpenCode's Zen API is the immediate case: it needs x-opencode-request, x-opencode-session, and a
  specific User-Agent to authenticate alongside the standard Bearer token.

  Adds a headers field at both the global upstream level and per-agent level in config, so any such API can be supported without touching the handler code.
 The AgentAdapter interface also gets an optional headers() method for adapters that need to generate dynamic per-request values.

  Merge priority: global headers → per-agent headers → adapter dynamic headers
  → auth (auth always wins)


Maxwell-Code07:

PR #810, opened 2026-08-05
fix(hermes): bound Gateway stdout/stderr logs with a runtime rotation guard

### LINKED ISSUE #583: [Bug] Hermes Gateway stdout/stderr logs grow without runtime bounds

YOMXXX:
Fixes #583.

## Problem

`GatewaySupervisor` opens `gateway.stdout.log` / `gateway.stderr.log` in append mode and keeps those descriptors for the child's lifetime, with **no size limit or rotation**. A high-frequency logging defect can grow the files without bound and exhaust disk — observed **~63 GiB** in one incident.

## Change

A runtime log-size guard (per the issue's suggested design):

- daemon thread polls each active stream every `250 ms`
- when a stream exceeds **`LOG_ACTIVE_MAX_BYTES` (10 MiB)**, the last **`LOG_TAIL_KEEP_BYTES` (1 MiB)** is preserved as `<path>.1` (bounded diagnostic tail) and the active file is **truncated in place** — keeping the child's inherited descriptor valid (renaming the path would orphan the old inode)
- guard is stopped/joined before log handles close (in both `shutdown()` and `_close_log_handles()`)

## Tests

`MemoryCore/hermes-plugin/tests/test_log_rotation.py` (unittest):
- oversized file → truncated in place, final bytes preserved in `.1`
- undersized file → untouched, no `.1` created

`python3 -m unittest tests.test_log_rotation` passes.

### DIFF
--- a/MemoryCore/hermes-plugin/memory/memory_tencentdb/supervisor.py
+++ b/MemoryCore/hermes-plugin/memory/memory_tencentdb/supervisor.py
@@ -40,6 +40,13 @@ HEALTH_CHECK_RETRIES = 3     # retries for is_running check
 
 # Log file rotation parameters
 LOG_TAIL_BYTES_ON_CRASH = 2048  # bytes of stderr log to surface on startup crash
+# Runtime disk bound for Gateway stdout/stderr logs (issue #583): when an
+# active stream exceeds LOG_ACTIVE_MAX_BYTES, keep the last
+# LOG_TAIL_KEEP_BYTES as a bounded diagnostic tail in "<path>.1" and truncate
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #814, opened 2026-08-05
fix(memory-core): recover SerialQueue after synchronous task throw

### LINKED ISSUE #518: Bug: SerialQueue deadlocks after a synchronous task throw

Hotragn:
## Description | 描述

Ports the #518 fix to the current default branch. `SerialQueue.drain()` in `MemoryCore/src/utils/serial-queue.ts` invokes `entry.task()` bare, so a task that throws **synchronously** (before returning its promise) escapes ahead of the `.finally()` bookkeeping: the caller's promise rejects, but the queue is left with `running = true` forever. All three pipeline queues (`l1Queue`/`l2Queue`/`l3Queue`) share this class, so one bad task silently stalls every later task, and `flushSession()` / `onIdle()` waiters hang until `destroy()`'s timeout fallback.

Fix: start task invocation from `Promise.resolve().then(() => entry.task())`, so synchronous throws and async rejections share the existing rejection + cleanup path. FIFO ordering, error propagation to the caller, and idle notification semantics are unchanged.

Credit where due: this is the same fix shape as @yangjj-iso's #519, which targets `main`'s pre-v2 layout (`src/utils/serial-queue.ts` at the repo root) and does not apply to `feat/server_team`'s `MemoryCore/` tree. The identical bug in MemoryKnowledge's separate `SerialQueue` was already addressed by #694 — this PR closes the remaining copy.

Regression tests (`MemoryCore/src/utils/serial-queue.test.ts`, vitest, no mocks):
- FIFO ordering and async-failure recovery (baseline behaviour lock);
- a synchronously-throwing task rejects its caller, the queue finishes bookkeeping, the **next task actually runs**, and `onIdle()` resolves — before the fix these hang (verified: with the fix stashed, the new tests time out at 120s);
- `onIdle()` waiters resolve when a sync throw is followed by queued tasks.

## Related Issue | 关联 Issue

Fix #518 (port of #519 to feat/server_team; MemoryKnowledge twin fixed in #694)

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

`npm test` passes (Node v24, Windows 11). Negative check: with the fix stashed and only the tests applied, the regression tests hang and fail on timeout, confirming they capture #518 exactly.

## Additional Notes | 其他说明

None — two-file change, no dependency or API surface changes.

### DIFF
--- a/MemoryCore/src/utils/serial-queue.test.ts
+++ b/MemoryCore/src/utils/serial-queue.test.ts
@@ -1,0 +1,80 @@
+import { describe, expect, it } from "vitest";
+
+import { SerialQueue } from "./serial-queue.js";
+
+describe("SerialQueue", () => {
+  it("runs tasks serially in FIFO order", async () => {
+    const queue = new SerialQueue("fifo");
+    const order: number[] = [];
+    await Promise.all([
+      queue.add(async () => {
+        await new Promise((r) => setTimeout(r, 10));
+        order.push(1);
+      }),
+      queue.add(async () => {
+        order.push(2);
+      }),
+      queue.add(async () => {
+        order.push(3);
+      }),
+    ]);
+    expect(order).toEqual([1, 2, 3]);
+    // running=false is flipped in the drain chain's .finally, one microtask
+    // after the last caller promise resolves — wait for idle, don't poll it.
+    await queue.onIdle();
+    expect(queue.idle).toBe(true);
+  });
+
+  it("rejects the caller on async task failure and keeps draining", async () => {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Thank you for your contribution! We will review this PR as soon as possible and get back to you shortly.

PR #821, opened 2026-08-06
fix(memory-core): sqlite queryL1Records ignores recordIds filter

mittalpk:
## Description | 描述
`VectorStore.queryL1Records()` in `MemoryCore/src/core/store/sqlite.ts` never reads `filter.recordIds`, even though the `L1QueryFilter` interface documents it ("Query by document primary keys") and the TCVDB backend (`tcvdb.ts`) handles it correctly. With no other filter field set, it falls through to a full-table scan ordered by `updated_time ASC` and returns everything.

This matters because `handleAtomicUpdate()` in `gateway/v2-router.ts` does `store.queryL1Records({ recordIds: [id] })` to read a record by primary key before updating it, with a comment saying exactly that. On the SQLite backend, `existing[0]` is actually the oldest record in the entire store, not the one matching `id`:
- a lookup for a nonexistent `id` never returns 404,
- an update can silently copy `type`/`priority`/`scene_name`/timestamps/`team_id`/`user_id`/`agent_id` from an unrelated record before upserting under the caller-supplied `id`.

Fix adds a dedicated primary-key lookup branch (same approach as `tcvdb.ts`'s `documentIds` handling), given priority over the session-based branches, so SQLite and TCVDB behave the same way for this filter.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复

## Self-test Checklist | 自测清单

Added `MemoryCore/src/core/store/sqlite.recordids.test.ts`. Confirmed all 3 new tests fail against the pre-fix code (`git stash`) with exactly the described wrong-record/no-404 behavior, and pass after the fix. `npx vitest run` and `npx tsdown` (the `build:plugin` step) both clean.

## Additional Notes | 其他说明
`npm run build:seed-v2` fails on this branch even without my change (`scripts/seed-v2/tsconfig.json` doesn't exist) — confirmed pre-existing on `feat/server_team` HEAD via `git stash`, unrelated to this PR.

### DIFF
--- a/MemoryCore/src/core/store/sqlite.recordids.test.ts
+++ b/MemoryCore/src/core/store/sqlite.recordids.test.ts
@@ -1,0 +1,69 @@
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+import type { MemoryRecord } from "../record/l1-writer";
+import { VectorStore } from "./sqlite";
+
+function makeRecord(id: string, overrides: Partial<MemoryRecord> = {}): MemoryRecord {
+  return {
+    id,
+    content: `content-${id}`,
+    type: "episodic",
+    priority: 50,
+    scene_name: "",
+    source_message_ids: [],
+    metadata: {},
+    timestamps: ["2026-01-01T00:00:00.000Z"],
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:
Thank you for your contribution! We will review this PR as soon as possible and get back to you shortly.

mittalpk:
Still available to make any changes here if that would help move review along.

PR #834, opened 2026-08-07
fix(compact): stop infinite loop when tool-result truncation frees no tokens

### LINKED ISSUE #832: [Bug] Infinite loop in tool-result truncation during compact

N/A — self-hosted deployment

### Plugin Version | 插件版本

v1.0.1 & v2.0.0

### Operating System | 操作系统

Linux

_No response_

### Describe the bug | 问题描述

`truncateTailToolResults` and `truncateRemainingToolResults` can spin forever,
pinning the Node main thread at 100% CPU until the `compact` request times out.

**Root cause.** Truncation appends a notice but does not count it against the
character budget (`compressor.ts:251`):

```ts
const truncated = oldContent.slice(0, truncateChars) +
  `\n\n[... content truncated, only first ${truncateChars} characters retained ...]`;
```

With the default `TOOL_RESULT_TRUNCATE_CHARS = 2000`, the result is 2066
characters. The selection loop picks messages *longer than* the threshold
(`compressor.ts:239`):

```ts
if (content.length <= truncateChars) continue; // already small enough
```

2066 > 2000, so the same message is selected again on the next iteration. From
the second round on, the content is a fixed point — slicing to 2000 and
appending the same notice yields the identical string — so `newTokens` stops
changing and `freed` becomes 0:

```ts
const freed = tokenArray[maxIdx] - newTokens;   // 0 from round 2 onward
tokensToFree -= freed;                          // never decreases
```

`while (tokensToFree > 0)` (lines 231 and 281) then never terminates.

This is reachable whenever a single tool result cannot free enough tokens on
its own, which in our workload happens with large tool outputs in the protected
tail.

### To Reproduce | 复现步骤

1. Build a message array with one `tool_result` whose text is slightly longer
   than `TOOL_RESULT_TRUNCATE_CHARS` (e.g. 2500 characters).
2. Call `truncateTailToolResults(messages, tokenArray, 0, 100000, estimateTokens)`
   — i.e. ask for far more tokens than that single message can release.
3. The call never returns; CPU goes to 100%.

Deterministic version that fails fast instead of hanging the test runner —
inject a counting estimator:

```ts
function countingEstimator(maxCalls: number) {
  let calls = 0;
  return (text: string) => {
    if (++calls > maxCalls) {
      throw new Error(`estimateTokens called ${calls} times — truncation is not converging`);
    }
    return Math.max(1, Math.ceil(text.length / 4));
  };
}

const messages = [toolResult(TOOL_RESULT_TRUNCATE_CHARS + 500)];
const tokenArray = [Math.ceil((TOOL_RESULT_TRUNCATE_CHARS + 500) / 4)];
truncateTailToolResults(messages, tokenArray, 0, 100_000, countingEstimator(20));
```

Result on v2.0.0:

```
Error: estimateTokens called 21 times — truncation is not converging
```

And asserting the post-truncation length:

```
AssertionError: expected 2066 to be less than or equal to 2000
```

In production this surfaced as the Memory Node main thread stuck at 100% CPU
and `compact` requests timing out.

### Expected behavior | 预期行为

Truncation should always make progress and terminate:

- a message that has already been truncated should not be selected again;
- the truncated output, notice included, should fit within `truncateChars`;
- when a round frees no tokens, the loop should still move forward rather than
  re-pick the same message.

No error is logged — the process simply stops making progress. Symptoms are
100% CPU on the main thread and the `compact` HTTP request timing out.

### Additional context | 补充信息

_No response_

YOMXXX:
Fixes #832.

## Problem

`truncateTailToolResults` / `truncateRemainingToolResults` could **spin forever at 100% CPU** until the `compact` request timed out. Root cause: the truncation appended a notice without counting it against the budget, so the result (`2000 + ~66` chars) stayed **above** `TOOL_RESULT_TRUNCATE_CHARS`; the selection loop (`content.length <= truncateChars → skip`) re-picked the same message every pass, `freed` became `0`, and `tokensToFree` never decreased → `while (tokensToFree > 0)` never terminated.

## Change

Both functions:

- **count the notice against the budget**: slice to `truncateChars - notice.length`, so the truncated result (including the notice) is **at or under** `truncateChars` and is not re-selected
- **defensively break when a round frees `<= 0` tokens** (a fixed point) so the loop can never spin

## Tests

`src/offload_server/compact/compressor.test.ts` (first tests for the module):
- a single oversized `tool_result` terminates and stays within budget (the issue's repro: `tokensToFree=100000`, content 2500 chars)
- already-small tool results untouched

`npm test` passes, `npm run build:plugin` passes.

### DIFF
--- a/MemoryCore/src/offload_server/compact/compressor.test.ts
+++ b/MemoryCore/src/offload_server/compact/compressor.test.ts
@@ -1,0 +1,47 @@
+/**
+ * Regression tests for #832 — tool-result truncation must terminate instead
+ * of spinning at 100% CPU when a single message cannot free enough tokens.
+ */
+
+import { describe, expect, it } from "vitest";
+import { truncateTailToolResults } from "./compressor.js";
+import type { Message } from "./helpers.js";
+
+function toolResult(content: string): Message {
+  return {
+    role: "user",
+    content: [{ type: "tool_result", content }],
+  } as unknown as Message;
+}
+
+function getToolResultContent(msg: Message): string {
+  const blocks = (msg as any).content as { type: string; content: string }[];
+  return blocks[0]?.content ?? "";
+}
+
+describe("truncateTailToolResults (#832)", () => {
+  it("terminates when a single message cannot free enough tokens", () => {
+    const content = "x".repeat(2500); // > default TOOL_RESULT_TRUNCATE_CHARS (2000)
+    const messages = [toolResult(content)];
+    const tokenArray = [Math.ceil(content.length / 4)]; // ~625
+    const estimate = (t: string) => Math.max(1, Math.ceil(t.length / 4));
+
+    // Would previously loop forever (freed=0 after round 1) until timeout.
+    const freed = truncateTailToolResults(messages, tokenArray, 0, 100_000, estimate);
+
+    expect(freed).toBeGreaterThan(0);
+    // Truncated result (including the notice) stays within the budget, so the
+    // same message is not re-selected on the next pass.
+    const result = getToolResultContent(messages[0]);
+    expect(result.length).toBeLessThanOrEqual(2000);
+    expect(result).toContain("content truncated");
+  });
+
+  it("leaves already-small tool results untouched", () => {
+    const content = "short";
+    const messages = [toolResult(content)];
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #841, opened 2026-08-07
fix(proxy): fallback conversation id for CodeBuddy session

dong-frank:
## Problem

CodeBuddy client versions `4.10.2`–`4.10.4` do not send a `sessionId`/conversation header on requests. When such a request reaches the MemoryProxy, the session cannot be initialized: `sessionId=none` / `conversationId=null`, so the conversation never gets recorded in memory. The current documentation (`INSTALL.md`) only instructs users to avoid these broken versions.

## Root cause

The proxy's session resolution depends entirely on an incoming session header. When the header is absent, `resolveConversationId` returns `null` and the request falls through without a session key, so no memory write can be associated with a conversation.

## Change

Add a fallback that recovers the conversation id from the client instead of relying solely on the header:

- `deploy/global-images/hooks/session_start.py` — a CodeBuddy `SessionStart` hook that injects a `[tdai-proxy-session] conversation_id: <id>` marker into the prompt.
- `MemoryProxy/src/session/session-key.ts` — new `extractConversationIdFromPrompt` parses the marker (with a generic fallback); `stripConversationMarker` removes it before forwarding so the upstream LLM never sees the injected text.
- `MemoryProxy/src/handler.ts` / `MemoryProxy/src/anthropicHandler.ts` — wire extraction + stripping into both OpenAI- and Anthropic-style request paths.
- `INSTALL.md` / `INSTALL_CN.md` — reframe the CodeBuddy known-bug workaround from an avoid-only limitation into a documented fallback (optional `SessionStart` hook setup), with anchor links between the two docs.

## Tests

- Pure-function unit tests: `session-key.ts` extraction + stripping, 13/13 pass (`npx tsx`).
- End-to-end checks against a local fake LLM across three scenarios:
  - OpenAI request with marker → `conversationId` extracted, marker stripped.
  - OpenAI request without marker (control) → `null` / body unchanged (no false positives).
  - Anthropic request with marker → `conversationId` extracted, marker stripped.

## Related Issue

N/A — resolves a documented known limitation (CodeBuddy `4.10.2`–`4.10.4` missing session header) rather than a filed issue.

## Change Type

- [x] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Code optimization

## Self-test Checklist

- [x] Verified locally

## Additional Notes

- Split into two commits on `dev/codebuddy-adaptation`:
  - `ea1209c fix(proxy): fallback conversation id from SessionStart hook prompt` — proxy code + hook script.
  - `1ba5075 docs(proxy): document SessionStart hook workaround for broken CodeBuddy versions` — `INSTALL.md` / `INSTALL_CN.md`.
- Both commits carry a DCO `Signed-off-by: dong-frank <1057762929@qq.com>`.
- The hook directory artifacts (`__pycache__/`, `session_start.log`) are excluded via `.gitignore`.
- Target branch: `feat/server_team`.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -280,13 +280,15 @@ Disable the full pipeline (passthrough only): `PROXY_FULL_STACK=0 ./start-proxy.
 
 [CodeBuddy](https://www.codebuddy.ai/) is Tencent's AI coding assistant IDE plugin. By configuring a custom model, you can route CodeBuddy's chat requests through the Proxy to get the same memory capabilities as Claude Code, directly within your IDE.
 
-### ⚠️ Version Restrictions
+### ⚠️ Known Session-Init Limitation
 
-> CodeBuddy versions **4.10.2, 4.10.3, and 4.10.4** have a known bug: these
-> versions do not send a `sessionId` in requests, preventing the Proxy from
-> completing session initialization.
->
-> **Use CodeBuddy ≥ 4.10.5 or ≤ 4.10.1.**
+CodeBuddy versions **4.10.2, 4.10.3, and 4.10.4** have a known bug: these
+versions do not send a `sessionId` / conversation header in requests, preventing
+the Proxy from completing session initialization. The Proxy mitigates this via a
+[`SessionStart` hook](https://www.codebuddy.ai/docs/zh/ide/User-guide/Hooks) that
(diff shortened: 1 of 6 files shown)


Maxwell-Code07:

dong-frank:
使用最新版codebuddy 即可解决

PR #853, opened 2026-08-07
fix(compact): compress protected tail after emergency

### LINKED ISSUE #838: [Bug] compressProtectedTail is never called after emergency compact

N/A — self-hosted deployment

### Plugin Version | 插件版本

v1.0.0, v1.0.1 & v2.0.0

### Operating System | 操作系统

Linux

_No response_

### Describe the bug | 问题描述

When a *single* user turn fans out into a long agent STEP (many
assistant / tool_result pairs), emergency compaction can finish still above
the target budget. The over-budget prompt is then returned to the caller.

The codebase already has an escape hatch for this case —
`compressProtectedTail` in
`MemoryCore/src/offload_server/compact/compressor.ts` (exported since v1.0.0,
~line 378; on v1 the path is `src/offload_server/compact/compressor.ts`).
Its docstring describes exactly this scenario:

```
 * Compress the protected tail zone (last user message and after).
 * Strategy (ordered by priority):
 *   1. Truncate oversized tool_results ...
 *   2. If still above target, delete oldest tool pairs in the tail
 *      while preserving the last user message itself and at least the
 *      most recent tool pair
```

But nothing calls it. On tags `v1.0.0`, `v1.0.1`, and `v2.0.0`,
`git grep compressProtectedTail` only hits the definition.
`compaction-handler.ts` Step 6 (~lines 214–220 on v2.0.0) ends after
`emergencyCompress` and never falls back:

```ts
  if (level === "emergency") {
    const emTargetTokens = Math.floor(contextWindow * (config.aggressiveCompressRatio - 0.10));
    const em = emergencyCompress(messages, emTargetTokens, tokenArray, aggRemainingTokens);
    report.emergencyDeleted = em.deletedCount;
    aggRemainingTokens = em.remainingTokens;
    compactState.deletedOffloadIds.push(...em.deletedIds);
  }
```

**Why emergency alone is not enough.** Both `aggressiveCompress` and
`emergencyCompress` treat everything from the last user message onward as a
protected tail and refuse to delete inside it. Emergency Phase 1 truncates
oversized `tool_result`s anywhere in the array; Phase 2 only deletes
*before* `protectedStart`. Once truncation is exhausted, a long current
turn (the protected zone itself) still exceeds `emTargetTokens`, and the
handler returns without further work.

Protecting the last user turn is intentional. Leaving
`compressProtectedTail` unused, with no over-budget check after emergency,
looks like incomplete wiring rather than a deliberate "never touch the
current turn" policy — the function's delete-oldest-tool-pairs phase only
exists for the case where the protected zone itself is the problem.

### To Reproduce | 复现步骤

1. Build a compact request whose messages are essentially one user turn + a
   long chain of assistant/tool pairs whose total tokens already sit above
   `contextWindow * (aggressiveCompressRatio - 0.10)` even after truncating
   every `tool_result` to `TOOL_RESULT_TRUNCATE_CHARS`.
2. `POST /v2/offload/compact` with `totalTokens` / `contextWindow` such that
   `resolveLevel` returns `"emergency"`.
3. Observe the response: `report.emergencyDeleted` may be 0 (or only head
   deletions from earlier turns, of which there are none), and remaining
   tokens / ratio stay above the emergency target.

### Expected behavior | 预期行为

After `emergencyCompress`, if `remainingTokens > emTargetTokens`, the
handler should call the already-exported `compressProtectedTail` to delete
the oldest complete tool-call groups inside the protected tail, keeping the
last user message and (by default) the most recent tool pair, until the
budget is met or no further progress is possible.

No hard error — compact returns 200 with an over-budget message list.
Downstream symptoms are context-window overflows / provider 400s on the next
LLM call, or the host having to drop the turn.

### Additional context | 补充信息

_No response_

iroiro147:
## Summary

- Wire the existing protected-tail compressor into emergency compaction when the emergency target is still exceeded.
- Preserve deleted tool-result IDs and emergency deletion reporting for the additional tail compression.

## Issue

Closes #838

## Verification

- `git apply --check --recount` passed for the reviewed artifact.
- `git apply --recount` applied cleanly.
- `git diff --check` passed.
- local verification not run; relying on upstream CI (this fresh worktree has no installed dependencies).

## Notes

The change is limited to `MemoryCore/src/offload_server/compact/compaction-handler.ts`.

### DIFF
--- a/MemoryCore/src/offload_server/compact/compaction-handler.ts
+++ b/MemoryCore/src/offload_server/compact/compaction-handler.ts
@@ -14,7 +14,7 @@ import { CompactionRequestSchemaV2 } from "../schemas.js";
 import { buildOffloadBasePath } from "../session-utils.js";
 import { applyFastPath } from "./fast-path.js";
 import { injectActiveMmd, injectHistoryMmds } from "./mmd-injector.js";
-import { resolveLevel, mildCompress, aggressiveCompress, emergencyCompress } from "./compressor.js";
+import { resolveLevel, mildCompress, aggressiveCompress, emergencyCompress, compressProtectedTail } from "./compressor.js";
 import { estimateMessageTokens, extractToolResultId } from "./helpers.js";
 import type { Message } from "./helpers.js";
 import { traceServerCompaction } from "../opik-tracer.js";
@@ -217,6 +217,17 @@ export async function handleCompaction(
     report.emergencyDeleted = em.deletedCount;
     aggRemainingTokens = em.remainingTokens;
     compactState.deletedOffloadIds.push(...em.deletedIds);
+
+    // If tokens remain above the emergency target, compress the protected tail:
+    // delete the oldest complete tool-call groups inside the tail while
+    // preserving its documented invariants.
+    if (aggRemainingTokens > emTargetTokens) {
+      const tailDeletedIds: string[] = [];
+      const tail = compressProtectedTail(messages, tokenArray, emTargetTokens, aggRemainingTokens, tailDeletedIds);
+      report.emergencyDeleted += tail.deletedCount;
+      aggRemainingTokens = tail.remainingTokens;
+      compactState.deletedOffloadIds.push(...tailDeletedIds);
+    }
   }
 
   // Step 7: Inject active MMD (after all compression, so position is correct)


Maxwell-Code07:

PR #854, opened 2026-08-07
fix: prevent tool-result truncation loop

iroiro147:
## Summary

- Ensure truncated tool-result content, including its notice, stays within the configured character budget.
- Reuse the bounded truncation helper in all compact truncation paths so already-truncated messages cannot cause a non-progressing loop.
- Addresses TencentCloud/TencentDB-Agent-Memory#832.

## Verification

- Artifact validated with `git apply --check --recount` and applied with `git apply --recount`.
- `git diff --check` passed.
- local verification not run; relying on upstream CI (fresh worktree has no installed dependencies).


Maxwell-Code07:

PR #856, opened 2026-08-08
fix(metadata): bind acl/check and list-accessible subject to caller

SashaMIT:
## Problem

Follow-up to #848 (same auth model gap, different root cause).

`handleV3MetaRoute` authenticates `x-tdai-user-key` into a `V3AuthContext`, but `acl/check` and `asset/list-accessible` discard that context (`_c`) and resolve the permission subject from the request body via `resolveUserId(user_id | user_key)`:

```ts
[`${V3_PREFIX}/asset/list-accessible`]: bind(..., (d, _c, s) => s.listAccessibleAssets(d)),
[`${V3_PREFIX}/acl/check`]: bind(..., async (d, _c, s) => s.checkAssetPermission(d)),
```

Consequence: any holder of a valid user key can pass another user's `user_id` and:

1. **`asset/list-accessible`**, dump every asset that user can read, including `visibility: "private"` rows they own, with `content_ref` / `source_ref` / `metadata_json`.
2. **`acl/check`**, evaluate that user's permissions against arbitrary `asset_id` values (permission / ownership oracle).

This contradicts the caller-binding already enforced on writes (`grantAclForCaller` requires `granted_by === caller`) and on user-key routes (`assertUserScope`).

## Fix

- `checkAssetPermissionForCaller` / `listAccessibleAssetsForCaller` take identity from `V3AuthContext`.
- If the body still supplies `user_id` / `user_key` (schema still requires one for honest clients), it must match the caller; otherwise `permission_denied`.
- Handlers now pass `c` instead of discarding it.

No behavior change for legitimate callers who send their own id.

## Verification

Local stub-store repro against the real service methods:

- Pre-fix: attacker body `user_id=victim` on `listAccessibleAssets` returns the victim's private asset (`content_ref` included); `checkAssetPermission` returns `{ allowed: true, reason: "owner" }`.
- Post-fix: same spoof on `*ForCaller` throws `permission_denied`; attacker self-subject sees zero assets / `not_team_member` on the victim private asset.

## Scope note

Same `_c` IDOR pattern remains on `team/get`, `agent/get`, `task/get`, `agent/list`, `task/list`, `task-agent/list`, and `agent-fixed-asset/*` (notably `list-with-detail` returns `agent.prompt`). Left out to keep this PR reviewable; happy to follow up.

### DIFF
--- a/MemoryCore/src/metadata/router/v3-meta-router.ts
+++ b/MemoryCore/src/metadata/router/v3-meta-router.ts
@@ -246,8 +246,8 @@ const routeTable: Record<string, Handler> = {
     const { team_id, limit, offset, ...filter } = d;
     return s.listAssetsByTeam(team_id, resolvePagination({ limit, offset }), filter);
   }),
-  [`${V3_PREFIX}/asset/list-accessible`]: bind(S.assetListAccessibleSchema, (d, _c, s) =>
-    s.listAccessibleAssets(d)),
+  [`${V3_PREFIX}/asset/list-accessible`]: bind(S.assetListAccessibleSchema, (d, c, s) =>
+    s.listAccessibleAssetsForCaller(d, c)),
 
   [`${V3_PREFIX}/asset/touch-usage`]: bind(S.assetTouchUsageSchema, async (d, c, s) => {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #857, opened 2026-08-08
fix(metadata): authorize v3 team/agent/task get by caller

SashaMIT:
## Problem

Follow-up to #848 / #856 (same `_c` authz gap on entity reads).

`handleV3MetaRoute` authenticates `x-tdai-user-key` into a `V3AuthContext`, and write handlers already route through caller-scoped wrappers. The read handlers for `team/get`, `agent/get`, and `task/get` instead discarded the context (`_c`) and called the raw store getters:

```ts
// before
[`${V3_PREFIX}/team/get`]:  bind(..., async (d, _c, s) => orNotFound(await s.getTeamById(d.team_id), ...)),
[`${V3_PREFIX}/agent/get`]: bind(..., async (d, _c, s) => orNotFound(await s.getAgentById(d.agent_id), ...)),
[`${V3_PREFIX}/task/get`]:  bind(..., async (d, _c, s) => orNotFound(await s.getTaskById(d.task_id), ...)),
```

Consequence: any holder of a valid user key can `POST /v3/meta/agent/get` with an arbitrary `agent_id` and receive the full agent row, including `prompt`, plus team/task metadata for any id. That bypasses the membership model already enforced on writes and on `team-member/*` reads.

## Fix

Three caller-scoped getters, mirroring the `getAssetForCaller` / `listTeamMembersForCaller` pattern:

- `getTeamForCaller` / `getTaskForCaller` — require active membership of the entity's team; denial returns `null` so the router renders `*_not_found` (no existence oracle).
- `getAgentForCaller` — same membership gate; `visibility: "private"` is owner-only (aligned with asset private semantics). Owner short-circuit still works.

Handlers now pass `c` instead of discarding it. Legitimate teammates and owners see identical results for team-visible agents.

## Verification

Local stub-store repro against the real service methods:

- Pre-fix: outsider `getAgentById` returns the victim prompt (`SYSTEM: ... xyzzy`).
- Post-fix: outsider `*ForCaller` returns `null` for team/agent/task.
- Teammate still reads team-visibility agent (prompt intact); private agent denied to non-owner teammate; owner reads private agent.

## Scope note

Same `_c` IDOR pattern remains on `agent/list`, `task/list`, `task-agent/list`, and `agent-fixed-asset/*` (notably `list-with-detail` returns `agent.prompt` + bound asset metadata). Left out to keep this PR reviewable; happy to follow up.

### DIFF
--- a/MemoryCore/src/metadata/router/v3-meta-router.ts
+++ b/MemoryCore/src/metadata/router/v3-meta-router.ts
@@ -128,7 +128,7 @@ const routeTable: Record<string, Handler> = {
 
   // Team
   [`${V3_PREFIX}/team/create`]: bind(S.teamCreateSchema, (d, c, s) => s.createTeamForCaller(d, c)),
-  [`${V3_PREFIX}/team/get`]: bind(S.teamGetSchema, async (d, _c, s) => orNotFound(await s.getTeamById(d.team_id), "team_not_found", d.team_id)),
+  [`${V3_PREFIX}/team/get`]: bind(S.teamGetSchema, async (d, c, s) => orNotFound(await s.getTeamForCaller(d.team_id, c), "team_not_found", d.team_id)),
   [`${V3_PREFIX}/team/update`]: bind(S.teamUpdateSchema, async (d, c, s) => {
     const { team_id, ...patch } = d;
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #872, opened 2026-08-08
fix(gateway): fail loudly when conversation delete removes nothing

### LINKED ISSUE #871: [Bug] POST /v2/conversation/delete is a silent no-op — returns code 0 / "ok" and deletes nothing (both session_id and message_ids forms)
## Summary

`POST /v2/conversation/delete` reports success and deletes nothing. It returns `code: 0`, `message: "ok"`, `data.deleted_count: 0`, and the messages remain queryable afterwards. This happens for **both** documented request forms — by `session_id` and by explicit `message_ids`.

Because the endpoint reports success, a caller has no signal that anything went wrong. In practice this means **L0 is write-only: there is no way to un-write a mistake.** The only recovery path we found is `./stop-all.sh --purge`, which destroys the entire memory volume.

## Environment

- Deploy: `deploy/global-images` three-in-one stack, `agentmemory/memory-core:latest` (digest `sha256:f9b28624…`), pulled 2026-08-08
- Host: macOS 15 (arm64), OrbStack, Docker 29.4.0
- Config: stock `.env.example` with `MEMORY_LLM_*` pointed at DeepSeek; `MEMORY_CORE_GATEWAY_API_KEY` empty (Bearer gate off, as shipped)
- Auth: admin `user_key` from `.admin-key`, sent as both `x-tdai-user-key` and `Authorization: Bearer`

## Reproduction

Four calls, using the same isolation quadruple throughout (`team_id`, `user_id`, `agent_id`), against a throwaway session:

```
$ POST /v2/conversation/add   (1 message)
  -> {"accepted_ids": ["msg-3374c3ddfe03"], "accepted_versions": ["v1"], "total_count": 1}   (code=0 message=ok)

$ POST /v2/conversation/query
  -> total=1  id=msg-3374c3ddfe03

$ POST /v2/conversation/delete  {session_id}
  -> {"deleted_count": 0}   (code=0 message=ok)
  -> query again: total=1

$ POST /v2/conversation/delete  {message_ids}
  -> {"deleted_count": 0}   (code=0 message=ok)
  -> query again: total=1
```

The row is also still present on disk inside the container afterwards:

```
$ docker exec tdai-memory-core grep -c "msg-3374c3ddfe03" /data/tdai-memory/conversations/2026-08-08.jsonl
1
```

## Expected vs actual

**Expected** — one of:
1. the messages are deleted and `deleted_count` reflects the number removed; or
2. if delete is not supported against the append-only L0 log, the endpoint returns a non-zero `code` with a message saying so.

**Actual** — `code: 0`, `"ok"`, `deleted_count: 0`, nothing deleted, no indication to the caller.

## Likely cause

L0 appears to be an append-only JSONL log (`/data/tdai-memory/conversations/<date>.jsonl`). Delete looks unimplemented against that store, but the handler still returns the success envelope. `deleted_count: 0` is arguably the honest part of the response — the problem is that `code: 0` sits next to it, and `0` is also the legitimate result of "matched nothing", so a caller cannot distinguish "nothing matched" from "delete is not implemented".

## Impact

We hit this seeding a real corpus. An earlier replay run hit the (undocumented) 100-messages-per-call cap partway through a session, so a partial write landed. Re-running the corrected script re-wrote the full session — and because delete silently did nothing, we could not remove the partial write. That session now holds **719 rows for 427 real turns**, permanently, in an otherwise clean store.

For a memory product this is more serious than an ordinary CRUD gap: a memory store that cannot be corrected will eventually hold something wrong and hold it forever. Correction is arguably a core operation, not a convenience.

## Suggested fix

Short term — make the failure visible: return a non-zero `code` (or an explicit `supported: false`) when the configured L0 backend cannot delete, so callers fail loudly instead of silently trusting the no-op.

Longer term — implement delete against the append-only log, e.g. tombstone records honoured by `/v2/conversation/query` and by the L1/L2/L3 extraction workers, with compaction on restart. Related closed request #116 asked for view/edit/delete over L1 memories; this issue is the L0 layer underneath that, and the same argument applies.

## Note on a related endpoint

`/v3/wiki/delete` is documented as never returning 409 and cleaning up resources. We did not test it, so this report is scoped strictly to `/v2/conversation/delete`.

YOMXXX:
Fixes #871.

## Problem

`POST /v2/conversation/delete` returned `code: 0` / `"ok"` with `deleted_count: 0` even when **nothing was removed** — L0 is an append-only log and the delete either didn't match (isolation mismatch) or isn't supported. A caller had no signal that the delete was a no-op, so **L0 became effectively write-only**: a mistake could not be un-written (only `--purge` destroyed everything).

## Change

An explicit delete (`message_ids` or `session_id`) that removes **nothing** now returns a **non-zero code (400)** with a message explaining that nothing was deleted — callers fail loudly instead of silently trusting the no-op. Idempotent deletes that match no records still surface the failure.

This is the issue's short-term failure-visibility fix; tombstones honoured by query/L1-L3 workers + compaction remain a longer-term follow-up.

## Verification

`npm run build:plugin` passes.

### DIFF
--- a/MemoryCore/src/gateway/v2-router.ts
+++ b/MemoryCore/src/gateway/v2-router.ts
@@ -985,6 +985,22 @@ async function handleConversationDelete(body: unknown, auth: V2AuthContext, requ
     }
   }
 
+  // Issue #871: an explicit delete that removed nothing must not silently
+  // return code 0 — the caller cannot distinguish "nothing matched" from
+  // "delete not implemented / isolation mismatch". Fail loudly so a memory
+  // store that cannot be corrected is never silently trusted.
+  if (deletedCount === 0 && (message_ids?.length || session_id)) {
+    return errorEnvelope(
+      400,
+      `No messages deleted: L0 backend did not remove ` +
+      (message_ids?.length
+        ? `${message_ids.length} message(s)`
+        : `session "${session_id}"`) +
+      ` (verify isolation ids match the records, or that delete is supported)`,
+      requestId,
+    );
+  }
+
   // Report memory deletion (non-fatal)
   if (deps.quotaManager && deletedCount > 0) {
     deps.quotaManager.reportMemoryDeleted(auth.serviceId, deletedCount).catch(() => {});


Maxwell-Code07:

Maxwell-Code07:

YuhaoLin2005:
**Review: this surfaces the failure, but the message_ids case may now 400 on legitimately-owned records**

Thanks for adding failure visibility — a silent no-op on an append-only store is genuinely bad. However, I believe the root cause of the message_ids case is a default-isolation mismatch that this PR doesn't address, so it risks converting a silent no-op into a **false** failure for valid deletes.

Chain:
1. `resolveIsolation` (`v2-schemas.ts:326-352`) fills a missing `session_id` with `DEFAULT_ISOLATION_ID`, so `ctx.sessionId = "default"` when the request carries no session.
2. `handleConversationDelete` passes the full ctx to `deleteL0` / `deleteL0BySession`, where `rowMatchesIsolation` (`isolation.ts:159-170`) does an exact `row.session_id !== filter.sessionId` comparison.
3. A `message_ids`-only delete (body has no `session_id`) therefore fails isolation for every record whose `session_id` is a real session — `deleteL0` returns `false`, even though the same `userId`/`agentId` owns the record and the record exists.

Precedent: `handleConversationSearch` already works around exactly this by stripping `sessionId` from the isolation filter (`v2-router.ts:893-896`, "避免默认 sessionId='default' 错误过滤真实 session 数据").

Suggestion: mirror that — build the delete filter from explicit dims only (`teamId`/`userId`/`agentId`/`taskId`, no `sessionId`), since `message_ids` / `session_id` params already scope the operation. Then legitimate deletes actually work, and this PR's fail-loudly remains the correct fallback for records that genuinely don't exist.

Happy to send a follow-up PR along those lines if the direction is welcome.

PR #885, opened 2026-08-09
fix(store-pool): pass sendDimensions when creating embedding service

TreyDong:
## Problem

When using a self-hosted / OSS embedding model (e.g. BGE-M3) behind an OpenAI-compatible endpoint that **rejects the `dimensions` parameter** (HTTP 400 "The parameter is invalid"), the **L1 extraction write path** always fails to vectorize:

```
[l1-writer] [vec-dual-write] Embedding FAILED ... Embedding API error: HTTP 400 Bad Request — The parameter is invalid
```

while the **recall path works fine** (query embedding succeeds).

## Root cause

`StorePool.createSqliteStore` constructs the `OpenAIEmbeddingService` **without passing `sendDimensions`**:

https://github.com/TencentCloud/TencentDB-Agent-Memory/blob/main/MemoryCore/src/core/store/store-pool.ts#L377-L384

`OpenAIEmbeddingService` defaults `sendDimensions` to `true` (`config.sendDimensions ?? true`), so every embedding request from this path includes `dimensions: 1024`, which BGE-M3 and similar models reject.

The other creation site (`factory.ts` → `createStoreBundle`, used by the recall/pipeline path) **does** pass `config.embedding.sendDimensions`, which is why only the L1 writer path breaks.

## Fix

Pass `sendDimensions: embCfg.sendDimensions` through to `createEmbeddingService` in `createSqliteStore`, so both paths honor the configured value (users can already set `memory.embedding.sendDimensions: false` in `tdai-gateway.yaml`).

## Verification

Deployed with `memory.embedding.sendDimensions: false` and model `BAAI/bge-m3` (1024 dims):

- Before fix: `vec-dual-write Embedding FAILED ... HTTP 400` on every L1 write
- After fix: `vec-dual-write Embedding OK: dims=1024, norm=1.0000`, and hybrid recall returns `(keyword=2, embedding=1)`

### DIFF
--- a/MemoryCore/src/core/store/store-pool.ts
+++ b/MemoryCore/src/core/store/store-pool.ts
@@ -380,6 +380,7 @@ export class StorePool {
         apiKey: embCfg.apiKey,
         model: embCfg.model,
         dimensions: embCfg.dimensions,
+        sendDimensions: embCfg.sendDimensions,
         maxInputChars: embCfg.maxInputChars,
       }, this.logger as StoreLogger);
     }


Vocllum:
Thanks for the PR — I hit the exact same bug in production and can confirm the root cause chain. A few additional findings that may help get this merged:

1. **Concrete error**: SiliconFlow's `BAAI/bge-m3` endpoint rejects the `dimensions` parameter with `HTTP 400 {"code":20015,"message":"The parameter is invalid. Please check again."}`. So this affects any OpenAI-compatible / OSS embedding backend that doesn't accept `dimensions`, not just one provider.

2. **Existing deployments have empty vector tables (data migration needed)**: because the failure is silent (search falls back to BM25-only, `searchL1Vector` returns 0 candidates), 1.0.1-era deployments never wrote any vectors. In my case `l1_vec`/`l0_vec` had **0 rows** while 497 L1 records and 1311 L0 conversations existed. Fixing the create path alone won't repair old data — a backfill step is needed. I just ran one successfully using the gateway's own `VectorStore` (`upsertL1` / `updateL0Embedding`, idempotent by record_id): 497 + 1311 rows, `dims=1024`, `norm=1.0000`. It may be worth calling this out in the PR description so maintainers/other users know a re-embed of existing data is required.

3. **The batch path is affected too**: `embedBatch` (L0 conversations) hits the same 400. Also, very long messages (20k+ chars) should be truncated (`maxInputChars`) before sending.

Verified after applying the one-line fix: `[hybrid-vec] Embedding OK, dims=1024` and hybrid recall returns strategy=`hybrid` with real vector candidates (previously keyword-only).

Maxwell-Code07:

PR #891, opened 2026-08-09
fix(i18n): ConsoleLayout hardcoded Chinese + scan-chinese scanner improvements

escaleraalcielo:
## Summary

Fixes hardcoded Chinese string in ConsoleLayout that broke the admin group filter when the UI was rendered in English, and improves the Chinese-residual scanner with proper comment stripping and CI integration.

## Changes

- **ConsoleLayout.tsx:111** — replaced hardcoded `'资产管理'` with `t('menu.group.assets')` so the global-admin group filter matches correctly in both Chinese and English locales
- **scan-chinese.mjs** — full rewrite: proper comment stripping (line, block, JSX) with string-literal awareness to avoid false positives on URLs like `http://...`; supports `// i18n-ignore` directive; adds `--strict` mode for CI (exits non-zero on hits)
- **LanguageSwitcher.tsx** — added `// i18n-ignore` comments to native language name strings (`中文`, `English`) so they are excluded from the scanner
- **.gitignore** — added entries to exclude local deployment secrets and configs (`AGENTS.md`, `CREDENTIALS.md`, `deploy/global-images/.admin-key`, `.daniel-key`, `.admin-key`, `.memory-core-config/`, `.proxy-config/`)

## Verification

- Scanner run: `node scripts/scan-chinese.mjs --strict` → 0 hits, exit 0 ✅

### DIFF
--- a/MemoryPanel/web/scripts/scan-chinese.mjs
+++ b/MemoryPanel/web/scripts/scan-chinese.mjs
@@ -9,8 +9,9 @@
  * 规则：
  *   - 扫描 .ts / .tsx 文件
  *   - 排除 src/i18n/ 目录（locale 文件本身就是中文/英文映射表）
- *   - 排除代码注释（// 开头、* 开头、/* 开头的行）
- *   - 检测含 CJK 统一汉字（U+4E00–U+9FFF）的行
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

PR #892, opened 2026-08-09
fix(queue): survive a synchronous task throw in SerialQueue

### LINKED ISSUE #518: Bug: SerialQueue deadlocks after a synchronous task throw

YOMXXX:
Fixes #518.

## Problem

`SerialQueue.drain()` called `entry.task()` directly. If a queued task throws **synchronously**, the call escapes before the `.finally()` handler is registered — the promise returned by `add()` rejects, but the queue stays permanently stuck with `running = true`.

This affects all three pipeline queues (L1/L2/L3, `pipeline-manager.ts`). Subsequent tasks never start, and callers such as `flushSession()` and `_doFlush()` can wait forever on `onIdle()`.

Reproduction (on `main`):
```ts
await queue.add(() => { throw new Error("sync failure"); }).catch(() => undefined);
// hangs forever — queue.running stays true
```

## Change

Invoke the task from a resolved promise (`Promise.resolve().then(() => entry.task())`) so a synchronous throw is routed through the catch/finally chain and queue bookkeeping is finalized.

## Verification

- New `src/utils/serial-queue.test.ts` (4 tests): sync-throw recovery, FIFO order across mixed tasks, `onIdle()` after a sync-throw, and no double-run.
- Confirmed the first test **hangs** on the unfixed code and passes after the fix.
- `npm test` + `npm run build:plugin` pass.

### DIFF
--- a/MemoryCore/src/offload/hooks/after-tool-call.ts
+++ b/MemoryCore/src/offload/hooks/after-tool-call.ts
@@ -34,11 +34,8 @@ import type { PluginConfig, PluginLogger, ToolPair } from "../types.js";
 import type { BackendClient } from "../backend-client.js";
 import {
   buildL3TriggerReport,
-  classifyPatchEffectiveness,
   reportL3Trigger,
   recordToolCall,
-  REPORT_TYPE_L3,
-  L3_FIXED_PATCH_COST_TOKENS,
 } from "../state-reporter.js";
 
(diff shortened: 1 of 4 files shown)


Maxwell-Code07:

Issue #880, opened 2026-08-09
[Bug] L1.5 never settles, MMD never injected, L2/L3 never run in OpenClaw plugin mode


dailylights:
## Summary

In the OpenClaw plugin mode (not standalone Gateway), the `context-offload` module's L1.5 settle, MMD canvas injection, L2 scene extraction, and L3 persona generation are all **silently skipped** or **never executed**. The system has been running for over 48 hours with L0 capture and L1 extraction working, but zero L2/L3 outputs.

### L1.5 never settles naturally (always 60s force-settle)
```
after_tool_call MMD: SKIP (L1.5 not settled yet)
mmd-injector inject: SKIPPED — L1.5 not settled yet (waitForL15=true)
L2 poll: waiting for L1.5 to settle, deferring...
L2 poll: L1.5 settle timeout (60s), force-settling to unblock L2
```

### MMD injection never succeeds
```
mmd-injector inject: injectionReady=false, activeMmdFile=null
```

### L2 scene extraction never runs
```
L2 cold-start skip: not updating l2LastRunTime (minInterval won't block next trigger)
No new L1 records since cursor, skipping scene extraction
```

### L3 checks run but always return below_mild
```
L3(after_tool_call) token snapshot: utilisation=47.4% ✓ OK
state-report OK: stage=after_tool_call reason=below_mild
```

## Root Cause

The `L1.5` settle timer requires a `pre_llm_call` hook or similar to be triggered, but in the OpenClaw plugin mode, this hook may not be invoked by the OpenClaw framework. The `L1.5` flag is set in `after_tool_call` but never cleared by the `pre_llm_call` handler, causing L2 to wait indefinitely.

Additionally, the `assemble SKIP: no stateManager (sessionKey=none)` issue (#878) prevents context compaction, which further degrades the pipeline.

## Impact

- **L1.5**: Never settles naturally (always 60s force-settle)
- **MMD canvas**: Never injected (`injectionReady=false`)
- **L2 scene**: Never extracted (always `No new L1 records`)
- **L3 persona**: Never generated (L2 never produces scenes)
- **Recall**: Always 0 hits (`hits=0` every time — no L1/L2 data to recall)
- **Pipeline**: Only L0 capture and L1 extraction work; L2/L3 are dead code

## Related

- #416 (L2 timer issue in standalone Gateway mode — different root cause)
- #550 (pre_llm_call hook issue in Hermes mode — similar symptom, different platform)
[...]

Issue #862, opened 2026-08-08
[BUG] /compact fails with 'no_session_manager' — Context Engine compact() receives no sessionKey


dailylights:
## Environment
- **OpenClaw**: 2026.7.2-beta.7 (dabe191)
- **Node.js**: v24.18.0
- **Plugin**: @tencentdb-agent-memory/memory-tencentdb@1.0.1 (local mode)
- **OS**: Debian 13 (Linux 6.12.95+deb13-amd64)

## Summary
Running `/compact` slash command fails with `Compaction failed: no_session_manager`.

The `OffloadContextEngine.compact()` method returns `{ ok: false, compacted: false, reason: "no_session_manager" }` because **OpenClaw's framework calls `compact()` without passing a `sessionKey`**.

## Reproduced with Debug Logging

Enabled debug logging and triggered compaction. The complete log chain:

```
2026-08-08T14:55:56.477+08:00 [plugins] [context-offload] >>> CE.compact CALLED: sessionKey=?
2026-08-08T14:55:56.481+08:00 [plugins] [context-offload] >>> compact START: params={sessionId:faaaf089-8cd8-4d4c-9cb8-cfb613f6f735,agentId:main,tokenBudget:1000000,compactionTarget:threshold,force:true,abortSignal:{}}
2026-08-08T14:55:56.485+08:00 [plugins] [context-offload] <<< compact SKIP: no session manager (8ms)
```

**Critical evidence**: The `compact START` log shows the exact `params` received by `compact()`:

```json
{
  "sessionId": "faaaf089-8cd8-4d4c-9cb8-cfb613f6f735",
  "agentId": "main",
  "tokenBudget": 1000000,
  "compactionTarget": "threshold",
  "force": true,
  "abortSignal": {}
}
```

**There is NO `sessionKey` field in the params.** The `sessionKey=?` in the first log confirms `params.sessionKey` is `undefined`.

### Plugin Side — `OffloadContextEngine.compact()` (offload/index.ts:2144)

```typescript
async compact(params: any) {
    let stateManager = params._offloadManager;
    if (!stateManager && params.sessionKey) {   // ← params.sessionKey is undefined → block SKIPPED
      try {
        const entry = await this._sessions.resolveIfAllowed(params.sessionKey, params.sessionId);
        if (entry) stateManager = entry.manager;
      } catch { /* ignore */ }
    }
    ...
    if (!stateManager) {
[...]

PR #893, opened 2026-08-09
fix(offload): resolve session manager in assemble/afterTurn without sessionKey

### LINKED ISSUE #878: [Bug] assemble() always SKIPs: no stateManager (sessionKey=none) — context compaction never runs via afterTurn
## Summary

The `assemble()` method is called on every `afterTurn` but consistently skips with `no stateManager (sessionKey=none)`, because OpenClaw's framework does not pass a `sessionKey` to the `afterTurn` hook.

## Log Evidence

This appears repeatedly on every turn:
```
2026-08-09T10:32:38.835+08:00 [plugins] [context-offload] assemble CALLED: msgs=1067, budget=1000000, prompt=none, sessionKey=?
2026-08-09T10:32:38.839+08:00 [plugins] [context-offload] assemble SKIP: no stateManager (sessionKey=none)
```

Also:
```
2026-08-09T10:32:38.830+08:00 [plugins] [context-offload] >>> CE.afterTurn CALLED: sessionKey=?
```

## Root Cause

`assemble()` is reached without a `sessionKey`, so `stateManager` cannot be resolved. The `assemble()` method has a fallback chain (like the `compact()` fix in #862), but it appears the fallback is not working for the `afterTurn` path, or the session key is genuinely absent from the hook context.

## Impact

- **Context compaction never runs** via the `afterTurn`/`assemble` path
- **Context window grows unbounded** — the L3 token check shows `total=474407 msgCount=1068 utilisation=47.4%`, but compaction is skipped
- **Memory pressure** — the gateway heap grows toward the 608 MiB threshold

## Related

- Issue #862 (same root cause: OpenClaw framework doesn't pass sessionKey) — but #862 only addresses `/compact`, this affects the `afterTurn` path too
- PR #863, #866, #873 (fix the `/compact` path)

## Suggested Fix

- Apply the same sessionKey fallback chain (sessionKey → sessionTarget?.sessionKey → sessionId) to the `afterTurn`/`assemble` path
- Verify the OpenClaw `afterTurn` hook context actually contains a sessionKey or sessionId

YOMXXX:
Fixes #878. Also addresses #880 (same root cause).

## Problem

`OffloadContextEngine.assemble()` and `afterTurn()` only resolved their per-session manager when a top-level `sessionKey` was present. OpenClaw's `afterTurn`/`assemble` hook path passes `sessionId` (or `sessionTarget.sessionKey`) but **not always `sessionKey`**, so every turn logged:

```
assemble CALLED: ... sessionKey=?
assemble SKIP: no stateManager (sessionKey=none)
```

Since `assemble()` is the **sole L1.5 trigger point**, skipping it cascaded:
- L1.5 never settled → MMD injection never ran (`waiting for L1.5 to settle`)
- L2 scene extraction deferred forever
- L3 compression skipped (`below_mild` with an unbounded context)
- Context window grew unchecked toward the heap threshold

This is the same root cause as #862, but #862 only fixed the `/compact` path (`compact()`); the `afterTurn`/`assemble` path was left broken.

## Change

Extend the stateManager fallback in both `assemble()` and `afterTurn()` to:
`sessionKey → sessionId → sessionTarget?.sessionKey`

mirroring the `compact()` resolution path (#862).

## Verification

New `src/offload/session-resolve.test.ts` (3 tests):
- `assemble` resolves manager from `sessionId` alone and reaches the L1.5 trigger
- `afterTurn` resolves manager from `sessionId` alone without throwing
- `assemble` skips cleanly when no session is resolvable

The first test **fails** on the unfixed code (judgeL15 never fires) and passes after the fix.

`npm test` + `npm run build:plugin` pass.

### DIFF
--- a/MemoryCore/src/offload/hooks/after-tool-call.ts
+++ b/MemoryCore/src/offload/hooks/after-tool-call.ts
@@ -34,11 +34,8 @@ import type { PluginConfig, PluginLogger, ToolPair } from "../types.js";
 import type { BackendClient } from "../backend-client.js";
 import {
   buildL3TriggerReport,
-  classifyPatchEffectiveness,
   reportL3Trigger,
   recordToolCall,
-  REPORT_TYPE_L3,
-  L3_FIXED_PATCH_COST_TOKENS,
 } from "../state-reporter.js";
 
 function isHeartbeatToolCall(event: any, cachedParams: any): boolean {
@@ -97,6 +94,9 @@ export function createAfterToolCallHandler(
   getContextWindow: (() => number) | undefined,
   pluginConfig: Partial<PluginConfig> | undefined,
   backendClient?: BackendClient | null,
+  /** Official OpenClaw API to fetch session messages (issue #851) — used when
+   *  the hook event does not carry messages (no dist-file patch). */
+  getSessionMessages?: (opts: { sessionKey: string; limit?: number }) => Promise<unknown[]>,
 ) {
   return async (event: any, ctx: any) => {
     // Skip internal memory-pipeline sessions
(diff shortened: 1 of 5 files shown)


Maxwell-Code07:

PR #895, opened 2026-08-09
fix: quote admin key path next to full-width punctuation

mangca0:
## Summary

- Quote `ADMIN_KEY_FILE` in the admin initialization log message.
- Prevent macOS Bash 3.2 from treating adjacent full-width punctuation as part of the variable name.

## Validation

- Passed `bash -n` for all deployment shell scripts.
- Reproduced and verified the fix with macOS Bash 3.2.


Maxwell-Code07:

PR #896, opened 2026-08-09
wayneos/local-fix-patches: 4 fixes verified end-to-end on Aug 7+9 2026

Matrix-ops77:
## Summary

Four patches that have been verified end-to-end against the live stack (L0 conversation capture, L1 atom extraction, L2 scene writing, vector retrieval via 1024-dim Ollama snowflake-arctic-embed2, ~4869-token memory injection vs 79-token baseline = 62x injection).

These are the exact patches baked into the `agentmemory/memory-core:local-fix`, `agentmemory/memory-proxy:local-fix`, and `team-memory-panel-knowledge:amd64` images we use in production.

## Changes

| File | Lines | What |
|---|---|---|
| `MemoryCore/src/adapters/standalone/llm-runner.ts` | +243 / -85 | (1) AI SDK bypass for pure-text L1/L2 calls — plain `fetch` with `stream: false` against OpenAI-compatible endpoint, because the Vercel AI SDK's SSE parser fails silently on certain backends (`createOpenAI({compatibility: compatible})` always sends `stream: true`). Tool-using paths still go through the SDK. (2) `directFetchInternal` typo fix (was `directFetchWithMessages`) that crashed the L2 scene-file writer's tool-call loop. |
| `MemoryProxy/src/handler.ts` | +24 / -1 | chat-completions wrapper now forwards `x-team-id`, `x-agent-id`, `x-task-id`, `x-conversation-id` extra_headers to upstream so routing metadata survives the proxy hop. |
| `deploy/global-images/_lib.sh` | +35 | `wait_healthy()` extended with service-id header awareness for the new Core v3 endpoints (v2 router needs Bearer + `x-tdai-service-id`; v3 adds `x-tdai-user-key`). |
| `deploy/global-images/start-memory-core.sh` | +31 / -9 | Port assignments adjusted for the new combined panel+knowledge image (`team-memory-panel-knowledge:amd64` serves both on one container, ~600MB saved vs the split deployment). |

## Why a PR vs just patches

These fixes are required to run against real OpenAI-compatible backends (Ollama, OmniRoute, vLLM, LiteLLM, etc.). The default upstream code path returns 0 L0/L1/L2 records on those backends because of the SDK SSE issue and the missing header forwarding.

## Verification

| Component | Before | After |
|---|---|---|
| L1 pipeline tasks consumed | 0 | 8 |
| L1 pipeline tasks completed | 0 | 8 (failed: 0) |
| L2 scene files written | 0 | 6 .md |
| L0 conversations captured | 0 | 58 |
| Memory injection tokens | 79 (baseline) | 4,869 (62x) |
| End-to-end latency | n/a | 2.18s |

Environment: macOS M4, 16GB RAM, OrbStack (Docker alternative), Ollama with snowflake-arctic-embed2 (1.2GB, 1024-dim), OmniRoute MiniMax-M3 upstream, MacBook + 228GB drive.

## Notes

- The 243-line AI SDK bypass in `llm-runner.ts` is the load-bearing fix. Everything else follows once that path works.
- We considered breaking this into 4 separate PRs, but they're tightly coupled (the handler.ts change is meaningless without the llm-runner.ts SDK bypass; the deploy script change requires the lib.sh change). Happy to split if maintainers prefer.

Happy to iterate. `Matrix-ops77`

### DIFF
--- a/MemoryCore/src/adapters/standalone/llm-runner.ts
+++ b/MemoryCore/src/adapters/standalone/llm-runner.ts
@@ -280,110 +280,227 @@ export class StandaloneLLMRunner implements LLMRunner {
       `tools=${effectiveEnableTools}${callerProvidedTools ? "(caller)" : ""}, timeout=${timeoutMs}ms`,
     );
 
-    // Create OpenAI-compatible provider via AI SDK
-    // Use "compatible" mode to call /chat/completions (not Responses API),
-    // which works with all OpenAI-compatible backends (DeepSeek, Qwen, etc.)
-    const provider = createOpenAI({
-      baseURL: this.config.baseUrl,
-      apiKey: this.config.apiKey,
-      compatibility: "compatible",
-    });
+    // ── Direct-fetch fallback for pure-text tasks (no tools) ──
+    //
+    // The Vercel AI SDK's createOpenAI({ compatibility: "compatible" }) path
+    // requests `stream: true` regardless of the caller's intent, then parses
+    // the SSE response. Some OpenAI-compatible backends (notably OmniRoute
+    // 3.8.48) emit trailing SSE comments and an empty `usage` chunk that the
+    // SDK's stream parser rejects with "Invalid JSON response" — even though
+    // the model itself returned valid JSON.
+    //
+    // For pure-text tasks (L1 extraction, summarization) we bypass the SDK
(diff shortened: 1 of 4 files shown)


Maxwell-Code07:

PR #898, opened 2026-08-09
fix(hermes-plugin): always advertise tool schemas so they are routable before initialize()

francialisomlimoeiro:
## What it does

`MemoryManager.add_provider()` indexes `get_tool_schemas()` into the tool-routing table BEFORE `initialize()`. The old gateway-availability gate returned `[]` when the Gateway was not yet reachable and no Gateway env var was set, so the routing table stayed empty forever and every tool call failed with `Unknown tool: memory_tencentdb_*` — even after `initialize()` succeeded.

Changes:
- `get_tool_schemas()` always returns both static schemas; Gateway health is enforced at call time (`handle_tool_call` / `_ensure_alive_for_request`).
- `src/core/hooks/auto-recall.ts`: MEMORY_TOOLS_GUIDE + truncation suffix now reference the tool names the plugin actually registers (removes legacy `tdai_*` names from advertised guidance).
- README troubleshooting entry updated.
- New test `tests/test_tool_schema_routing.py` (4 tests).

## Tests

```
PYTHONPATH=<hermes-agent checkout> python3 -m pytest tests/test_tool_schema_routing.py
# 4 passed
```

(schemas advertised while uninitialized and without Gateway env; advertised names dispatchable via stub client; unknown names rejected)

### DIFF
--- a/hermes-plugin/memory/memory_tencentdb/README.md
+++ b/hermes-plugin/memory/memory_tencentdb/README.md
@@ -287,11 +287,13 @@ aliases:
   preference list (in-tree first, then `$HOME`). If you want to pin a
   specific path, set `MEMORY_TENCENTDB_GATEWAY_CMD` explicitly — it always
   wins over discovery.
-- **Search tools silently missing from the LLM**: `get_tool_schemas()`
-  returns `[]` until either the Gateway is reachable or one of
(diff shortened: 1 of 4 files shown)


Maxwell-Code07:

PR #434, opened 2026-07-08
fix(hermes): require explicit memory indexing

YOMXXX:
## Summary
- build on #418 and require explicit Hermes memory writes to reach the searchable L1 index before reporting `stored: true`
- keep normal L1 extraction writes fault-tolerant by making the strict VectorStore requirement opt-in
- treat Gateway `stored:false` as a Hermes provider sync failure so it is logged and routed through the existing recovery path

## Verification
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/core/record/explicit-memory.test.ts`
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm test`
- `COREPACK_ENABLE_AUTO_PIN=0 pnpm build`
- `python3 -m py_compile hermes-plugin/memory/memory_tencentdb/__init__.py hermes-plugin/memory/memory_tencentdb/client.py hermes-plugin/memory/memory_tencentdb/tests/test_memory_tencentdb_recovery.py`

## Notes
- `python3 -m pytest hermes-plugin/memory/memory_tencentdb/tests/test_memory_tencentdb_recovery.py -q` could not run locally because this Python environment does not have `pytest` installed.

Fixes #417
Supersedes #418

PR #904, opened 2026-08-10
feat(hermes): mirror built-in memory writes into TencentDB L1

### LINKED ISSUE #417: [Bug] Hermes memory_tencentdb does not mirror explicit memory() writes into TencentDB memory search

YOMXXX:
Fixes #417.

## Problem

Hermes' `memory_tencentdb` provider stubbed `on_memory_write()`:

```python
def on_memory_write(self, action, target, content):
    pass
```

So explicit `memory(action="add", ...)` writes succeeded locally but never became retrievable through TencentDB **memory search** — only conversation search surfaced them. Built-in durable memory diverged from L1 memory search.

## Change

Implement the full mirror path end-to-end:

- **`__init__.py`** — `on_memory_write()` now mirrors non-add/empty/gateway-down cases as no-ops and otherwise calls `client.write_explicit_memory()`; on failure it records a metric + tries to recover the gateway.
- **`client.py`** — `write_explicit_memory()` posts to `/v3/memories/explicit` (v3 data plane, tenancy-isolated with team/agent/user).
- **`v2-router.ts`** — new `/memories/explicit` route (mounted on both `/v2/` and `/v3/`, added to `V3_ALLOWED_SUBPATHS`); `handleExplicitMemoryWrite` enforces isolation and ingests via `explicit-memory.ts`.
- **`explicit-memory.ts`** (new) — classifies the target (`user`/`profile` → `persona`, else `instruction`) and writes directly into L1 via `writeMemory()`.
- **`l1-writer.ts`** — gains `requireVectorStoreWrite`: the write must reach the **searchable** VectorStore index (embedding + upsert), otherwise it is rejected. A JSONL-only write would never surface in recall, so this guarantees `stored=true` ⇔ retrievable.

## Relationship to #434

This is the v2 (monorepo, `feat/server_team`) re-implementation of my earlier #434, which targeted the pre-refactor `main` tree. It adds the same stricter guarantee (#434's requirement: explicit writes reach the searchable L1 index before reporting stored) but on the v2 data plane with tenancy isolation.

## Verification

- `explicit-memory.test.ts` (5 tests): store success + classification, persona classification, non-add/empty rejection, no-vector-store rejection, upsert-failure rejection.
- `explicit-memory-route.test.ts` (4 tests): handler stores, 503 without store, 400 on missing fields, `/v3/memories/explicit` dispatch.
- `python3 -m py_compile` on the two Python files, `npm test` + `npm run build:plugin` pass.

### DIFF
--- a/MemoryCore/hermes-plugin/memory/memory_tencentdb/__init__.py
+++ b/MemoryCore/hermes-plugin/memory/memory_tencentdb/__init__.py
@@ -910,7 +910,29 @@ class MemoryTencentdbProvider(MemoryProvider):
 
     def on_memory_write(self, action: str, target: str, content: str) -> None:
         """Mirror built-in memory writes to memory-tencentdb for indexing."""
-        pass
+        if action.strip().lower() != "add":
+            return
+        if not content.strip():
+            return
+        if not self._ensure_alive_for_request() or not self._client:
+            return
+
+        try:
+            result = self._client.write_explicit_memory(
+                content=content,
+                target=target,
+                action=action,
+                session_id=self._session_id,
+                user_id=self._user_id,
+            )
+            data = result.get("data") or {}
+            if data.get("stored") is False:
+                raise RuntimeError("Gateway did not index explicit memory")
+            self._record_success()
+        except Exception as e:
+            self._record_failure()
+            logger.warning("memory-tencentdb explicit memory sync failed: %s", e)
+            self._try_recover_gateway()
 
     def on_session_end(self, messages: List[Dict[str, Any]]) -> None:
         """Trigger session-level flush on the Gateway.
--- a/MemoryCore/hermes-plugin/memory/memory_tencentdb/client.py
+++ b/MemoryCore/hermes-plugin/memory/memory_tencentdb/client.py
@@ -154,6 +154,36 @@ class MemoryTencentdbSdkClient:
             body["session_id"] = session_id
         return self._post("/v3/conversation/search", body)
 
+    def write_explicit_memory(
+        self,
+        content: str,
+        *,
+        target: str = "memory",
+        action: str = "add",
+        session_id: str = "",
+        team_id: str = "default",
+        agent_id: str = "default",
+        user_id: str = "default",
+    ) -> Dict[str, Any]:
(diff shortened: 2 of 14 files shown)


Maxwell-Code07:

PR #909, opened 2026-08-10
fix(panel): parse legacy chat_memory asset id without agt prefix

Vocllum:
## Problem

The Chat_Memory layer API (`POST /api/v1/chat-memory/layer`) returns empty results for chat_memory assets created by `MemoryCore`'s `buildChatMemoryAssetId`.

`MemoryCore/src/metadata/utils/chat-memory-asset.ts` generates:

```
chat_memory-{team_id}-{agent_id}
```

but `parseChatMemoryAssetId` in `MemoryPanel/src/panel/http/routes/chat-memory.ts` only accepts the newer format with an `agt` marker:

```
chat_memory-{team_id}-agt{agent_id}
```

When the marker is absent, `parseChatMemoryAssetId` returns `null` and the route falls into the "user-created UserAsset" branch, returning an empty list — even though the asset exists and the data is there. This affects every asset created by v2.0.0 and earlier, where agent ids have no `agt` prefix (e.g. `agent_id = 'default'`).

## Root cause evidence

- Generator: `MemoryCore/src/metadata/utils/chat-memory-asset.ts:17` — `` return `chat_memory-${teamId}-${agentId}`; `` (its own comment states the id is not expected to be reverse-parsed, yet the panel does parse it).
- Parser: `MemoryPanel/src/panel/http/routes/chat-memory.ts:1070` — requires `lastIndexOf('-agt') >= 0`, otherwise returns `null`.
- Consequence: `chat_memory-default-default` (produced by the default generator) → `parseChatMemoryAssetId` returns `null` → `respondEnvelope(c, okEnvelope(c, { layer, items: [], total: 0, ... }))`.

## Fix

Fall back to splitting on the last `-` when the `-agt` marker is absent. The new-format branch is unchanged, so ids with `agt`-prefixed agents keep parsing exactly as before.

## Verification

- Behavior cases: 8/8 pass (new format with `agt`, legacy format, team ids containing `-`, malformed ids).
- `tsc --noEmit` in `MemoryPanel/`: 0 errors.

## Duplicate check

Searched issues/PRs for `parseChatMemoryAssetId`, `chat_memory`, `chat-memory asset` — no existing report of this mismatch.


Maxwell-Code07:

azprin:
Hit this exact issue on prod after v1.0.1 → v2.0.0 migration. 

Symptom: L0 counter in Panel UI shows 287 (old) and doesn't grow despite
new conversations being written. Cause matches the bug report exactly:
plugin writes `chat_memory-{team_id}-{agent_id}` for non-`agt`-prefixed
agent_id="default", but Panel parser requires `-agt` marker.

User: usr-qaaf1oxnu2 / team-qem9cpvr7y / agent_id='default' (literal name from
migrated v1 plugin). Asset key generated: `chat_memory-team-qem9cpvr7y-default`.

Posting because the dup check says no previous reports — but there should be
more users hitting this after migration. Please prioritize the fix.

Vocllum:
Thanks for the detailed report — it's very useful to have a production reproduction.

**On this PR (#909)**: the parser fix is still the right scope for the legacy-asset *visibility* problem you hit (the L0 counter showing 287 and not growing). I've also rebased the branch onto the latest `feat/server_team` (it had drifted into a conflict) and added a guard so a trailing-dash legacy id (`chat_memory-team-`) returns `null` instead of an empty `agentId`. `tsc --noEmit` passes and the parser behavior cases pass (new `-agt` format, legacy format, team ids containing `-`, malformed/empty ids).

**On the orphan-write problem you filed in #995**: you're right that this PR alone doesn't fix future writes. The plugin hardcoding `agent_id="default"` is a separate layer (Hermes-side `memory_tencentdb` plugin), and the fix belongs there — lazy-resolving the real agent id via the API on session start, as you sketched. That should land as its own change tracked in #995 rather than being folded into the panel parser PR. I've confirmed the hardcode exists (`_DEFAULT_AGENT_ID = "default"` in the bundled plugin), so #995 is a real issue, not a duplicate.

cc @Maxwell-Code07 — could you also take a look at #995? It affects every v1→v2 migration user.

PR #918, opened 2026-08-10
fix(memory-core): apply recall score threshold to vector search

### LINKED ISSUE #902: [Bug] conversation_search / memory_search never apply recall.scoreThreshold — the vector branch returns top-k regardless of similarity
## Summary

`recall.scoreThreshold` is honored on the auto-recall path but not by the two search tools. In `executeMemorySearch` and `executeConversationSearch` the vector branch returns its top-k candidates unconditionally, so a query about a fact that is not in the corpus comes back with `limit` irrelevant results instead of none.

The same guard already exists a few files over, in `auto-recall.ts::searchByEmbedding`. The proposed fix is to extend it to the tool path, reusing the existing config key.

Verified on `@tencentdb-agent-memory/memory-tencentdb@1.0.1` (npm) and still present on `main` (`MemoryCore`, `2.0.0-beta.1`).

## Environment

| | |
|---|---|
| Package | `@tencentdb-agent-memory/memory-tencentdb@1.0.1` (npm), standalone gateway mode |
| Store | SQLite (`sqlite-vec@0.1.7-alpha.2`), FTS5 available |
| Embeddings | OpenAI `text-embedding-3-large`, `dimensions=1536` |
| Config | `recall.strategy: hybrid`, `recall.scoreThreshold: 0.3` (shipped defaults, `MemoryCore/tdai-gateway.yaml:68-73`) |
| Host | macOS, Node 22 |

## Symptom

`POST /search/conversations` (and the `tdai_conversation_search` / `tdai_memory_search` tools, and the `/v2` search endpoints) with a query whose subject does not exist in the corpus returns the full `limit` of results. There is no score floor: the caller cannot distinguish "here are 5 relevant memories" from "here are the 5 least-distant rows in the table".

The auto-recall path with `strategy: embedding` on the same store and the same query returns nothing, which is the expected behavior.

## Root cause

Line numbers are from the npm 1.0.1 tarball (`src/…`); the `main` equivalents (`MemoryCore/src/…`) follow in brackets.

**1. The guard exists on the recall path.**

`src/core/hooks/auto-recall.ts:439` [main: `:459`] reads the key:

```ts
const threshold = cfg.recall.scoreThreshold ?? 0.3;
```

and `searchByEmbedding` filters with it before returning — `src/core/hooks/auto-recall.ts:591-593` [main: `:612`]:

```ts
const filtered = vecResults
  .filter((r) => r.score >= threshold)
  .slice(0, maxResults);
```

`searchByKeyword` does the same for the FTS branch at `src/core/hooks/auto-recall.ts:524-526` [main: `:545`].

**2. The tools have no threshold at all.**

`executeMemorySearch` (`src/core/tools/memory-search.ts:82-99` [main: `:87`]) and `executeConversationSearch` (`src/core/tools/conversation-search.ts:83-98` [main: `:86`]) neither accept nor read a threshold. Their vector branch returns everything the store gives back:

`src/core/tools/memory-search.ts:211-212` [main: `:233-235`]:

```ts
const vecResults: L1SearchResult[] = await vectorStore.searchL1Vector(queryEmbedding, candidateK, query);
logger?.debug?.(`${TAG} [hybrid-vec] Vector search returned ${vecResults.length} candidates`);
```

`src/core/tools/conversation-search.ts:204-205` [main: `:217-219`] is the same with `searchL0Vector` / `L0SearchResult`.

Those candidates go straight into the RRF merge (`memory-search.ts:251`, `conversation-search.ts:242`) and then `slice(0, limit)` (`conversation-search.ts:259`). With `candidateK = limit * 3` and no floor, `limit` results are essentially always produced whenever the corpus has ≥ `limit` rows.

**3. No caller supplies one either.**

- `src/core/tdai-core.ts:316-324` and `:338-345` — `TdaiCore.searchMemories` / `searchConversations` pass `vectorStore`, `embeddingService`, `logger` and nothing else, although `this.cfg.recall.scoreThreshold` is right there. This path serves both the MCP tools and the v1 HTTP routes (`src/gateway/server.ts:637-640` → `handleSearchConversations` at `:974-993`).
- `src/gateway/v2-router.ts:437-444` and `:615-620` — the `/v2` handlers, same omission.

**4. Related, same shape:** `searchHybrid` in `src/core/hooks/auto-recall.ts:618-622` [main: `:638-642`] takes the threshold as `_threshold` and never uses it. Since `recall.strategy` defaults to `"hybrid"` (`src/config.ts:539`), a default install applies `scoreThreshold` on no path at all — the `scoreThreshold: 0.3` in the shipped `tdai-gateway.yaml` is inert unless the strategy is explicitly set to `embedding` or `keyword`. Worth fixing with the same one-liner, though the tool path is the one this report is about.

## Reproduction

1. Standalone gateway, SQLite store, any embedding provider configured, FTS5 available.
2. Capture a handful of unrelated facts (we used 13).
3. Query for something that is definitely not among them:

```bash
curl -s localhost:8420/search/conversations \
  -H 'content-type: application/json' \
  -d '{"query":"what is the company parental leave policy","limit":5}'
```

Observed: `total: 5`, five unrelated messages.
Expected: `total: 0` (or only rows above the configured similarity floor), matching what `auto-recall` returns for the same query under `strategy: embedding`.

Debug logs show the vector branch reporting its candidates with cosine scores well below `recall.scoreThreshold`, and nothing dropping them.

## Measurement

Synthetic corpus, mean of 3 runs per threshold, `limit=5`:

- 13 facts, each queried with a paraphrase that shares no distinctive term with the fact (one fact is in Chinese, queried in English).
- **hit rate** = fraction of the 13 paraphrase queries whose correct fact appears in the returned set.
- **false positives** = mean number of results returned for 3 queries about facts absent from the corpus. 0 is correct, 5 is the worst case (`limit`).
- **lexical control** = the same 13 facts queried with their own keywords, to confirm the keyword path does not regress.

| threshold (vector branch) | hit rate | MRR | false positives | lexical control |
|---|---|---|---|---|
| none (current behavior) | 97.4% | 0.949 | **5.00** | 12.7/13 |
| 0.20 | **100.0%** | **0.962** | 2.33 | 13/13 |
| 0.30 | 92.3% | 0.885 | 1.33 | 13/13 |
| 0.35 | 92.3% | 0.885 | **0.33** | 13/13 |
| 0.40 | 76.9% | 0.731 | 0.00 | 13/13 |

Two observations:

- The current behavior floors false positives at exactly `limit`. That is structural, not a model quality issue — no embedding model fixes it.
- At 0.20 the hit rate goes **up** relative to no filtering (100.0% vs 97.4%). Dropping far candidates before the RRF merge frees rank slots for the correct document, so the filter is not purely a precision/recall trade in a hybrid setting.

Caveats, stated plainly: small synthetic corpus, single embedding model, one language pair. The numbers argue that *a* threshold helps and that the current absence of one is a real defect; they are not a recommendation for a specific default. We deployed 0.35 locally because our workload is more sensitive to noise than to occasional forgetting; a different workload would reasonably pick 0.20.

## Proposed fix

Minimal, and symmetric with `auto-recall.ts:591-593`. Add an optional threshold to the tool signature and filter the vector branch before the RRF merge.

`src/core/tools/memory-search.ts`:

```diff
 export async function executeMemorySearch(params: {
   query: string;
   limit: number;
   type?: string;
   scene?: string;
   vectorStore?: IMemoryStore;
   embeddingService?: EmbeddingService;
   logger?: Logger;
+  /** Minimum cosine similarity for the vector branch (recall.scoreThreshold). 0 disables. */
+  scoreThreshold?: number;
 }): Promise<MemorySearchResult> {
   const {
     query,
     limit,
     type: typeFilter,
     scene: sceneFilter,
     vectorStore,
     embeddingService,
     logger,
+    scoreThreshold = 0,
   } = params;
```

```diff
-        const vecResults: L1SearchResult[] = await vectorStore.searchL1Vector(queryEmbedding, candidateK, query);
-        logger?.debug?.(`${TAG} [hybrid-vec] Vector search returned ${vecResults.length} candidates`);
-        return vecResults.map((r) => ({
+        const vecRaw: L1SearchResult[] = await vectorStore.searchL1Vector(queryEmbedding, candidateK, query);
+        // Same guard as auto-recall.ts::searchByEmbedding: drop candidates below the
+        // configured cosine floor before they reach the RRF merge.
+        const vecResults = scoreThreshold > 0
+          ? vecRaw.filter((r) => r.score >= scoreThreshold)
+          : vecRaw;
+        logger?.debug?.(
+          `${TAG} [hybrid-vec] Vector search returned ${vecResults.length}/${vecRaw.length} candidates (threshold=${scoreThreshold})`,
+        );
+        return vecResults.map((r) => ({
```

`src/core/tools/conversation-search.ts` is identical modulo `searchL0Vector` / `L0SearchResult`.

Wiring:

- `TdaiCore.searchMemories` / `searchConversations` (`src/core/tdai-core.ts:316`, `:338`) pass `scoreThreshold: this.cfg.recall.scoreThreshold`. This covers the MCP tools and the v1 HTTP routes in one edit.
- The `/v2` handlers (`src/gateway/v2-router.ts:437`, `:615`) need the value threaded in: `V2RouterDeps` (`v2-router.ts:78-96`) currently exposes no config accessor. `v2Deps` is constructed with `this.core` in scope (`src/gateway/server.ts:524-535`), so a getter on `TdaiCore` alongside the existing `getVectorStore()` / `getEmbeddingService()` would fit the current style.

### Two deliberate constraints on the patch

**Vector branch only.** The vector score is cosine similarity — `score = 1.0 - distance` (`src/core/store/sqlite.ts:1167` for L1, `:1614` for L0) over a `distance_metric=cosine` vec0 table (`:587`, `:668`). The FTS branch score is BM25 squashed to 0–1 by `bm25RankToScore` (`src/core/store/sqlite.ts:274-281`), whose value depends on corpus statistics and is not on the same scale. Applying one number to both would silently mean two different things. (`auto-recall.ts:524-526` does apply the same key to the FTS branch; that is a separate question and we did not touch it.)

**No new config key.** `recall.scoreThreshold` already exists and is already documented as the recall score floor (`src/config.ts:87-88`, default 0.3 at `:538`, shipped at `MemoryCore/tdai-gateway.yaml:71`).

**Default behavior.** The diff above defaults to `0` (disabled) so the change is inert until a caller passes a value. If you prefer the behavior to be symmetric with auto-recall out of the box, wire `cfg.recall.scoreThreshold` directly and the existing default of 0.3 applies — that changes results for current users, so it is your call, not ours.

Happy to open a PR if the approach looks right.

aoright:
## Description | 描述

- Apply the existing `recall.scoreThreshold` to L0 and L1 vector candidates before client-side RRF.
- Thread the configured threshold through `TdaiCore` and the v2 router while preserving a disabled-by-default value for direct tool callers.
- Honor the threshold in the default hybrid auto-recall path.
- Keep FTS and native-hybrid scores untouched because they are not raw cosine similarities.
- Add regression coverage for L0, L1, backward compatibility, and hybrid auto-recall.

## Related Issue | 关联 Issue

Fixes #902

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

Validation:
- `npm test` — 1 file, 4 tests passed
- `npm pack --dry-run --ignore-scripts` — passed
- `git diff --check` — passed

## Additional Notes | 其他说明

The repository-wide `npm run build` reaches and passes `build:plugin`, but its existing `build:seed-v2` step fails because `scripts/seed-v2/tsconfig.json` is absent from the current base branch.

### DIFF
--- a/MemoryCore/src/core/hooks/auto-recall.ts
+++ b/MemoryCore/src/core/hooks/auto-recall.ts
@@ -639,7 +639,7 @@ async function searchHybrid(
   userText: string,
   _pluginDataDir: string,
   maxResults: number,
-  _threshold: number,
+  threshold: number,
   vectorStore: IMemoryStore,
   embeddingService: EmbeddingService,
   logger?: Logger,
@@ -710,7 +710,17 @@ async function searchHybrid(
   ]);
 
   const keywordResults = keywordResult.records;
-  const embeddingResults = embeddingResult.results;
+  // Only vector scores are cosine similarities. Apply the configured floor
+  // before RRF while leaving BM25-derived keyword scores untouched.
+  const embeddingResults = threshold > 0
+    ? embeddingResult.results.filter((r) => r.score >= threshold)
+    : embeddingResult.results;
+  if (embeddingResults.length !== embeddingResult.results.length) {
+    logger?.debug?.(
+      `${TAG} [hybrid-embedding] Kept ${embeddingResults.length}/${embeddingResult.results.length} ` +
+      `candidates above threshold=${threshold}`,
+    );
+  }
   const timing: SearchTiming = {
     ftsMs: keywordResult.ms,
     embeddingMs: embeddingResult.ms,
--- a/MemoryCore/src/core/tdai-core.ts
+++ b/MemoryCore/src/core/tdai-core.ts
@@ -443,6 +443,7 @@ export class TdaiCore {
       scene: params.scene,
       vectorStore: this.vectorStore,
       embeddingService: this.embeddingService,
+      scoreThreshold: this.cfg.recall.scoreThreshold,
       logger: this.logger,
     });
 
@@ -464,6 +465,7 @@ export class TdaiCore {
       sessionKey: params.sessionKey,
       vectorStore: this.vectorStore,
       embeddingService: this.embeddingService,
+      scoreThreshold: this.cfg.recall.scoreThreshold,
       logger: this.logger,
     });
 
@@ -528,6 +530,11 @@ export class TdaiCore {
   /** Get the shared EmbeddingService (may be undefined if not configured). */
   getEmbeddingService(): EmbeddingService | undefined {
     return this.embeddingService;
+  }
+
+  /** Get the configured minimum cosine similarity for recall candidates. */
+  getRecallScoreThreshold(): number {
+    return this.cfg.recall.scoreThreshold;
   }
 
   /** Get the pipeline scheduler (may be undefined if extraction disabled). */
--- a/MemoryCore/src/core/tools/conversation-search.ts
+++ b/MemoryCore/src/core/tools/conversation-search.ts
@@ -90,6 +90,8 @@ export async function executeConversationSearch(params: {
   filter?: IsolationFilter;
   vectorStore?: IMemoryStore;
   embeddingService?: EmbeddingService;
+  /** Minimum cosine similarity for vector candidates. A value of 0 disables filtering. */
+  scoreThreshold?: number;
   logger?: Logger;
 }): Promise<ConversationSearchResult> {
   const {
@@ -99,6 +101,7 @@ export async function executeConversationSearch(params: {
     filter: isolationFilter,
     vectorStore,
     embeddingService,
+    scoreThreshold = 0,
     logger,
   } = params;
 
@@ -213,10 +216,18 @@ export async function executeConversationSearch(params: {
         logger?.debug?.(
           `${TAG} [hybrid-vec] Embedding OK, dims=${queryEmbedding.length}, searching top-${candidateK}...`,
         );
-        const vecResults: L0SearchResult[] = isolationFilter
+        const vecRaw: L0SearchResult[] = isolationFilter
           ? await vectorStore.searchL0Vector(queryEmbedding, candidateK, query, isolationFilter)
           : await vectorStore.searchL0Vector(queryEmbedding, candidateK, query);
-        logger?.debug?.(`${TAG} [hybrid-vec] Vector search returned ${vecResults.length} candidates`);
+        // Vector scores are cosine similarities; filter before RRF so weak
+        // candidates cannot gain relevance merely by occupying a rank slot.
+        const vecResults = scoreThreshold > 0
+          ? vecRaw.filter((r) => r.score >= scoreThreshold)
+          : vecRaw;
+        logger?.debug?.(
+          `${TAG} [hybrid-vec] Vector search returned ${vecResults.length}/${vecRaw.length} candidates ` +
+          `(threshold=${scoreThreshold})`,
+        );
         return vecResults.map((r) => ({
           id: r.record_id,
           session_key: r.session_key,
--- a/MemoryCore/src/core/tools/memory-search.ts
+++ b/MemoryCore/src/core/tools/memory-search.ts
@@ -92,6 +92,8 @@ export async function executeMemorySearch(params: {
   filter?: IsolationFilter;
   vectorStore?: IMemoryStore;
   embeddingService?: EmbeddingService;
+  /** Minimum cosine similarity for vector candidates. A value of 0 disables filtering. */
+  scoreThreshold?: number;
   logger?: Logger;
 }): Promise<MemorySearchResult> {
   const {
@@ -102,6 +104,7 @@ export async function executeMemorySearch(params: {
     filter: isolationFilter,
     vectorStore,
     embeddingService,
+    scoreThreshold = 0,
     logger,
   } = params;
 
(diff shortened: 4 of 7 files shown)


Maxwell-Code07:

YuhaoLin2005:
**Review: scoreThreshold is still dead on the native-hybrid short-circuit path**

The per-branch vector filtering before RRF is exactly the right call — thanks for threading `scoreThreshold` through both tools and wiring `getRecallScoreThreshold` in the router.

One gap remains: both search tools short-circuit **before** reaching the filtered vector branch when the store supports native hybrid search:

- `conversation-search.ts:146-173` — `if (vectorStore.getCapabilities().nativeHybridSearch && vectorStore.searchL0Hybrid)` → maps results straight to `items.slice(0, limit)` with no threshold applied
- `memory-search.ts:149-…` — same path for `searchL1Hybrid`

So on a store where `nativeHybridSearch` is true (e.g. the TCVDB adapter), results never pass through the `scoreThreshold` filter and #902 is fixed for the SQLite dual-path but not for native-hybrid stores.

One caveat to consider: the fused score returned by a native-hybrid call may not be on the cosine-similarity scale, so a blanket `r.score >= scoreThreshold` (default 0.3) could over-filter there. If the store's fused score is comparable, the fix is to apply the same filter before `slice(0, limit)` in both short-circuit paths.

Happy to open a small follow-up PR for the two short-circuit paths if that direction is welcome.

Issue #879, opened 2026-08-09
[Bug] L1 extraction fails with TypeError: ByteString conversion rejects Unicode ellipsis (U+2026)


dailylights:
## Summary

L1 extraction fails with a `TypeError` when the conversation contains the Unicode ellipsis character `…` (U+2026, decimal 8230). The error is `Cannot convert argument to a ByteString because the character at index 13 has a value of 8230 which is greater than 255`.

## Log Evidence

From `observability.log` (2026-07-31):
```
[2026-07-31T23:55:12.274Z][ERROR] core.offload-l1.failed {"error":"Cannot convert argument to a ByteString because the character at index 13 has a value of 8230 which is greater than 255.","error.type":"TypeError","task_type":"offload-l1"}
[2026-07-31T23:55:17.333Z][ERROR] core.offload-l1.failed {"error":"Cannot convert argument to a ByteString because the character at index 13 has a value of 8230 which is greater than 255.","error.type":"TypeError","task_type":"offload-l1"}
[2026-07-31T23:55:32.342Z][ERROR] core.offload-l1.failed {"error":"Cannot convert argument to a ByteString because the character at index 13 has a value of 8230 which is greater than 255.","error.type":"TypeError","task_type":"offload-l1"}
```

All 3 attempts (initial + 2 retries) failed with the same error.

## Root Cause

The character at index 13 has value 8230 (0x2026 = `…` ellipsis). This suggests the L1 extraction is passing conversation content through a protobuf `ByteString` conversion that only accepts ASCII (0-255), and the Unicode ellipsis character breaks it.

## Impact

- L1 extraction completely fails for any conversation containing the `…` character
- Retries also fail (all attempts use the same content)
- L2/L3 never run for affected sessions

## Suggested Fix

- Use a proper Unicode-safe encoding (e.g., UTF-8) for the L1 extraction payload
- Or sanitize/escape non-ASCII characters before passing to the ByteString conversion
- Or use a String field instead of ByteString for message content

## Related

- Issue #120178 (OpenClaw side, now closed) — was about SQLite scope, not this encoding issue

dailylights:
## 更新：根因确认（2026-08-09）

经过进一步调查，确认此错误来自**外部 tcvdb SDK 的 protobuf 层**，不是插件代码的问题。

### 根因

[...]

PR #923, opened 2026-08-10
fix: replace Unicode ellipsis (U+2026) before ByteString conversion

dailylights:
## Summary

L1 extraction fails with a protobuf `ByteString` TypeError when the conversation contains the Unicode ellipsis character `…` (U+2026, decimal 8230), which exceeds the ASCII 0-255 range.

## Root Cause

The offload backend receives conversation text via HTTP JSON, then converts it to protobuf ByteString. The character `…` (U+2026) at index 13 has value 8230 (`> 255`), which the protobuf ByteString conversion rejects.

## Fix

In `MemoryCore/src/utils/sanitize.ts` (`sanitizeText()`), replace `\u2026` (Unicode ellipsis) with ASCII `...` before any downstream ByteString/ASCII-only encoding.

## Testing

- `sanitizeText()` already handles similar cases (null chars, base64 data URIs, etc.)
- The fix is additive — no existing functionality is affected
- Verified: `\u2026` → `...` in all paths

## Related

- Issue #879: L1 extraction fails with TypeError: ByteString conversion rejects Unicode ellipsis (U+2026)

### DIFF
--- a/MemoryCore/src/utils/sanitize.ts
+++ b/MemoryCore/src/utils/sanitize.ts
@@ -70,6 +70,11 @@ export function sanitizeText(text: string): string {
   // become empty after sanitization and are naturally filtered by length checks.
   cleaned = cleaned.replace(/data:image\/[a-z+]+;base64,[A-Za-z0-9+/=]+/gi, "");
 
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:

PR #930, opened 2026-08-10
fix(memory-proxy): allow members to select team-visible agents

### LINKED ISSUE #924: [Bug] Team members cannot use team-visible agents in CodeBuddy session init; only admin/owner token works

code buddy cn 4.11.0

### Plugin Version | 插件版本

server_team version commit fe3230f176f1bf5832fee79d12494bbc2d19a8aa

### Operating System | 操作系统

Windows 11

_No response_

### Describe the bug | 问题描述

When a non-admin team member invokes the CodeBuddy model through the Memory Proxy, session initialization is bypassed (bypassed=true, hasAgentDetail=false), so skill/knowledge/tdai-memory injection hooks are skipped. Switching to the admin/owner user key makes the same agent available and injection works.

Root cause: MemoryProxy/src/session/codebuddy/init.ts calls metadataClient.listAgents(team_id, userId) for each team, which forces an owner_user_id filter on the MemoryCore /v3/meta/agent/list endpoint. As a result, team members who are not the agent's owner_user_id receive an empty agent list (totalAgents === 0) and hit the bypass branch. The agent's visibility is team in the DB and should be visible to all team members, but the Proxy's owner-scoped query ignores that visibility.

### To Reproduce | 复现步骤

Create a team and add a normal member (non-owner), e.g. usr-l9m3qrbf3l.

Create an agent agt-mchb46kw2u under that team with an admin/owner account.

Bind knowledge/code_graph assets to the agent in the Panel.

Configure CodeBuddy models.json to route glm-5.2 through Memory Proxy 
http://xxxx:8096/codebuddy/default.

Send a chat request from the normal member's IDE/CodeBuddy client.

Observe Proxy logs:
[session-init:cb] session=codebuddy:<sessionId> user=usr-l9m3qrbf3l has no active agents, passing through
[injection-debug] initResult ... bypassed=true hasAgentDetail=false
[session-init] ... bypassed → skipping all injection

Switch CodeBuddy to use the admin/owner user key and repeat → injection succeeds.

### Expected behavior | 预期行为

Team members should be able to see and use agents whose visibility='team' within the same team during session initialization, so that skill/knowledge/tdai-memory context injection works for them.

_No response_

### Additional context | 补充信息

_No response_

guomengjia618-dot:
## Summary

- Query active agents at team scope during CodeBuddy session initialization.
- Allow team members to select agents they own and agents with `visibility="team"`.
- Keep private agents owned by other team members hidden.
- Add a regression test covering owned private, team-visible, and other private agents.

## Root cause

CodeBuddy session initialization passed the current user ID to `MetadataClient.listAgents`, which added an `owner_user_id` filter to the team query. Team-visible agents created by an admin or another member were therefore omitted, causing session initialization and downstream memory injection to be bypassed.

## Impact

Team members can now select shared agents during CodeBuddy session initialization without gaining access to another member's private agents.

## Testing

- `npm test` in `MemoryProxy`
- Targeted strict TypeScript check for the new selector and regression test

Fixes #924

### DIFF
--- a/MemoryProxy/src/session/__tests__/accessible-agents.test.ts
+++ b/MemoryProxy/src/session/__tests__/accessible-agents.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, it, vi } from "vitest";
+
+import type { MetadataClient } from "../../meta/client.js";
+import { listAccessibleAgentsForTeam } from "../accessible-agents.js";
+
+describe("listAccessibleAgentsForTeam", () => {
+  it("includes owned agents and team-visible agents without exposing other private agents", async () => {
+    const listAgents = vi.fn().mockResolvedValue([
+      {
+        agent_id: "agent-owned-private",
+        team_id: "team-1",
+        owner_user_id: "member-1",
+        name: "My private agent",
+        visibility: "private",
+      },
+      {
+        agent_id: "agent-admin-team",
+        team_id: "team-1",
+        owner_user_id: "admin-1",
+        name: "Shared team agent",
+        visibility: "team",
+      },
+      {
+        agent_id: "agent-admin-private",
+        team_id: "team-1",
+        owner_user_id: "admin-1",
+        name: "Admin private agent",
+        visibility: "private",
+      },
+    ]);
+    const metadataClient = { listAgents } as unknown as MetadataClient;
+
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

PR #932, opened 2026-08-10
fix(memory-core): stamp cold-start floor at construction time, not capture time

### LINKED ISSUE #903: [Bug] The first turn of a new session is silently dropped ~48% of the time (cold-start cursor is `Date.now()` at capture time)
## Summary

`TdaiCore` resolves the cold-start floor with `Date.now()` **at capture time**
instead of at process start. Because the synthetic messages built by the Gateway
carry no `timestamp` and get `Date.now()` assigned microseconds later, both land
in the same millisecond as the cursor and are dropped by a strict `>` filter.

Measured on `memory-tencentdb@1.0.1`: **19 of 40 brand-new sessions lost their
opening turn entirely** (41 of 80 messages stored). No error, no warning,
`/capture` returns 200.

For a memory system this is the worst turn to lose: the opening turn is where
the user states the topic.

## Root cause

Four things line up:

1. `src/core/tdai-core.ts:303`
   ```ts
   pluginStartTimestamp: turn.startedAt ?? Date.now(),
   ```
   `turn.startedAt` is **never set by anyone**. It is declared in
   `src/core/types.ts:199` and the only caller of `handleTurnCommitted` is
   `src/gateway/server.ts:930`, which does not pass it. So the value is the
   clock **at the moment of capture**.

   This contradicts the field's own contract in `src/core/hooks/auto-capture.ts:62-65`:
   > `Epoch ms when the plugin was registered (cold-start time).`

2. `src/utils/checkpoint.ts:492-494` adopts that value as the cursor when the
   session has no prior `last_captured_timestamp`:
   ```ts
   if (afterTimestamp === 0 && pluginStartTimestamp && pluginStartTimestamp > 0) {
     afterTimestamp = pluginStartTimestamp;
   }
   ```

3. `src/gateway/server.ts:932-935` synthesizes the messages **without a
   `timestamp` field**, so `src/core/conversation/l0-recorder.ts:575` assigns
   `Date.now()` at extraction time — microseconds after step 1.

4. `src/core/conversation/l0-recorder.ts:158-161` filters with a **strict** `>`:
   ```ts
   const extracted = cursor !== 0 ? allExtracted.filter((m) => m.timestamp > cursor) : allExtracted;
   ```

If the millisecond has not ticked between (1) and (3) — the common case — every
message of the first turn fails `>` and is discarded before anything is written.

The comment at `l0-recorder.ts:146-157` asserts the opposite ("no
same-millisecond collision", "always > previous cursor"), which is true for the
OpenClaw host path but false for the Gateway path.

### It cascades

When both messages are filtered, `src/core/hooks/auto-capture.ts:135-137`
returns `null` and `src/utils/checkpoint.ts:498-505` leaves
`last_captured_timestamp` at 0 — so the **next** capture takes a fresh floor and
plays the same lottery. Losses are therefore contiguous from the start of a
session, not isolated.

## Reproduction

Start a Gateway with an empty data dir, then POST 40 captures each with a
**distinct** `session_key` (so each one is a session's first turn):

```bash
for i in $(seq -w 0 39); do
  curl -s -X POST 127.0.0.1:8420/capture \
    -H 'Content-Type: application/json' \
    -d "{\"user_content\":\"S$i first turn\",\"assistant_content\":\"S$i ack\",\"session_key\":\"sess-$i\"}" >/dev/null
  sleep 0.1
done
sqlite3 "$TDAI_DATA_DIR/vectors.db" \
  'select count(*) from l0_conversations, (select count(distinct session_key) c from l0_conversations);'
```

Observed (`memory-tencentdb@1.0.1`, macOS, node 22.22.3, sqlite backend,
embeddings disabled to remove noise):

| | messages stored | sessions with zero rows |
|---|---|---|
| unpatched | **41 / 80** | **19 / 40** |
| patched (below) | **80 / 80** | **0 / 40** |

Six consecutive fresh-Gateway runs of 10 captures each: 120/120 stored after the
patch, versus a 2.5% overall loss before it, with 100% of losses at capture
index 0.

## Proposed fix

Stamp the process start once, which is what the field claims to be:

```diff
--- a/MemoryCore/src/core/tdai-core.ts
+++ b/MemoryCore/src/core/tdai-core.ts
@@ class TdaiCore
   private readonly bgTasks = new Set<Promise<void>>();
+
+  /** Epoch ms when this core was constructed — the actual cold-start time. */
+  private readonly processStartedAt = Date.now();
@@ handleTurnCommitted
-      pluginStartTimestamp: turn.startedAt ?? Date.now(),
+      pluginStartTimestamp: turn.startedAt ?? this.processStartedAt,
```

Two notes on alternatives we rejected:

- **Changing `>` to `>=` does not work.** For every turn after the first, the
  cursor is the timestamp of the last stored message, so `>=` would re-import it
  as a duplicate. The strict comparison is correct; the cursor value is not.
- **Stamping `timestamp` on the synthetic messages in `server.ts:932-935`** also
  removes the collision and is a reasonable complement, but on its own it leaves
  `pluginStartTimestamp` violating its documented meaning.

Happy to open a PR.

ashvin-to:
## Summary

`pluginStartTimestamp` is the cold-start cursor floor used to filter messages on a sessions first capture. It was set to `Date.now()` at capture time (`handleTurnCommitted`), not at process start.

Gateway-synthesized messages also get `Date.now()` assigned microseconds later. When both land in the same millisecond, the strict `>` filter in `l0-recorder.ts` drops every message of the first turn. No error, no warning, `/capture` returns 200.

Measured by the issue reporter: **19 of 40 brand-new sessions lost their opening turn entirely**.

## Root cause

Four things line up:

1. `tdai-core.ts` resolves `pluginStartTimestamp` with `Date.now()` at the moment of capture
2. `checkpoint.ts` adopts that value as the cursor when no prior checkpoint exists
3. `server.ts` synthesizes messages without a `timestamp` field, so `l0-recorder.ts` assigns `Date.now()` at extraction time -- microseconds after step 1
4. `l0-recorder.ts` filters with strict `>`: same millisecond = dropped

## Fix

Store the timestamp once when `TdaiCore` is constructed (`processStartedAt`), use that as the fallback instead of a fresh `Date.now()` at each capture call. This matches the documented contract of the field ("epoch ms when the plugin was registered") and guarantees a time gap between the floor and gateway-assigned message timestamps.

```diff
--- a/src/core/tdai-core.ts
+++ b/src/core/tdai-core.ts
   private readonly bgTasks = new Set<Promise<void>>();

+  /** Epoch ms when this core was constructed -- the actual cold-start time. */
+  private readonly processStartedAt = Date.now();
+
   constructor(opts: TdaiCoreOptions) {
@@ handleTurnCommitted
-      pluginStartTimestamp: turn.startedAt ?? Date.now(),
+      pluginStartTimestamp: turn.startedAt ?? this.processStartedAt,
```

## Why not change `>` to `>=`

For every turn after the first, the cursor is the timestamp of the last stored message. `>=` would re-import it as a duplicate. The strict comparison is correct; the cursor value was not.

## Tests

All 67 existing tests pass (`vitest run`).

Fixes #903

### DIFF
--- a/src/core/tdai-core.ts
+++ b/src/core/tdai-core.ts
@@ -122,6 +122,9 @@ export class TdaiCore {
    */
   private readonly bgTasks = new Set<Promise<void>>();
 
+  /** Epoch ms when this core was constructed — the actual cold-start time. */
+  private readonly processStartedAt = Date.now();
+
   constructor(opts: TdaiCoreOptions) {
     this.hostAdapter = opts.hostAdapter;
     this.cfg = opts.config;
@@ -276,7 +279,7 @@ export class TdaiCore {
       scheduler: this.scheduler,
       originalUserText: turn.userText,
       originalUserMessageCount: turn.originalUserMessageCount,
-      pluginStartTimestamp: turn.startedAt ?? Date.now(),
+      pluginStartTimestamp: turn.startedAt ?? this.processStartedAt,
       vectorStore: this.vectorStore,
       embeddingService: this.embeddingService,
       bgTaskRegistry: this.bgTasks,


Maxwell-Code07:

PR #933, opened 2026-08-10
Add a request timeout to the MCP http client

bunlongheng:
I noticed callApi in MemoryKnowledge/src/mcp/http-client.ts fires its fetch with no timeout or abort signal. If the knowledge service is slow or hangs (a stuck CodeGraph or Wiki query, for example), the MCP tool call just sits there open forever and never returns, which ties up the caller.

Every other HTTP client in the repo already bounds its requests (auth.ts, the tcvdb and bm25 clients, meta/client, tdai/client, core-client all use AbortSignal.timeout or an abort controller), so this one looked like an oversight. I added a default 30s timeout with an optional timeoutMs override on HttpClientOptions so it lines up with the rest of the code. The existing catch block already handles the abort, so there is no behavior change on the happy path.

### DIFF
--- a/MemoryKnowledge/src/mcp/http-client.ts
+++ b/MemoryKnowledge/src/mcp/http-client.ts
@@ -10,10 +10,15 @@ import { createLogger } from "../logger.js";
 
 const log = createLogger("mcp-http");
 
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:

PR #937, opened 2026-08-10
fix(l1): inject user display name into extraction prompt

### LINKED ISSUE #936: L1 extraction prompt requires a user name but never injects one — memories carry inconsistent/garbled user names
## Summary

L1 memory extraction produces inconsistent or garbled **user names** inside memory `content` text. In a single-user deployment, extracted L1 records for the *same* user can carry different name labels (a generic placeholder, an occupation-like label, a first-person pronoun, or a misspelled variant of the user's display name). The structured `user_id` field is always correct — only the free-text name prefix is unreliable.

For team/enterprise deployments (the `work_memory` prompt mode with multiple users), this makes memory attribution text untrustworthy and hurts recall readability.

## Root cause

The extraction prompt **requires** the LLM to write a user name, but **never provides one**.

1. `MemoryCore/src/core/prompts/l1-extraction.ts` (`EXTRACT_MEMORIES_SYSTEM_PROMPT`):
   - line 34: "提取主体必须以「用户（姓名）」或「AI」为核心" (memories must be centered on "user (name)" or "AI")
   - lines 42/48/54: sentence templates `用户（[姓名]）喜欢/是/擅长…`, `用户（[姓名]）在 [时间] 于 [地点] …`, `用户要求/希望 AI…`
   - `[姓名]` is a placeholder with **no defined source** anywhere in the prompt.

2. `formatExtractionPrompt()` (same file, lines 389–417) injects only `previousSceneName`, `backgroundMessages` and `newMessages`. No user identity is passed.

3. `callLlmExtraction()` (`MemoryCore/src/core/record/l1-extractor.ts`, lines 387–448) accepts `newMessages / backgroundMessages / previousSceneName / config / model / promptMode` — no user field either.

4. The system *does* store a user display name (metadata `UserEntity.name`, accessible via `getUserById` in `src/metadata/store/interface.ts`), but the L1 extraction path never queries it. Zero references to user display name exist in the extraction pipeline (`pipeline-factory.ts` → `extractL1Memories` → `callLlmExtraction`).

So the LLM must guess the name from message text: it copies a name when one appears in context (and can mis-copy it), falls back to a generic 「用户」, invents an occupation-like label, or preserves the user's first-person wording 「我」.

## Reproduction

1. Deploy memory-core with a user whose display name differs from their `user_id`.
2. Have a few conversations, some where the name is mentioned in messages, some where it is not, some where the user speaks in first person.
3. Run L1 extraction and inspect the `content` of extracted atomic memories.

Observed patterns on a single-user instance (all records share the same `user_id`, but `content` varies):

- `用户（用户）…` — generic placeholder when no name appears in context
- `用户（开发者）…` — invented occupation-like label
- `用户（我）…` — first-person wording preserved from the user's own message
- `用户（<display name, misspelled>）…` — name copied from context but garbled

## Expected behavior

All extracted memories for the same user should carry the same, correctly spelled user name — or no name at all, relying on the structured `user_id` field for attribution.

## Suggested fix

Inject the resolved user display name into the extraction prompt:

1. In `pipeline-factory.ts`, resolve the user's display name via metadata `getUserById(userId).name` (optional — standalone/OpenClaw mode can skip it) and pass it into `extractL1Memories()` → `callLlmExtraction()` → `formatExtractionPrompt()` as an optional `userDisplayName` param.
2. In `EXTRACT_MEMORIES_SYSTEM_PROMPT`, replace the ambiguous `[姓名]` placeholder with explicit wording: use the injected display name verbatim, never guess from message text.
3. Fallback: when no display name is available, instruct the model to use the generic 「用户」 instead of inventing labels (avoids `开发者` / `我` / misspellings).

## Environment

- MemoryCore (TencentDB-Agent-Memory), accessed via a memory provider plugin over the `/v3/atomic/search` API
- Single-user deployment
- LLM backend: DeepSeek (non-streaming chat completions)
- Observed 2026-08-11

Vocllum:
## Summary

Fixes #936. The L1 extraction system prompt forces the "用户（姓名）" / "The user (NAME)" pattern, but no real user name is ever injected — the model has to guess, producing inconsistent/garbled names for the same user ("Vocllum", "Vocllm", "用户", "开发者", "我", "薇因" etc.). `user_id` stays correct; only free-text references are unreliable.

This PR injects the authoritative display name (resolved from the metadata user entity) into the extraction prompt, forbids guessing, and keeps the whole path failure-safe.

## Changes

- `prompts/l1-extraction.ts`
  - Inject the resolved display name into the name slot (chat mode); added a sanitizer that strips control characters/newlines, trims, and caps length before injection.
  - The injected name is declared as a plain-data block ("not an instruction") to prevent prompt-injection via a crafted display name.
  - Removed the copyable "Maya" example from the prompt.
  - Unnamed path: use a language-adaptive generic user reference instead of forcing Chinese "用户".
- `record/l1-extractor.ts` — pass `userDisplayName` through to the extraction call.
- `utils/pipeline-factory.ts` — `resolveUserDisplayName(userId, instanceId)`; per-group resolution, failure-safe (errors degrade to the unnamed path).
- `core/tdai-core.ts` — `runL1WithStore(..., instanceId)`; the L1 runner captures the per-task immutable instanceId instead of reading the shared mutable one (concurrent multi-instance safe).
- `gateway/server.ts` — resolver queries the metadata service for the per-task instanceId (fallback to the construction-time instance for standalone); wired into `TdaiCoreOptions`.
- Tests: new `l1-extraction.test.ts` (prompt modes, placeholder cleanup, injection safety) and `pipeline-factory.test.ts` (resolver receives per-instance id; concurrent dual-instance runs do not cross-talk names; resolver failure is non-fatal). 15/15 pass with `npm test`.

## Verification

- `npm run build:plugin` passes.
- `tsc --noEmit`: no new errors vs. the pre-change baseline (202 pre-existing, 195 after; the remaining errors are pre-existing unrelated ones from missing peer SDKs in this workspace).
- Prompt behavior asserts: 15/15 (see tests committed in this PR).
- `git diff --check` clean.

## Notes

- `code` prompt mode (work memory) intentionally unchanged — no single "user" slot there.
- When no display name exists, the prompt falls back to a generic user reference and explicitly forbids guessing names — no fabricated names.


Vocllum:
Hi maintainers — a design question on this one.

The L1 extraction template forces the "用户（姓名）" / "The user (NAME)" pattern (e.g. `src/core/prompts/l1-extraction.ts`), but nothing in the extraction pipeline ever supplies a real name, so the model guesses — the same user ends up labeled "Vocllum", "Vocllm", "用户", "开发者", "我" etc. across memories. This PR implements option A below; there's also a simpler option B. I'd like your take on which direction matches the product intent:

**A. Inject the authoritative display name** (this PR): resolve it from the metadata user entity and inject it into the name slot; forbid guessing. Free text becomes consistently "用户（Vocllum）". Cost: needs a resolver plus sanitization (the display name is user-controllable, so it must be treated as data, not instructions — see the sanitizer in this PR).

**B. Drop the name entirely**: use a generic "用户 / the user" and rely on the structured `user_id` field for attribution. Much smaller change, zero injection surface, but loses in-text attribution readability in multi-user deployments.

Why does the template require a name in the first place — is in-text attribution (e.g. enterprise dashboards) the intent? Happy to rework the PR toward B if that fits better.

Maxwell-Code07:

PR #938, opened 2026-08-11
fix(memory-core): filter English FTS stop words

### LINKED ISSUE #901: fix(search): buildFtsQuery filters Chinese stop-words but has no English equivalent — RRF then demotes semantically-matched results
## Summary

`buildFtsQuery()` removes high-frequency Chinese function words via `ZH_STOP_WORDS`, but there is no equivalent list for English. English queries are therefore emitted as an OR of function words, e.g. `"where" OR "does" OR "the" OR "user" OR "live"`.

Because the write side (`tokenizeForFts()`) does not drop stop-words either, `"the"` is present in the FTS index of virtually every English document. The FTS candidate list becomes close to "all English documents". Hybrid search then merges FTS and vector lists with Reciprocal Rank Fusion, where appearing in *both* lists is what wins — so a mediocre document present in both outranks an excellent document present only in the vector list.

Net effect: hybrid search systematically penalizes exactly the documents that only semantic search can find, which is the case the vector path exists for.

Version: `@tencentdb-agent-memory/memory-tencentdb@1.0.1` (npm tarball). Line numbers below are from that tarball's `src/`; the corresponding built code is in `dist/index.mjs` (`ZH_STOP_WORDS` at L11635, `buildFtsQuery` at L11698, `tokenizeForFts` at L11734).

## Root cause

`src/core/store/sqlite.ts:154` — the stop-word list is Chinese-only:

```ts
const ZH_STOP_WORDS = new Set([
  "的", "了", "在", "是", "我", "有", "和", "就", "不", "人", "都", "一",
  ...
]);
```

`src/core/store/sqlite.ts:182-214` — `buildFtsQuery()`. The jieba branch filters only against `ZH_STOP_WORDS` (`sqlite.ts:197`), and the regex fallback branch (`sqlite.ts:202-209`) applies no stop-word filter at all:

```ts
      .filter((t) => {
        if (!t) return false;
        if (!/[\p{L}\p{N}]/u.test(t)) return false;
        if (ZH_STOP_WORDS.has(t)) return false;   // sqlite.ts:197 — no EN equivalent
        return true;
      });
```

Tokens are then OR-joined (`sqlite.ts:212-213`).

`src/core/store/sqlite.ts:237-250` — `tokenizeForFts()` (write side) joins all jieba tokens with no stop-word filtering, so the indexed text of every English record contains `the`, `is`, `of`, etc. Called at `sqlite.ts:1059`, `1471`, `2358`, `2401` when populating `l1_fts` / `l0_fts` (`sqlite.ts:755`, `sqlite.ts:773`).

Fusion happens in the per-tool RRF helpers, which use the standard `1/(k + rank + 1)` sum with `k = 60`:

- `src/core/tools/conversation-search.ts:58` (`rrfMergeL0`), invoked at `conversation-search.ts:242`; FTS query built at `conversation-search.ts:171`
- `src/core/tools/memory-search.ts:57` (`rrfMergeL1`), invoked at `memory-search.ts:251`; FTS query built at `memory-search.ts:176`
- the shared generic version lives at `src/core/store/search-utils.ts:38` (`rrfMerge`)

Because each list contributes an additive term, a document ranked ~30th in FTS and ~10th in vector beats a document ranked 1st in vector only:

- both lists: `1/(60+30+1) + 1/(60+10+1) ≈ 0.0110 + 0.0141 = 0.0251`
- vector only, rank 1: `1/(60+0+1) ≈ 0.0164`

The near-universal `"the"` match is what puts the weaker document into the FTS list in the first place.

Same call site pattern in the other consumers of `buildFtsQuery()`:

- `src/core/hooks/auto-recall.ts:638` (hybrid auto-recall keyword leg)
- `src/core/hooks/auto-recall.ts:515` (`searchByKeyword`, FTS5-only strategy — here the noise shows up directly as BM25 candidates rather than through RRF)
- `src/core/record/l1-dedup.ts:255` (`findCandidatesByFts`, L1 dedup candidate lookup)

### Cross-language case

A fact stored in Chinese scores zero on FTS for an English query (no shared tokens), so it can only ever appear in the vector list — i.e. it always gets the single-list RRF score while any English record that happens to contain `the` gets the two-list score. Mixed-language memory stores are hit hardest.

## Reproduction

The query-side behavior reproduces standalone, with the exact token pipeline of `sqlite.ts:182-214`:

```js
const { Jieba } = require("@node-rs/jieba");
const { dict } = require("@node-rs/jieba/dict");
const j = Jieba.withDict(dict);

for (const q of ["where does the user live", "What is the name of the pet?"]) {
  const toks = j.cutForSearch(q, true)
    .map((t) => t.trim())
    .filter((t) => t && /[\p{L}\p{N}]/u.test(t));
  console.log(JSON.stringify(q), "=>", [...new Set(toks)].map((t) => `"${t}"`).join(" OR "));
}
```

Output (jieba from this package's own dependency tree):

```
"where does the user live" => "where" OR "does" OR "the" OR "user" OR "live"
"What is the name of the pet?" => "What" OR "is" OR "the" OR "name" OR "of" OR "pet"
```

End-to-end: store English memories, then query with a paraphrase that shares no content word with the stored text. With `logger.debug` enabled, the `[hybrid-fts] FTS5 query:` line shows the stop-word OR query, and the `[hybrid] RRF merged:` line shows an FTS candidate list roughly the size of the English corpus.

## Measurement

Local harness, SQLite backend, embeddings `text-embedding-3-large` @1536 dims. Corpus of 13 facts; the "semantic" query set is deliberately paraphrased so queries share no content word with the stored fact; the "lexical" control set shares content words.

| Query set | Metric | Before | After (EN stop-word filter) |
|---|---|---|---|
| Semantic (paraphrase) | hit@5 | 10/13 | 13/13 |
| Semantic (paraphrase) | MRR | 0.592 | 0.962 |
| Lexical control | hit@5 | 13/13 | 13/13 |

The lexical control is unchanged, i.e. keyword search is not degraded — the removed tokens carry no retrieval signal in an OR query where every document matches them.

For the 4 semantic queries that failed before the change, pure cosine similarity ranks the correct document at position 1 in all 4; 3 of those did not appear in the hybrid top-5 at all. The vector path had the right answer; RRF discarded it.

## Suggested fix

Mirror what the code already does for Chinese. Add an `EN_STOP_WORDS` set next to `ZH_STOP_WORDS` at `sqlite.ts:154` and check it in `buildFtsQuery()` — in both the jieba branch and the regex fallback, since the fallback currently has no stop-word filter at all:

```ts
/**
 * Common English stop-words that add noise to FTS5 queries.
 * Kept small on purpose — only high-frequency function words.
 */
const EN_STOP_WORDS = new Set([
  "a", "an", "the", "and", "or", "but", "if", "then", "than", "that", "this",
  "is", "are", "was", "were", "be", "been", "am", "do", "does", "did",
  "have", "has", "had", "i", "you", "he", "she", "it", "we", "they",
  "of", "in", "on", "at", "to", "for", "with", "by", "from", "as",
  "what", "which", "who", "when", "where", "why", "how",
  "can", "could", "will", "would", "should", "may", "might", "must",
  "not", "no", "so", "very", "just", "also", "any", "some", "all",
]);
```

```ts
       .filter((t) => {
         if (!t) return false;
         if (!/[\p{L}\p{N}]/u.test(t)) return false;
         if (ZH_STOP_WORDS.has(t)) return false;
+        if (EN_STOP_WORDS.has(t.toLowerCase())) return false;
         return true;
       });
```

```ts
     // Fallback: simple Unicode regex split
     tokens =
       raw
         .match(/[\p{L}\p{N}_]+/gu)
         ?.map((t) => t.trim())
-        .filter(Boolean) ?? [];
+        .filter((t) => Boolean(t) && !EN_STOP_WORDS.has(t.toLowerCase())) ?? [];
```

Notes on scope:

- Query-side only. `tokenizeForFts()` is untouched, so no reindex or migration is required.
- Degenerate queries are already handled: if every token is filtered out, `buildFtsQuery()` returns `null` at `sqlite.ts:211`, and callers skip the FTS leg (`conversation-search.ts:172-175`, `memory-search.ts:177-180`), leaving vector-only search — which is the correct behavior for an all-stop-word query.
- The fix is symmetric with the existing `ZH_STOP_WORDS` handling twenty lines above, and inherits the same "kept small on purpose" rule: function words only, nothing with semantic load. It applies to every `buildFtsQuery()` consumer at once (`conversation-search.ts:171`, `memory-search.ts:176`, `auto-recall.ts:515`, `auto-recall.ts:638`, `l1-dedup.ts:255`).

Happy to send a PR if the approach and list scope look right. Related but distinct: #160 touches the same function (FTS5 operator sanitization).

quan020406:
## Description | 描述

Fixes Issue #901.

Filter low-signal English function words from `buildFtsQuery()` before generating SQLite FTS5 queries. The stop-word list is intentionally conservative to preserve technical, negation, and relational terms such as `where`, `in`, `or`, `not`, and `to`.

The same filtering predicate is applied to both jieba and Unicode-regex fallback paths. Existing Chinese stop-word behavior is preserved and covered by regression tests.

## Related Issue | 关联 Issue

Fix #901

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Validation completed:

- `npm.cmd test` — 9 tests passed
- `npm.cmd run build:plugin` — passed
- `npm.cmd pack --dry-run --ignore-scripts` — passed
- `git diff --check` — passed

The full `npm.cmd run build` remains blocked by the existing missing
`scripts/seed-v2/tsconfig.json` baseline issue, unrelated to this change.

### DIFF
--- a/MemoryCore/src/core/store/sqlite-stop-words.test.ts
+++ b/MemoryCore/src/core/store/sqlite-stop-words.test.ts
@@ -1,0 +1,89 @@
+import { afterEach, describe, expect, it } from "vitest";
+
+import {
+  _resetJiebaForTest,
+  _setJiebaForTest,
+  buildFtsQuery,
+} from "./sqlite.js";
+
+describe("buildFtsQuery English stop-word filtering", () => {
+  afterEach(() => {
+    _resetJiebaForTest();
+  });
+
+  it("filters English function words from jieba tokens while preserving content words", () => {
+    _setJiebaForTest({
+      cutForSearch: () => ["does", "the", "user", "live"],
+    });
+
+    expect(buildFtsQuery("does the user live")).toBe(
+      '"user" OR "live"',
+    );
+  });
+
+  it("filters English function words in the Unicode-regex fallback path", () => {
+    _setJiebaForTest(null);
+
+    expect(buildFtsQuery("does the user live")).toBe(
+      '"user" OR "live"',
+    );
+  });
+
+  it("applies case-insensitive filtering and preserves negation in jieba", () => {
+    _setJiebaForTest({
+      cutForSearch: () => ["THE", "user", "IS", "not", "allergic"],
+    });
+
+    expect(buildFtsQuery("THE user IS not allergic")).toBe(
+      '"user" OR "not" OR "allergic"',
+    );
+  });
+
+  it("keeps technical tokens while filtering case-insensitive function words", () => {
+    _setJiebaForTest(null);
+
+    expect(buildFtsQuery("THE user IS in Berlin")).toBe(
+      '"user" OR "in" OR "Berlin"',
+    );
+  });
+
+  it("keeps negation words searchable", () => {
+    _setJiebaForTest(null);
+
+    expect(buildFtsQuery("user is not allergic to peanuts")).toBe(
+      '"user" OR "not" OR "allergic" OR "to" OR "peanuts"',
+    );
+  });
+
+  it("keeps technical query terms such as SQL keywords searchable", () => {
+    _setJiebaForTest(null);
+
+    expect(buildFtsQuery("WHERE user_id IN table")).toBe(
+      '"WHERE" OR "user_id" OR "IN" OR "table"',
+    );
+  });
+
+  it("returns null when a query contains no searchable tokens", () => {
+    _setJiebaForTest(null);
+
+    expect(buildFtsQuery("the does is are")).toBeNull();
+  });
+
+  it("preserves the existing Chinese stop-word behavior", () => {
+    _setJiebaForTest({
+      cutForSearch: () => ["用户", "的", "喜欢", "编程"],
+    });
+
+    expect(buildFtsQuery("用户的喜欢编程")).toBe(
+      '"用户" OR "喜欢" OR "编程"',
+    );
+  });
+
+  it("applies Chinese stop-word filtering in the Unicode-regex fallback path", () => {
+    _setJiebaForTest(null);
+
+    expect(buildFtsQuery("用户 的 喜欢 编程")).toBe(
+      '"用户" OR "喜欢" OR "编程"',
+    );
+  });
+});
--- a/MemoryCore/src/core/store/sqlite.ts
+++ b/MemoryCore/src/core/store/sqlite.ts
@@ -184,6 +184,27 @@ const ZH_STOP_WORDS = new Set([
 ]);
 
 /**
+ * Common English function words that add noise to FTS5 OR queries.
+ * Keep this list conservative: negation, quantity, identity, and domain
+ * words should remain searchable.
+ */
+// Keep this list conservative: technical, negation, and relational terms can
+// be meaningful search tokens in database, code, and requirements content.
+const EN_STOP_WORDS = new Set([
+  "a", "an", "the",
+  "is", "are", "was", "were", "be", "been", "being",
+  "do", "does", "did", "have", "has", "had",
+]);
+
+function isSearchableFtsToken(token: string): boolean {
+  if (!token) return false;
+  if (!/[\p{L}\p{N}]/u.test(token)) return false;
+  if (ZH_STOP_WORDS.has(token)) return false;
+  if (EN_STOP_WORDS.has(token.toLowerCase())) return false;
+  return true;
+}
+
+/**
  * Build an FTS5 MATCH query from raw text.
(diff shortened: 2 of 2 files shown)


Maxwell-Code07:

PR #942, opened 2026-08-11
fix(embedding): retry transient API failures (#927)

### LINKED ISSUE #927: [Bug] L0 embedding timeout silently degrades to metadata-only write with code:0 — MAX_RETRIES=0 leaves retry scaffold disabled

N/A (standalone gateway deployment)

### Plugin Version | 插件版本

v2.0.0 (commit 0aff21a)

### Operating System | 操作系统

macOS (Node.js 22)

Remote OpenAI-compatible embedding provider (Qwen3-Embedding-8B, 4096 dims), SQLite backend, service mode gateway on port 8420.

### Describe the bug | 问题描述

When the embedding API times out, the gateway silently degrades the affected message to a **metadata-only write** (no vector) while still returning `code: 0` to the caller. Callers have no way to detect the data loss, and semantic recall degrades silently.

**Root cause chain (v2.0.0 source, verified against pinned commit 0aff21a):**

1. `src/core/store/embedding.ts:369` — `MAX_RETRIES = 0`: the retry loop executes exactly once; an `AbortError` (timeout) is never retried.
2. `src/core/store/embedding.ts:371` — `DEFAULT_API_TIMEOUT_MS = 10_000`: a single slow API call (>10s) aborts the whole embedding for that message.
3. `src/gateway/v2-router.ts:720` — `handleConversationAdd` catches the exception with only a `console.warn` and continues with `emb = undefined`:
   ```ts
   try { emb = await embedding.embed(msg.content); } catch (e) { console.warn(`[v2-router] L0 embedding failed:`, e); }
   await store.upsertL0(record, emb);   // emb undefined → metadata-only write
   acceptedIds.push(id);                 // message reported as accepted
   ```
4. The HTTP response is `code: 0` with no per-message embedding status — the caller cannot distinguish a fully-vectorized write from a partially-degraded one.

**Observed impact over 6 days of production use** (incremental ingestion of ~260 markdown files, one L0 session per file, messages added via `/v2/conversation/add` in batches of 20):

- 24 `DOMException [AbortError]` events across 4 ingestion days (Aug 5 / 7 / 9 / 10).
- Each timed-out message is persisted **without** a vector row, leaving `l0_conversations` rows with no matching `l0_vec` entry.
- Probability scales with message count: a 40-message session succeeded 40/40 on one run, while a 44-message session of the same file (same content size) hit 3 timeouts on a later run. For sessions of ~44 messages, P(≥1 timeout) ≈ 78% given our observed per-call failure rate (~3%).
- Detected only because we built an external post-write integrity check against the SQLite database.

### To Reproduce | 复现步骤

1. Deploy the gateway with a remote embedding provider whose p99 latency occasionally exceeds 10s.
2. Add a session with ~40+ messages via `/v2/conversation/add`.
3. Repeat across days; compare `SELECT COUNT(*) FROM l0_conversations WHERE session_id=?` against vector rows for the same session.
4. Observe rows without vectors; the add API returned `code: 0` for all of them.

### Expected behavior | 预期行为

Proposed fixes (any/all):
- **A (retry):** set `MAX_RETRIES >= 2` with backoff for timeout/5xx — the retry scaffold already exists but is disabled by `MAX_RETRIES = 0`.
- **B (visibility):** surface embedding failures to the caller — e.g. a per-message `embedding_status` field in the response, or a non-zero code when any message in the batch lost its vector.
- **C (config):** make the per-call timeout configurable (`EmbeddingConfig.timeoutMs` exists but is not exposed in documented config samples; default 10s is tight for 4096-dim models under load).
- **D (repair):** a background reconciliation pass that re-embeds L0 rows missing vectors (similar in spirit to #156's reconciliation idea).

```
[v2-router] L0 embedding failed: DOMException [AbortError]: This operation was aborted
    at async OpenAIEmbeddingService._callApi (src/core/store/embedding.ts:527:24)
    at async OpenAIEmbeddingService.embed (src/core/store/embedding.ts:458:22)
    at async handleConversationAdd (src/gateway/v2-router.ts:720:19)
```
Immediately followed by `[L0-upsert] ... (no embedding — metadata-only write)` and `[L0-upsert] OK ... (meta-only)`.

### Additional context | 补充信息

Related: #236 (silent L0 embedding failure via batch-size mismatch — different trigger, same silent-degradation class), #156 (JSONL/SQLite reconciliation after embedding failure, closed). This report covers the timeout path specifically.

We currently mitigate with a client-side post-write integrity check + bounded rewrite, but the silent `code: 0` contract is the root problem.

YOMXXX:
## Summary

Fixes #927 by enabling the existing bounded retry path for transient embedding API failures.

When the embedding provider returns a 5xx response or a request timeout occurs, the service now retries up to two times with the existing exponential backoff before failing. This reduces silent metadata-only L0 writes caused by transient embedding failures.

## Root cause

`OpenAIEmbeddingService` had `MAX_RETRIES = 0`, so the retry loop executed only once even though timeout and 5xx errors were classified as retryable.

## Validation

- `npm test` — 5 test files, 17 tests passed for this PR scope
- `git diff --check` — passed

## Scope

This PR intentionally covers the bounded retry portion of #927. It does not change the existing response contract or add a background reconciliation workflow.

### DIFF
--- a/MemoryCore/hermes-plugin/memory/memory_tencentdb/__init__.py
+++ b/MemoryCore/hermes-plugin/memory/memory_tencentdb/__init__.py
@@ -910,7 +910,29 @@ class MemoryTencentdbProvider(MemoryProvider):
 
     def on_memory_write(self, action: str, target: str, content: str) -> None:
         """Mirror built-in memory writes to memory-tencentdb for indexing."""
-        pass
+        if action.strip().lower() != "add":
+            return
+        if not content.strip():
+            return
+        if not self._ensure_alive_for_request() or not self._client:
+            return
+
+        try:
+            result = self._client.write_explicit_memory(
+                content=content,
+                target=target,
+                action=action,
+                session_id=self._session_id,
+                user_id=self._user_id,
+            )
+            data = result.get("data") or {}
+            if data.get("stored") is False:
+                raise RuntimeError("Gateway did not index explicit memory")
+            self._record_success()
+        except Exception as e:
+            self._record_failure()
+            logger.warning("memory-tencentdb explicit memory sync failed: %s", e)
+            self._try_recover_gateway()
 
     def on_session_end(self, messages: List[Dict[str, Any]]) -> None:
         """Trigger session-level flush on the Gateway.
--- a/MemoryCore/hermes-plugin/memory/memory_tencentdb/client.py
+++ b/MemoryCore/hermes-plugin/memory/memory_tencentdb/client.py
@@ -154,6 +154,36 @@ class MemoryTencentdbSdkClient:
             body["session_id"] = session_id
         return self._post("/v3/conversation/search", body)
 
+    def write_explicit_memory(
+        self,
+        content: str,
+        *,
+        target: str = "memory",
+        action: str = "add",
+        session_id: str = "",
+        team_id: str = "default",
+        agent_id: str = "default",
+        user_id: str = "default",
+    ) -> Dict[str, Any]:
+        """Mirror a host-native durable-memory write directly into L1.
+
+        (#417) Hermes built-in ``memory(action="add", ...)`` writes are mirrored
+        here so they become retrievable via memory search, instead of being
+        swallowed by the provider. Stored on the v3 data plane so tenancy
+        isolation applies.
+        """
+        body: Dict[str, Any] = {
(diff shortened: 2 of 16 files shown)


Maxwell-Code07:

PR #943, opened 2026-08-11
docs(benchmark): align SWE-bench token reduction (#813)

### LINKED ISSUE #813: docs(benchmark): SWE-bench token reduction (−33.09%) does not reproduce from the printed token counts (computed −31.63%)
### Summary

The SWE-bench token-reduction figure in the benchmark table does not reproduce from the token counts printed in the same row.

This is **not** a reproduction attempt — I have not run any benchmark. It is purely an internal-consistency check of the published table against itself.

### Where

`main` @ `3c6fc6425f22d24c4917dc3b7791a175d2d13545`, `README.md` line 41 (same row is present on `feat/server` @ `e57b8cb7`):

```
| **Short-term** | SWE-bench | 58.4% | **64.2%** | **+9.93%** | 3474.1M | **2375.4M** | **−33.09%** |
```

### The discrepancy

```
(3474.1 − 2375.4) / 3474.1 = 31.63%     ← computed from the printed values
                              33.09%     ← published
                              -1.46pp
```

For **−33.09%** to be correct with an *after* value of 2375.4M, the *before* value would need to be **≈3550.1M**, not 3474.1M. So one of the three printed numbers in that row appears to be off — most likely a before-token count from a slightly different run than the one the percentage was derived from.

### What does reproduce (for contrast)

Every other figure in the table is internally consistent, which is why this one row stands out:

| Row | Published | Computed | |
|---|---|---|---|
| WideSearch tokens | −61.38% | −61.30% | ok (rounding) |
| AA-LCR tokens | −30.98% | −30.98% | exact |
| **SWE-bench tokens** | **−33.09%** | **−31.63%** | **−1.46pp** |
| WideSearch pass rate | +51.52% | +51.52% | exact |
| SWE-bench pass rate | +9.93% | +9.93% | exact |
| AA-LCR pass rate | +7.95% | +7.95% | exact |

So the issue is isolated to the SWE-bench token columns; the pass-rate methodology looks fine.

### Possibly relevant history

`3cc6d2a` ("docs: fix benchmark data", 2026-05-13) revised the WideSearch row substantially — `8.5%→12%` / `174.31M→63.46M` / `−63.59%` became `33%→50%` / `221.31M→85.64M` / `−61.38%`. The SWE-bench row was left unchanged in that pass, so if the numbers came from an earlier run set, this row may simply have been missed.

### Ask

Could you confirm which of the SWE-bench values is authoritative — the `3474.1M` before-count, or the `−33.09%` reduction — and correct the other? A one-line note on how token totals are aggregated (per-session vs per-task, and whether cached prefix tokens are counted) would also help readers reconcile the table.

### Minor, related

The table appears on `main` and `feat/server`, but not on `feat/server_team`, which is currently the repository's **default branch** — its Benchmark section lists PersonaMem only. Readers landing on the repo home page therefore see a different benchmark section than the one above. If that divergence is unintentional, it may be worth syncing.

Related (not duplicates): #73 concerns empirical non-reproduction with a local model; #106 and #812 request a reproducible evaluation harness. This report is narrower — it needs no runs, only the printed values.

Thanks for open-sourcing the project.

YOMXXX:
## Summary

- sync the benchmark table into the default `feat/server_team` README pages
- correct the SWE-bench token reduction from `-33.09%` to `-31.63%`, matching the displayed `3474.1M` and `2375.4M` totals
- document that displayed token totals are rounded and explain the calculation
- keep English and Chinese README benchmark sections consistent

Fixes #813

## Verification

- Recomputed `(3474.1 - 2375.4) / 3474.1 = 31.63%`
- `git diff --check`
- Reviewed both `README.md` and `README_CN.md` for matching benchmark rows

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,9 +245,14 @@ This makes documents and code part of memory as well — but they remain availab
 
 ## Benchmark
 
-| Benchmark | Without TencentDB Agent Memory | With it enabled | Relative improvement |
-| :--- | :---: | :---: | :---: |
-| **PersonaMem** | 48% | **76%** | **+59%** |
+| Memory capability | Benchmark | OpenClaw success | With TencentDB Agent Memory | Relative Δ | OpenClaw tokens | With TencentDB Agent Memory | Relative Δ |
+| :--- | :--- | :---: | :---: | :---: | :---: | :---: | :---: |
+| **Short-term** | WideSearch | 33% | **50%** | **+51.52%** | 221.31M | **85.64M** | **−61.38%** |
+| **Short-term** | SWE-bench | 58.4% | **64.2%** | **+9.93%** | 3474.1M | **2375.4M** | **−31.63%** |
+| **Short-term** | AA-LCR | 44.0% | **47.5%** | **+7.95%** | 112.0M | **77.3M** | **−30.98%** |
+| **Long-term** | PersonaMem | 48% | **76%** | **+59%** | — | — | — |
+
+These results are measured over continuous long-horizon sessions rather than isolated turns. Token totals are rounded for display; the SWE-bench percentage is aligned with the displayed before/after totals (`(3474.1M − 2375.4M) / 3474.1M = 31.63%`).
 
 PersonaMem tests whether an Agent can correctly understand and apply user information after extended interactions.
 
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #944, opened 2026-08-11
[good first issue-platform adapt-OpenCode] Add Memory adapter for OpenCode

PAUL01zjb:
Zero-dependency OpenCode plugin that captures session.idle transcripts into the Memory Gateway and exposes a memory_search recall tool. Includes bilingual (EN/CN) docs.

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

### DIFF
--- a/adapters/opencode-memory/README.md
+++ b/adapters/opencode-memory/README.md
@@ -5,12 +5,27 @@ OpenCode adapter that gives every agent session persistent memory backed by
 
 ## Features
 
-- On `session.idle`, the turn transcript is written to the Memory Gateway (L0 conversation).
-- A `memory_search` tool lets the agent recall structured memories (`/v3/atomic/search`) mid-session.
-- Framework-agnostic: talks to the Gateway REST API directly (see
-  `MemoryCore/hermes-plugin/memory/memory_tencentdb/client.py` for the contract).
+- On `session.idle`, the finished session is read through the official OpenCode
+  SDK (`client.session.messages`) and persisted to the Memory Gateway as an L0
+  conversation (`POST /v3/conversation/add`).
+- A `memory_search` tool lets the agent recall structured memories
+  (`POST /v3/atomic/search`) mid-session.
+- Pull-based recall: the agent decides when to search, so memory context is
+  never silently injected into every prompt.
 - Zero dependencies: no build step, no transitive packages — the plugin is a
-  single auditable file that runs on any Node ≥ 18 runtime.
+  single auditable ESM file that runs on any Node ≥ 18 runtime.
+
+## Why this adapter
+
+- **Auditable by design.** The whole adapter is one file. Capture and recall
+  stay thin; extraction, embedding and retrieval stay in MemoryCore.
+- **No placeholder data.** If the transcript cannot be read, capture is skipped
+  with a warning instead of writing a fake message into memory.
+- **Deduplicated turns.** The same completed turn is persisted only once, even
+  if `session.idle` fires repeatedly.
+- **Bounded failure.** Every gateway call has a timeout and retries transient
+  failures once; the v3 envelope `{code, message, data}` is validated so the
+  agent sees real errors instead of silent partial success.
 
 ## How it works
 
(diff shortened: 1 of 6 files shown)


Maxwell-Code07:

yangjj-iso (changes requested):
发现一个具体的 API 用法问题，已定位到代码：

📍 adapters/opencode-memory/src/index.js:118

    "x-tdai-service-id": cfg.agentId,

这里把 x-tdai-service-id 这个 HTTP header 直接赋值成了 cfg.agentId
（定义见 index.js:32 DEFAULTS.agentId = "opencode"，index.js:49
loadConfig() 里通过 OPCODE_MEMORY_AGENT_ID 覆盖）。

但按你在文件头注释（index.js:24）自己点名引用的参考实现
MemoryCore/hermes-plugin/memory/memory_tencentdb/client.py，
service_id 是一个独立配置项，默认值 "default"，代表"访问哪个 Memory
服务/实例"；而 agent_id 是请求体里的业务隔离字段，代表"业务侧的智能体
身份"。两者是完全不同的维度，不应该复用同一个值。

如果 Gateway 端对 x-tdai-service-id 做实例级校验（比如白名单只认
"default" 或某个固定部署 ID），这个适配器发出的请求会被直接拒绝或
路由到错误的实例。

建议修复：新增一个独立的配置项（如 OPCODE_MEMORY_SERVICE_ID，默认值
"default"），在 index.js:32/49 DEFAULTS 和 loadConfig() 里补充，
并在 index.js:118 用这个新字段替换 cfg.agentId。

其余部分做得不错：gateway() 的超时（AbortController）+ 重试 + v3
信封校验实现正确，captureSession() 读取失败时跳过而不写占位数据，
turnKey() 去重 + capturedTurnsCap 容量上限实现也没问题，测试套件
覆盖了去重/重试/信封校验失败/读取失败跳过等场景。但上面这个 header
语义混用的问题需要先修复才能合并。

YOMXXX:
## Review: OpenCode `session.idle` capture needs two fixes before merge

Thanks for addressing the previous `x-tdai-service-id` vs `agent_id` issue — the separate `serviceId` config looks correct now.

I found two remaining compatibility problems in the capture path.

### 1. Blocking: each new idle event can re-write old transcript messages

`captureSession()` builds `transcriptFromMessages(messages).slice(-cfg.maxTranscriptMessages)` and posts that whole rolling window to `/v3/conversation/add`. The dedupe key is based only on the latest OpenCode message id, so it prevents the *same idle event* from replaying, but it does not prevent older messages from being re-posted when a later assistant message becomes the new latest id.

Example:

1. idle after turn 1 posts `[user1, assistant1]`
2. idle after turn 2 posts `[user1, assistant1, user2, assistant2]`

That duplicates turn 1 in L0 and can pollute downstream distillation/search.

Suggested fix:

- Capture only the latest completed user/assistant turn, not the whole session window.
- Pair the final assistant message with its `parentID` user message when available; otherwise fall back to the nearest previous user message.
- Use the event `sessionID` as the persisted `session_id`.
- Add a regression test: two sequential idle events over a growing message list should produce exactly two L0 writes, each containing only one user+assistant turn.

Pseudo-shape:

```js
function latestCompletedTurn(sessionID, messages) {
  const assistantIndex = findLastIndex(messages, m => m?.info?.role === "assistant" && !m?.info?.error);
  const assistant = messages[assistantIndex];
  const user = findParentUser(messages, assistant?.info?.parentID) ?? findPreviousUser(messages, assistantIndex);
  if (!user || !assistant) return null;
  return {
    key: `${sessionID}:${assistant.info.id ?? hash(userText + assistantText)}`,
    messages: [toGatewayMessage(user), toGatewayMessage(assistant)],
  };
}
```

### 2. Important: Gateway failures should fail open from the OpenCode event hook

`captureSession()` catches transcript-read failures, but the later `await gateway(.../v3/conversation/add...)` is not caught inside the event handler. If the Memory Gateway is down, times out, or returns a non-zero v3 envelope, the plugin event hook rejects.

Suggested fix:

```js
if (event.type === "session.idle" && event.properties?.sessionID) {
  try {
    await captureSession(ctx, cfg, event.properties.sessionID);
  } catch (error) {
    console.warn(`[opencode-memory] capture failed open: ${error?.message ?? error}`);
  }
}
```

And add a test where `/v3/conversation/add` returns 500/non-zero envelope; the event handler should resolve and OpenCode should continue normally.

### Follow-up after #1142 lands

Once the server-side idempotency contract from #1142 is available, the adapter should also send a deterministic `idempotency_key` for each completed turn. That will close the remaining cross-restart duplicate window cleanly.

Maxwell-Code07:
感谢您的贡献。针对 Codex 适配器，我们已完成相关适配工作，因此该 PR 现阶段暂不予合并。非常感谢您的投入与付出，敬请知悉。

PR #963, opened 2026-08-12
fix(proxy): strip session-init form artifacts before forwarding (400 in extended-thinking mode)

hongchuan666:
## Problem

When `sessionInit` is enabled and the upstream LLM runs in **extended-thinking mode** (e.g. DeepSeek's Anthropic-compatible endpoint), the first real request of every fresh session fails with:

```
API Error: 400 The `content[].thinking` in the thinking mode must be passed back to the API.
```

## Root cause

1. `sessionInit` synthesizes an `AskUserQuestion` tool_use (id prefix `toolu_cc_session_init_`) as a fake SSE response to Claude Code — with `stop_reason: tool_use` and **no `thinking` block**.
2. The client echoes that tool_use back with a matching `tool_result`.
3. Upstream providers in extended-thinking mode require every assistant tool_use to carry a **real, server-signed `thinking` block**. The synthetic form messages carry no valid signature, so the upstream rejects the whole request with 400.
4. These synthetic messages were forwarded to the upstream untouched (the previous `keepInitArtifacts` behavior), so the 400 occurred deterministically on the first request of every session.
5. Separately, when a team has **no real task**, session-init silently bypassed (team+agent+task contract), so a user who answered "yes, bind team assets" got no binding and no injection.

## Fix

- **`stripSessionInitArtifacts()`** in `src/session/claude-code/form.ts`: removes synthetic session-init messages before forwarding upstream:
  - assistant message containing a session-init tool_use → **dropped**
  - user message's `tool_result` referencing a session-init tool_use id → **block removed** (real text blocks preserved)
  - everything else passes through **unchanged** (identity for non-artifact input)
- Applied both in the session-init message pipeline (`init.ts`) **and** at the forwarding boundary (`anthropicHandler.ts`), so bypass / non-intercepted / fork / sidequery paths are covered too.
- **`sessionInit.defaultTaskId`** config: when a team has no real task, session-init can register with a virtual default task instead of silently bypassing — users can still bind and receive agent-context injection.

## Verification

- New unit tests for `stripSessionInitArtifacts` (6 cases) — all pass.
- E2E: fresh session → asset-confirm form → "yes" → auto-select team/agent → task → upstream **200** with injection active (system input tokens ~170 → ~3080).
- Real conversation messages and client-side history are untouched; only the upstream request body is stripped.

## Notes

- The pre-existing `sanitizeThinkingBlocks` logic in `anthropicHandler.ts` is separate and unchanged.

### DIFF
--- a/MemoryProxy/FIX-NOTES.md
+++ b/MemoryProxy/FIX-NOTES.md
@@ -1,0 +1,47 @@
+# Fix Notes — session-init form artifacts 400 (DeepSeek thinking)
+
+**Status**: 已修复并验证（2026-08-12），改动已 `docker cp` 进运行中 `tdai-proxy` 容器。
+
+## 问题
+
+代理 sessionInit 在首次请求时伪造 `AskUserQuestion`（id `toolu_cc_session_init_*`）消息，用户回答后这段
+**无 thinking 签名的合成消息被原样转发给上游**（DeepSeek thinking 模式）。DeepSeek 校验 assistant 消息
+thinking 签名真实性，合成消息无签名 → **400**：`The content[].thinking in the thinking mode must be passed back to the API.`
+
+触发窗口：仅首次请求（回答后下一条真实回复带有效签名压上去，历史合成消息不再触发校验）。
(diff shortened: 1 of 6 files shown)


Maxwell-Code07:

PR #966, opened 2026-08-12
fix: unbound variable crash in start-memory-core.sh admin-init log line

### LINKED ISSUE #965: start-memory-core.sh aborts on macOS: full-width ） after a variable reference is parsed as part of the name
### Summary

`deploy/global-images/start-memory-core.sh` fails on macOS at line 175 with:

```
start-memory-core.sh: line 175: ADMIN_KEY_FILE?: unbound variable
```

This happens *after* memory-core is up and healthy, so the container is left
running with no admin user created — `start-all.sh` never reaches memory-hub or
proxy.

### Cause

Line 175 is:

```bash
info "初始化 admin user（username=${MEMORY_CORE_ADMIN_USERNAME}, key 持久化 → $ADMIN_KEY_FILE）..."
```

`$ADMIN_KEY_FILE` is immediately followed by `）` (U+FF09, bytes `EF BC 89`).
macOS's character tables classify that as an identifier character, so bash folds
its first byte into the variable name, giving an undefined `ADMIN_KEY_FILE\xEF`
and aborting under `set -u`. glibc classifies it as punctuation, which is why
Linux is unaffected.

It is the locale, not the bash version — 3.2 and 5.3 fail identically:

```bash
for L in C en_US.UTF-8 en_IN.UTF-8; do
  printf '%-14s ' "$L"
  LC_ALL=$L bash -c 'set -u; V=works; echo "→ $V）"' 2>&1 | tail -1
done
# C            → works）
# en_US.UTF-8  bash: V?: unbound variable
# en_IN.UTF-8  bash: V?: unbound variable
```

### Suggested fix

Brace the references wherever CJK punctuation follows one — `${ADMIN_KEY_FILE}）`
— which is unambiguous in every locale. A blanket `export LC_ALL=C` at the top
of `_lib.sh` also works and is what we did locally, but bracing is the more
targeted fix.

Other scripts in `deploy/global-images/` use the same pattern and will hit this
as soon as execution reaches them.

### Related

`verify.sh` reports success with an unfunded API key, because its
OpenAI-protocol check is `GET /models`, which does not require credit. The first
real request then fails with `402`. A minimal chat completion, or a note in the
output that `/models` does not prove billing, would catch this at install time.

fissible:
## Description | 描述
`deploy/global-images/start-memory-core.sh` crashes on `./start-memory-core.sh` (and therefore `./start-all.sh`) during admin-user initialization, right after `memory-core` reports healthy.

Root cause: line 175 has an unbraced `$ADMIN_KEY_FILE` immediately followed by a full-width Chinese punctuation character (`）`) with no ASCII separator:

```bash
info "初始化 admin user（username=${MEMORY_CORE_ADMIN_USERNAME}, key 持久化 → $ADMIN_KEY_FILE）..."
```

Under a UTF-8 locale (the macOS default — reproduced with `en_US.UTF-8` on both bash 3.2 and bash 5.3), bash's tokenizer can absorb part of that multi-byte character into the variable name, so `set -u` fails with `ADMIN_KEY_FILE?: unbound variable` even though `ADMIN_KEY_FILE` is defined earlier in the script (line 29). It only avoids the bug under `LC_ALL=C`, which effectively no user has set by default. This means the script fails for essentially any user running it out of the box on macOS.

Fix: brace the reference (`${ADMIN_KEY_FILE}`) so bash's parser gets an unambiguous delimiter. Swept the rest of the repo's shell scripts for the same unbraced-`$VAR`-immediately-before-non-ASCII pattern; this was the only occurrence.

## Related Issue | 关联 Issue
Fixes #965

## Change Type | 修改类型
- [x] Bug fix | Bug 修复

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过 — reproduced the crash in isolation on bash 3.2 and bash 5.3 under `en_US.UTF-8`, confirmed the braced form resolves it, and re-ran `./start-memory-core.sh` end-to-end successfully after the fix.
- [x] No existing features affected | 无影响现有功能 — one-line change, log message text/formatting unchanged.

## Additional Notes | 其他说明

### DIFF
--- a/deploy/global-images/start-memory-core.sh
+++ b/deploy/global-images/start-memory-core.sh
@@ -172,7 +172,7 @@ verify_user_key() {
   [[ "$code" == "200" ]]
 }
 
-info "初始化 admin user（username=${MEMORY_CORE_ADMIN_USERNAME}, key 持久化 → $ADMIN_KEY_FILE）..."
+info "初始化 admin user（username=${MEMORY_CORE_ADMIN_USERNAME}, key 持久化 → ${ADMIN_KEY_FILE}）..."
 
 # 生成随机 key（首次 init-admin 用；若之前有 file 就复用）
 if [[ -s "$ADMIN_KEY_FILE" ]]; then


Maxwell-Code07:

PR #977, opened 2026-08-13
test(gateway): regression coverage for conversation delete isolation (#871)

YuhaoLin2005:
## Problem

`POST /v2/conversation/delete` silently no-ops for `message_ids`-only requests (no `session_id` in body/headers). Root cause: `handleConversationDelete` passes the full isolation ctx (`deps.requestIsolation`) into `deleteL0` / `deleteL0BySession`. `resolveIsolation` fills a missing `session_id` with `DEFAULT_ISOLATION_ID` (`"default"`), and `rowMatchesIsolation` exact-matches `session_id` — so a record stored under a real session never matches `"default"`, `deleteL0` returns `false`, and the endpoint reports `deleted_count: 0` (HTTP 200). This is #871.

Every other conversation handler already avoids this:
- `handleConversationQuery` / `handleConversationCount` take `session_id` from the request body and only use `teamId` / `userId` / `agentId` / `taskId` from the isolation ctx (never `iso.sessionId`)
- `handleConversationSearch` explicitly strips `sessionId` from the filter (`v2-router.ts:893-896`, "避免默认 sessionId='default' 错误过滤真实 session 数据")
- `handleAtomicDelete` (L1) builds its delete filter without `sessionId` (`v2-router.ts:1247-1253`, "按 id 删除不应被默认 sessionId 限制")

## Change

`handleConversationDelete` now builds the same delete filter as `handleAtomicDelete` — explicit isolation dims only, no `sessionId` — and passes it to `deleteL0` / `deleteL0BySession`.

The behavior change is confined to the `message_ids`-only path (broken → works). The `session_id` path already worked (the body value populated `iso.sessionId` and matched); it is unaffected, since `deleteL0BySession` is already scoped to that session and tenant dims (`teamId`/`userId`/`agentId`/`taskId`) remain enforced in both paths.

Complementary to #872 (fail loudly when nothing was removed): this makes legitimate deletes actually work, and fail-loudly remains the correct fallback for records that genuinely don't exist.

## Verification

Diff is 14+/3- in a single handler, mirroring the existing `handleAtomicDelete` pattern 300 lines away in the same file.

Adds a regression test (`src/gateway/v2-router.test.ts`) that drives the real dispatch path (`handleV2Route` → `resolveIsolation` → `handleConversationDelete`) and asserts a `message_ids`-only delete removes a record stored under a real session. Verified red before the fix (`deleted_count: 0`) and green after (`deleted_count: 1`); also covers the `session_id` path and cross-user isolation.

### DIFF
--- a/MemoryCore/src/gateway/v2-router.test.ts
+++ b/MemoryCore/src/gateway/v2-router.test.ts
@@ -1,0 +1,153 @@
+import { describe, expect, it, vi } from "vitest";
+import type http from "node:http";
+
+import type { IsolationFilter, IMemoryStore } from "../core/store/types.js";
+import type { Logger } from "../types.js";
+import { handleV2Route, type V2RouterDeps } from "./v2-router.js";
+
+// ── Test helpers ──
+
+const logger = {
+  debug: vi.fn(),
+  info: vi.fn(),
+  warn: vi.fn(),
+  error: vi.fn(),
+} as unknown as Logger;
+
+function makeReq(headers: Record<string, string>): http.IncomingMessage {
+  return { headers } as unknown as http.IncomingMessage;
+}
+
+function makeSend(): {
+  sendJson: (res: http.ServerResponse, status: number, body: unknown) => void;
+  last: () => { status: number; body: any };
+} {
+  const state = { status: 0, body: undefined as unknown };
+  return {
+    sendJson: (_res: http.ServerResponse, status: number, body: unknown) => {
+      state.status = status;
+      state.body = body;
+    },
+    last: () => state as { status: number; body: any },
+  };
+}
+
+interface StoredRecord {
+  recordId: string;
+  sessionId: string;
+  userId: string;
+  agentId: string;
+  teamId?: string;
+}
+
+/**
+ * Mock IMemoryStore whose deleteL0/deleteL0BySession emulate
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:

YuhaoLin2005:
Hi @Maxwell-Code07 — heads-up on the pivot before your internal review.

The base branch feat/server_team now includes commit 6382e8a ("feat: release v2.0.1-beta.1", 2026-08-14), which rewrites handleConversationDelete with the same semantics this PR proposed: excluding requestIsolation.sessionId from the delete filter and routing session deletes through deleteL0BySession. The functional fix is now covered upstream, so I've pivoted this PR to tests-only — the remaining value is regression coverage.

Verified against the fresh base (vitest run):
1. message_ids-only delete removes a record stored under a real session
2. session_id path still works with tenant scoping
3. cross-user isolation is preserved

All 3 pass on the fresh base. On the pre-fix base (before 6382e8a) case 1 fails — deleted_count 0 instead of 1 — the "default" sessionId placeholder silently filtered the record out, which is exactly the #871 behavior this suite guards. upstream had no v2-router.test.ts, so the tests fill a gap.

Happy to fold them into your own test suite if you prefer. No action needed from you until then.

PR #988, opened 2026-08-14
fix(proxy): finalize OpenAI streams on DONE

stoner-byte:
## Description | 描述

OpenAI-compatible clients may stop reading and cancel the response as soon as they receive the terminal `data: [DONE]` line. The proxy previously finalized streaming side effects only when the SSE event delimiter or transport EOF was observed, so cancellation could skip usage logging, Chat Memory L0 writeback, and skill extraction.

This PR:

- recognizes the terminal sentinel as soon as a complete SSE `data:` line arrives, without waiting for the following blank line or connection EOF;
- accepts LF and CRLF framing, with or without the optional space after `data:`;
- keeps response bytes unchanged by normalizing line endings only in the internal parsing buffer;
- uses a shared finalization promise so `[DONE]` and later EOF/flush paths remain idempotent;
- adds real HTTP regression coverage for immediate cancellation, a delayed blank-line terminator, CRLF framing, normal `[DONE]` plus EOF, and EOF without `[DONE]`.

## Related Issue | 关联 Issue

N/A — reproduced during Hermes/TencentDB Agent Memory integration testing.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

`cd MemoryProxy && npm test` passes 5/5 tests. `git diff --check` is clean. The commit includes the required DCO sign-off.

### DIFF
--- a/MemoryProxy/src/__tests__/handler-stream-finalization.test.ts
+++ b/MemoryProxy/src/__tests__/handler-stream-finalization.test.ts
@@ -1,0 +1,203 @@
+import { createServer, type Server } from "node:http";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { DEFAULT_CONFIG } from "../config.js";
+import { createApp } from "../server.js";
+import type { ProxyConfig } from "../types.js";
+
+const servers: Server[] = [];
+
+afterEach(async () => {
+  vi.restoreAllMocks();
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #989, opened 2026-08-14
fix(proxy): preserve space id during hook prewarm

stoner-byte:
## Description | 描述

First-turn Hook Cache prewarming did not receive the request's resolved `spaceId`. Prewarm writes therefore used a fallback namespace while the live injector lookup used the request namespace, producing cache misses, unnecessary self-healing, and, with the SQLite repository, foreign-key failures.

This PR:

- passes the resolved request `spaceId` into `prewarmFromConfig()` in both the OpenAI and Anthropic handlers;
- keeps cache prewarm and live lookup on the same `(spaceId, userId, agentSource, sessionId)` key;
- adds handler-level regression tests using real HTTP request paths for both protocols;
- verifies the prewarmed entry is hit, injection succeeds, and no self-heal write occurs.

## Related Issue | 关联 Issue

N/A — reproduced from first-turn MemoryProxy runtime logs during integration testing.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

`cd MemoryProxy && npm test` passes 2/2 tests. `git diff --check` is clean. The commit includes the required DCO sign-off.

### DIFF
--- a/MemoryProxy/src/__tests__/handler-prewarm-space-id.test.ts
+++ b/MemoryProxy/src/__tests__/handler-prewarm-space-id.test.ts
@@ -1,0 +1,335 @@
+import { createServer, type Server } from "node:http";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { initAuth } from "../auth.js";
+import { DEFAULT_CONFIG } from "../config.js";
+import {
+  __resetHookCacheRepoForTests,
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

PR #991, opened 2026-08-14
feat(recall): add task-aware LLM memory selection

ccdf1137:
## Description | 描述

Add an optional task-aware LLM selection stage after retrieval/RRF for automatic memory recall.

The selector receives an expanded set of retrieved candidates and returns memory IDs only. Its output is strictly validated: duplicate, unknown, invented, or malformed IDs cause the recall path to fall back to the original retrieval ranking. Manual `tdai_memory_search` behavior is unchanged, and the feature is disabled by default.

This change also:

- reuses the existing `LLMRunnerFactory` and supports an optional selector model override;
- exposes candidate multiplier and timeout configuration with bounded candidate expansion;
- preserves original record IDs and memory text without rewriting;
- documents configuration and behavior in English and Chinese;
- adds unit tests, auto-recall integration tests, and a reproducible real-model evaluation harness.

## Related Issue | 关联 Issue

N/A

## Change Type | 修改类型

- [ ] Bug fix | Bug 修复
- [x] New feature | 新功能

## Self-test Checklist | 自测清单

## Validation | 验证

- `npm test` — 7 test files, 82 tests passed
- `npm run build` — passed for the Node.js 22.16 target
- dry-run task-selector evaluation — passed
- real-model evaluation with `deepseek-v4-flash`, thinking disabled, 10 runs across 6 synthetic cases:
  - best run: selector nDCG@K 0.956 vs RRF baseline 0.107
  - successful selections achieved nDCG@K 1.0 in the fixture set
  - malformed or empty model output safely fell back to RRF with no critical-recall regression

The real-model evaluation is intentionally presented as a small synthetic regression benchmark, not a production-quality claim. Across repeated comparable runs, the model endpoint occasionally returned empty or truncated JSON; the fail-open behavior handled those cases as designed.

## Additional Notes | 其他说明

The selector is opt-in (`recall.taskSelector.enabled: false` by default). Model latency and structured-output reliability vary by provider, so timeout and model selection remain configurable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,7 +479,7 @@ TDAI_EVAL_RUNS=3 \
 npm run eval:task-selector
 ```
 
-The evaluator compares the selector with the original retrieval Top-K using Precision@K, Recall@K, nDCG@K, critical-memory recall, fallback rate, latency, and estimated token usage. Set `TDAI_EVAL_OUTPUT=benchmark-runs/task-selector.json` to save a report. The included cases are examples; use 30–50 representative, manually labelled cases before making a production decision. API credentials are read from environment variables only, and real-model evaluation is not part of normal CI.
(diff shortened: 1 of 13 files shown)


Maxwell-Code07:

PR #1007, opened 2026-08-15
fix(proxy): write TDAI L0 only for final answers

stoner-byte:
## Description | 描述

MemoryProxy currently writes a TDAI L0 turn after every upstream model response. A single agent turn that calls tools contains multiple upstream requests, so intermediate tool-call responses are recorded as separate L0 turns and the original user message is duplicated.

This change reuses the existing round-finalization predicate used by Skill extraction and writes L0 only when the upstream response is the final assistant answer. Both streaming and non-streaming paths are covered.

Result:

- L0 records one pair: the original user message and the final assistant answer.
- Intermediate tool calls and tool results are not written to L0.
- Skill extraction keeps its existing complete-round payload.
- Network retry idempotency remains out of scope.

## Related Issue | 关联 Issue

N/A

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Validation:

- `npm test -- --run src/__tests__/handler-l0-round-finalization.test.ts` (2 tests passed)
- Full related suite: 5 test files and 13 tests passed
- Verified with Hermes on a macOS deployment: one L0 write with `msgs=2`, while Skill extraction submitted the complete tool round once

### DIFF
--- a/MemoryProxy/src/__tests__/handler-l0-round-finalization.test.ts
+++ b/MemoryProxy/src/__tests__/handler-l0-round-finalization.test.ts
@@ -1,0 +1,309 @@
+import { mkdtempSync, rmSync } from "node:fs";
+import { createServer, type IncomingMessage, type Server, type ServerResponse } from "node:http";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1011, opened 2026-08-15
feat(memory-core): implement GitStorageBackend (experimental)

life-is-blue:
## Summary

Implements `GitStorageBackend` for `IStorageBackend`, per the RFC in #894 (`docs/rfc/git-storage-backend.md`). Experimental / single-writer-per-space pilot scope, per the RFC's own recommendation — not a default, not for high-frequency multi-writer online storage.

- **Composes, doesn't reimplement**: wraps an inner `LocalStorageBackend` for pure file I/O (`getObject`/`exists`/`listObjects`, and the file-write half of put/append/delete), layering a WAL + `SerialQueue`-serialized batched commit/push on the mutating path only.
- One independent `git clone --single-branch --no-tags` per memory space; branch names are a 128-bit hash of a length-prefixed `(tenantId, instanceId)` encoding, validated via `check-ref-format`.
- `appendObject`'s idempotency key is a backend-generated operation UUID recorded in each commit's message trailer — not a content hash, which would silently drop legitimately-duplicate content on replay.
- Single-writer lock via injected `IStateBackend` (dependency-injected, never imports `src/core/state/*` directly — respects the CI isolation red line) + in-process `SerialQueue`; git's own non-fast-forward push rejection remains the real correctness backstop.
- Shells out to the system `git` binary (`execFile`, argv arrays, never a shell string) — zero new npm dependencies.
- `server.ts` had three near-identical inline blocks deciding local-vs-COS storage; extracted into `gateway/storage-resolver.ts`'s `resolveFileStorageBackend()`, which both call sites now share, plus the new `"git"` branch. Existing local/COS behavior for deployments not using git storage is preserved (see commit `fix(memory-core): address Codex checkpoint B findings` for two real regressions caught and fixed during review).
- First-ever storage-layer test coverage in this repo: a shared `IStorageBackend` contract suite (7 methods, path traversal, pagination) runs against both `LocalStorageBackend` and `GitStorageBackend` (using a local `git init --bare` fixture, no network); git-specific tests cover crash recovery, push-rejected replay with no duplication across two uncoordinated writers, and 8 adversarial branch-name encodings. New CI `test:` job so these actually gate PRs going forward.
- **Security-sensitive**: this touches git credential handling (HTTP token / SSH key). Per `CONTRIBUTING.md`, flagging for security-focused review attention.

**Known CI gap** (not introduced by this PR, discovered while adding the `test:` job): `pr-ci.yml` triggers only on `pull_request.branches: [main]`, but this PR (like #894) targets `feat/server_team` — so none of `pr-ci.yml`'s jobs, including the new `test:` job, will actually run against this PR. Flagging for maintainers to decide whether to widen the trigger; left untouched here since it's a repo-wide CI policy decision beyond this PR's scope.

**Explicitly out of scope**: the RFC §8 migration tool (future work), Option B (per-object-version git plumbing), fixing `StorageAdapter.rename`'s pre-existing non-atomicity.

Design doc with full rationale and two rounds of Codex cross-review (design-only, then diff review) at `MemoryCore/docs/design/git-storage-backend-implementation-plan.md`.

## Test plan

- [x] `cd MemoryCore && npm test` — 101 tests pass (contract suite × local + git, crash recovery, replay dedup, branch-name edge cases, config parsing)
- [x] `tsc --strict --noEmit` against the full source tree: zero new errors vs. the pre-existing 325-error baseline (all pre-existing, unrelated to this change — e.g. optional COS/Redis modules absent from this checkout)
- [x] Manual: constructed a real `GitStorageBackend` against a `git init --bare` fixture, ran put/append/delete, inspected commit history — only the 6 RFC-declared key prefixes present, no `.meta.json`, no credentials
- [x] Manual: killed a flush mid-push, restarted with a fresh instance pointed at the same local clone, confirmed recovery resumed and completed the push without loss or duplication
- [x] Manual: two uncoordinated writers (separate local clones, separate lock backends) concurrently appending to the same key — non-fast-forward rejection correctly triggered replay, final content had both writers' lines with no duplication
- [x] `npm pack --dry-run` — no unexpected dependency/size changes (this PR adds zero new npm dependencies)

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -67,6 +67,27 @@ jobs:
           path: "MemoryCore/*.tgz"
           retention-days: 7
 
+  # ── 3. Unit tests ───────────────────────────────────────────
+  test:
+    name: Unit Tests
+    needs: install
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+
+      - name: Restore node_modules
+        uses: actions/cache@v4
+        with:
+          path: MemoryCore/node_modules
+          key: deps-${{ runner.os }}-${{ hashFiles('MemoryCore/package.json') }}
+
+      - name: npm test
+        run: npm test
+
   # ── 5. Plugin manifest validation ──────────────────────────
   manifest:
     name: Manifest
--- a/MemoryCore/docs/design/git-storage-backend-implementation-plan.md
+++ b/MemoryCore/docs/design/git-storage-backend-implementation-plan.md
@@ -19,8 +19,8 @@ RFC 已核实：文中每处代码引用都逐字节对照当前源码验证过�
 其余关键决定：
 - **Shell 出系统 `git` 二进制**（`execFile`，参数用数组不用字符串，杜绝 shell 注入），不加 `simple-git`/`isomorphic-git` 依赖——本仓库已有的风险姿态是"外部依赖是可以接受的"（COS 需要网络+密钥），这里只是换成"需要 PATH 上有 git"，且免去 COS 那套动态 import + optionalDependency 的仪式。
 - **`factory.ts` 静态 import** `GitStorageBackend`（不是 COS 式动态 import）——没有重量级可选依赖需要延迟加载。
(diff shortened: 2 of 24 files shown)


Maxwell-Code07:

Issue #990, opened 2026-08-14
[BUG] Claude‑Code proxy upstream DeepSeek‑V4‑Flash hit 400 The content[].thinking in the thinking mode must be passed back to the API after tool‑call turn


ryan-yuan-dev:
**Claude Code Version:**
2.1.232

**Plugin Version:**
2.0.0

**Operating System:**
MacOS 26.4 (25E246) 

**Describe the bug:**
1. Start full memory‑proxy stack (./start‑all.sh)
2. Configure upstream LLM provider to deepseek‑v4‑flash
    ```shell
    # ── memory 组：memory + memory-hub 内部调用的 LLM ────────────────────────────
    # 用途：memory 的 embed/summarize；memory-hub 里 knowledge 的 wiki ingest / 总结
    MEMORY_LLM_BASE_URL=https://api.deepseek.com/anthropic             # 例：https://api.deepseek.com/v1
    MEMORY_LLM_API_KEY=sk-xxxx               # 例：sk-xxxxxxxx
    MEMORY_LLM_MODEL="deepseek-v4-flash[1m]"                 # 例：deepseek-chat
    MEMORY_LLM_PROTOCOL=anthropic                  # openai | anthropic

    # ── proxy 组：proxy 把用户请求转发到的上游 LLM ──────────────────────────────
    # 用途：coding agent / 客户端调 proxy:8096，proxy 再转发到这里
    PROXY_UPSTREAM_URL=https://api.deepseek.com/anthropic               # 例：https://api.deepseek.com/v1
    PROXY_UPSTREAM_API_KEY=sk-xxxx           # 例：sk-xxxxxxxx（可与 memory 组不同）
    PROXY_UPSTREAM_MODEL="deepseek-v4-flash[1m]"             # 例：deepseek-chat
    ```
3. Configure Claude Code auth info
    ```shell
    export ANTHROPIC_BASE_URL=http://127.0.0.1:8096/claude-code/default
    export ANTHROPIC_AUTH_TOKEN='sk-mem-xxxx'
   claude --model deepseek-v4-flash[1m]
    ```
4. Launch claude cli then send "hello" message
5. Choose "关联团队资产" -> API Error: 400 The `content[].thinking` in the thinking mode must be passed back to the API.

**Expected behavior:**
Avoid 400 error

**Error Logs / Screenshots :**

- ScrreenShot
[...]

PR #1024, opened 2026-08-16
fix(proxy): accept SSE data fields without the optional space

pucedoteth:
## Problem

`createSseThinkingFixStream` in `anthropicHandler.ts` found the data line with:

```ts
if (line.startsWith("data: ")) dataLine = line;
...
if (!dataLine) { controller.enqueue(part + "\n\n"); continue; }
```

Per the SSE spec the space after the colon is optional framing, so `data:{...}` is a well-formed frame. It was not matched, and the stream enqueued the frame **unchanged**:

```
old detection -> ""  (SKIPPED - frame passes through with thinking:null)
new detection -> (patched)
```

That gap lands exactly where it hurts. This stream exists as a workaround for upstreams that emit `thinking: null` — its own comment says Claude Code then does `contentBlock.thinking += delta.thinking` and produces `"null"` strings. Those non-Anthropic upstreams (cf. #990, DeepSeek) are the ones least likely to follow the space convention, so the frames most in need of patching are the ones most likely to be skipped.

The codebase already knows about this: `codexHandler.ts` and `systemUserPassthrough.ts` both check `data: ` **and** `data:`. `anthropicHandler.ts` is internally inconsistent — line 2042 handles both forms while the two sites fixed here did not.

## Change

- Add `sseDataValue(line)`, which strips at most one framing space and returns `undefined` for non-data lines, so "no data field" stays distinguishable from "empty data field".
- Use it in the thinking-fix stream, in the tail-buffer usage parse, and in `workbuddyHandler`.
- The patched-frame rebuild matched `startsWith("data: ")` to substitute the new payload; it now targets the index of the line actually parsed, so a frame with more than one data line can't have them all overwritten with the same value.

## Testing

`MemoryProxy/src/__tests__/sseDataValue.test.ts` — 7 cases: both framing forms, single-space stripping, empty data field, non-data lines, and `database:` not being read as `data`. `npx vitest run` in `MemoryProxy` passes.

`npx tsc --noEmit` reports 6 errors in `anthropicHandler.ts:567`, `codexHandler.ts`, `config.ts` and `storage/factory.ts` — all present on an unmodified checkout, none from this change.

## Note

While looking for this I tried to reproduce #965 (`start-memory-core.sh` aborting on macOS on a full-width `）` after a variable). It does not reproduce for me: bash 3.2.57 and zsh both expand `$ADMIN_KEY_FILE` correctly there and exit 0, and `bash -n` is clean across every script in the repo. That issue may need a different diagnosis.

### DIFF
--- a/MemoryProxy/src/__tests__/sseDataValue.test.ts
+++ b/MemoryProxy/src/__tests__/sseDataValue.test.ts
@@ -1,0 +1,41 @@
+import { describe, expect, it } from "vitest";
+import { sseDataValue } from "../anthropicHandler.js";
+
+describe("sseDataValue", () => {
+  it("reads a data field written with the conventional space", () => {
+    expect(sseDataValue('data: {"type":"ping"}')).toBe('{"type":"ping"}');
+  });
+
+  // Previously the SSE parsers matched only `"data: "`, so a conforming
+  // upstream that omitted the optional space had its frames skipped — and the
+  // thinking-fix stream exists precisely for upstreams that do not conform.
+  it("reads a data field written without the optional space", () => {
+    expect(sseDataValue('data:{"type":"ping"}')).toBe('{"type":"ping"}');
+  });
+
+  it("strips only the single framing space", () => {
+    expect(sseDataValue("data:  padded")).toBe(" padded");
+  });
+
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

PR #1037, opened 2026-08-17
fix(sdk/python): annotate context manager exit arguments with object

aoright:
### Summary
- Update `__exit__` and `__aexit__` in `SkillClient` and `AsyncSkillClient` to annotate `*exc` with `object` instead of `Any`, following Python typing standards (PYI036).
- Ensures full compatibility with strict type checkers (MyPy, Pyright, Ruff).


Maxwell-Code07:

PR #1040, opened 2026-08-17
fix(queue): resolve idle waiters when clearing queued tasks

### LINKED ISSUE #1039: [Bug] SerialQueue.clear() leaves onIdle() pending when a paused queue is cleared

N/A — reproduced directly against the queue utility.

### Plugin Version | 插件版本

Commit `97f94654280b2932c35ba4806a491999ed244cc9` (`MemoryCore` reports `2.0.0-beta.1`).

### Operating System | 操作系统

macOS 26.5.1, Node.js v22.22.2.

### Describe the bug | 问题描述

Both `MemoryCore/src/utils/serial-queue.ts` and `MemoryKnowledge/src/store/serial-queue.ts` keep `onIdle()` resolvers in `idleResolvers`. When a paused queue contains pending work, `clear()` rejects and removes that work but never resolves those waiters. The queue reports `idle === true` while an already-returned `onIdle()` promise remains pending forever.

### To Reproduce | 复现步骤

```sh
node --experimental-strip-types --input-type=module - <<'JS'
import { SerialQueue } from './MemoryKnowledge/src/store/serial-queue.ts';
const queue = new SerialQueue('repro');
queue.pause();
const task = queue.add(async () => 'never-runs').catch(() => 'cleared');
const idle = queue.onIdle().then(() => 'idle');
queue.clear();
const result = await Promise.race([
  idle,
  new Promise(resolve => setTimeout(() => resolve('timeout'), 50)),
]);
console.log(JSON.stringify({ task: await task, onIdle: result, idleState: queue.idle }));
JS
```

Actual output:

```json
{"task":"cleared","onIdle":"timeout","idleState":true}
```

### Expected behavior | 预期行为

Once `clear()` removes the last pending task and no task is running, existing `onIdle()` waiters should resolve.

### Proposed fix | 建议修复

Drain `idleResolvers` in `clear()` only when no task is currently running. If a task is running, its existing `finally` path will notify the waiters. The same focused change and regression test should be applied to both implementations.

I have reproduced the bug against the current source and can submit the fix after maintainers confirm the appropriate PR target branch.

wutongyuonce:
## Description | 描述

Resolve existing `onIdle()` waiters when `clear()` makes a non-running `SerialQueue` idle. The same bug and focused regression test are addressed in both the MemoryCore and MemoryKnowledge implementations.

If a task is already running, `clear()` leaves notification to the existing task completion path, so waiters are not resolved early.

## Related Issue | 关联 Issue

Fixes #1039

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Focused verification:

- Both new Vitest regression tests pass.
- Both changed `SerialQueue` source files pass TypeScript checking.
- The source-level reproduction now returns `{"task":"cleared","onIdle":"idle","idleState":true}` instead of timing out.

AI assistance was used for investigation, implementation, and test drafting; the reproduction and tests were run locally.

### DIFF
--- a/MemoryCore/src/utils/serial-queue.test.ts
+++ b/MemoryCore/src/utils/serial-queue.test.ts
@@ -1,0 +1,24 @@
+import { describe, expect, it } from "vitest";
+
+import { SerialQueue } from "./serial-queue.js";
+
+describe("SerialQueue", () => {
+  it("resolves existing idle waiters when a paused queue is cleared", async () => {
+    const queue = new SerialQueue("test");
+    queue.pause();
+
+    const task = queue.add(async () => "never runs");
+    const rejection = expect(task).rejects.toThrow("Queue cleared");
+    let idle = false;
+    void queue.onIdle().then(() => {
+      idle = true;
+    });
+
+    queue.clear();
+    await rejection;
+    await Promise.resolve();
+
+    expect(queue.idle).toBe(true);
+    expect(idle).toBe(true);
+  });
+});
--- a/MemoryCore/src/utils/serial-queue.ts
+++ b/MemoryCore/src/utils/serial-queue.ts
@@ -95,6 +95,11 @@ export class SerialQueue {
       entry.reject(new Error("Queue cleared"));
     }
     this.queue = [];
+    if (!this.running) {
+      const resolvers = this.idleResolvers;
+      this.idleResolvers = [];
+      for (const resolve of resolvers) resolve();
+    }
   }
 
   private drain(): void {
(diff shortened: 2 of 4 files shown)


Maxwell-Code07:

wutongyuonce:
@Maxwell-Code07 Thanks for picking this up. I reconfirmed the hang against current `feat/server_team`: after `pause()` + `add()` + `onIdle()` + `clear()`, `idle` is already `true` but the waiter never resolves. This PR only drains `idleResolvers` in `clear()` when nothing is running, so an in-flight task still notifies through its existing `finally` path. Could you review when you have a chance?

PR #1043, opened 2026-08-17
feat: add native DSH TencentDB memory plugin

Vocllum:
## What changed

- add a native DSH bundle with automatic recall, exactly-once L0 capture, asynchronous Skill conversation delivery, and focused read-only search tools
- keep L1/L2/L3 and Skill extraction backend-owned asynchronous MemoryCore work
- document the reduced integration boundary and fail-open/security behavior

## Validation

- `node --check` for plugin modules
- DSH `--dump-config` and `--help` load checks
- mocked Gateway client contract check
- removed redundant Tailscale 8445 mapping; DSH remains on trusted 8443

The initial push to the upstream repository was read-only, so this PR is submitted from the Vocllum fork.

### DIFF
--- a/MemoryCore/dsh-plugin/README.md
+++ b/MemoryCore/dsh-plugin/README.md
@@ -2,6 +2,4 @@
 
 This bundle provides the smallest native DSH integration: automatic recall before a step, exactly-once per-turn L0 capture, asynchronous Skill conversation delivery, and three read-only search tools. L1/L2/L3 and Skill extraction remain backend-owned asynchronous MemoryCore work.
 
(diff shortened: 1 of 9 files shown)


Maxwell-Code07:

Vocllum:
Closing this PR because the DSH integration is better maintained as a standalone community plugin rather than merged into MemoryCore. The plugin now lives at https://github.com/Vocllum/tdai-memory-dsh-plugin, keeping the host-specific DSH lifecycle adapter independent from the TencentDB Agent Memory Gateway/Core implementation.

Vocllum:
Update: the DeepSeek Harness adapter has been moved to a clean, clearly named repository: https://github.com/Vocllum/DSH-TDAI-Memory-Adapter

This keeps the DSH adapter separate from the existing dsh-tdai-memory local MemoryCore port and avoids package/repository name ambiguity.

PR #1054, opened 2026-08-18
docs: correct the branch table on main — main is not the default branch

JohnZhangYH63:
## Description | 描述

`main`'s contributor guide states that `main` is the default branch and the PR target:

```
| Branch | Purpose |
|--------|---------|
| `main` | Default branch, PR target |
```

It isn't. `GET /repos/TencentCloud/TencentDB-Agent-Memory` reports `feat/server_team` as the default branch, and the repository has three long-lived branches — `feat/server_team`, `feat/server` and `main`.

This PR lists all three, and adds the one fact that actually helps a contributor choose: CI (`.github/workflows/pr-ci.yml`) is scoped to `branches: [main]`, so `main` is currently the only branch where a PR gets a CI run. Step 1 of the flow, which told contributors to branch off `main` unconditionally, now points at the target branch instead. Applied to both the English and Chinese guides.

## Related Issue | 关联 Issue

Companion to #1053, which fixes the *other* copy of this guide. The two branches carry different `CONTRIBUTING` documents with different errors:

| Branch | What its guide says | Status |
|---|---|---|
| `feat/server_team` (default) | branch off `master` / `develop_*`, PR to `develop_server_team` or `master` — none of those four branches exist | fixed in #1053 |
| `main` (this PR) | `main` is the default branch and the PR target | fixed here |

Happy to send the `feat/server` variant too if the guide there has drifted as well.

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [x] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

Markdown only, no code touched. Branch names verified against `GET /repos/TencentCloud/TencentDB-Agent-Memory/branches`; the default branch against the repository object. The diff touches none of the files guarded by `scripts/ci/check-skill-queue-isolation.sh`.

## Additional Notes | 其他说明

Unlike #1053, this one does get a CI run, since it targets `main`.

PR #1053, opened 2026-08-18
docs: fix stale branch names in CONTRIBUTING and maxScenes default comment

JohnZhangYH63:
## Description | 描述

Two stale contributor-facing references on the default branch.

**1. `CONTRIBUTING.md` / `CONTRIBUTING_CN.md` — the contribution flow names branches that do not exist.**

Step 2 says to cut a feature branch off `master` or `develop_*`; step 5 says to open the PR against `develop_server_team` or `master`. None of those four branches exist. The repository has `feat/server_team` (default), `feat/server` and `main`, so a contributor following the guide verbatim is blocked at step 2.

Replaced both references with a short **Branches** section listing the three real branches, and noted that CI (`.github/workflows/pr-ci.yml`) is scoped to `branches: [main]` — so PRs opened against the default branch currently get no CI run. Applied to both the English and Chinese guides.

**2. `MemoryCore/src/config.ts` — `PersonaConfig.maxScenes` is documented as `default: 20`, implementation is 15.**

Every location that actually sets it uses 15:

| Location | Value |
|---|---|
| `src/config.ts:556` | `?? 15` |
| `src/core/scene/scene-extractor.ts:114` | `?? 15` |
| `tdai-gateway.yaml:58` | `15` |
| `tdai-gateway.standalone.yaml:39` | `15` |
| `openclaw.plugin.json:63` | `"default": 15` |
| `deploy/global-images/start-memory-core.sh:86` | `15` |
| `scripts/e2e-memory-prompt-vdb-cos.ts:382` | `15` |

Every neighbouring field in the same interface has a comment matching its implementation (`triggerEveryN` 50, `backupCount` 3, `sceneBackupCount` 10) — `maxScenes` is the only drift, so the comment is corrected rather than the code. If 20 was the intended default, the fix belongs on the other side and in the seven locations above; flagging for a maintainer either way.

## Related Issue | 关联 Issue

Item 2 was reported in #1038 (closed without the comment being synced).

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [x] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

[...]

PR #1055, opened 2026-08-18
docs: correct the branch table on feat/server — main is not the default branch

JohnZhangYH63:
## Description | 描述

`feat/server` carries the same contributor guide as `main` (both `CONTRIBUTING.md` and `CONTRIBUTING_CN.md` are byte-identical between the two branches), and therefore the same inaccuracy:

```
| Branch | Purpose |
|--------|---------|
| `main` | Default branch, PR target |
```

`main` is not the default branch — `GET /repos/TencentCloud/TencentDB-Agent-Memory` reports `feat/server_team` — and the repository has three long-lived branches: `feat/server_team`, `feat/server`, `main`.

This lists all three and notes that CI (`.github/workflows/pr-ci.yml`) is scoped to `branches: [main]`, which is the fact that actually helps a contributor pick a target. Step 1 of the flow, which said to branch off `main` unconditionally, now points at the target branch. Both language versions updated.

This is the identical patch to #1054 (same `git patch-id`), applied to `feat/server`.

## Related Issue | 关联 Issue

Third and last of a set covering all three branches, whose guides had drifted apart:

| Branch | What its guide said | PR |
|---|---|---|
| `feat/server_team` (default) | branch off `master` / `develop_*`, PR to `develop_server_team` or `master` — none of those four branches exist | #1053 |
| `main` | `main` is the default branch and PR target | #1054 |
| `feat/server` | same text as `main` | this PR |

#1053 additionally corrects `PersonaConfig.maxScenes`, whose JSDoc says `default: 20` while all seven places that set it use 15 (reported in #1038). That field does not exist on `main` or `feat/server`, so this PR and #1054 are documentation-only.

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单
- [x] No existing functionality affected | 无影响现有功能

Markdown only. Branch names verified against `GET /repos/TencentCloud/TencentDB-Agent-Memory/branches`, the default branch against the repository object, and the patch equivalence with #1054 via `git patch-id --stable`. The diff touches none of the files guarded by `scripts/ci/check-skill-queue-isolation.sh`.

## Additional Notes | 其他说明

Since `pr-ci.yml` triggers only on `branches: [main]`, this PR gets no CI run — same as #1053.


Maxwell-Code07:

PR #1082, opened 2026-08-20
fix(memory-core): upsert L1 notes in atomic/update and honor recordIds lookup

VXenomac:
## Summary

Two related fixes in **MemoryCore** so that `/v3/atomic/update` works as an **upsert** and correctly resolves notes by primary key:

1. **`sqlite.ts` — `queryL1Records` now honors `filter.recordIds`**
   The SQLite backend ignored `recordIds`, falling back to a full-table read, so `existing[0]` was an arbitrary row instead of the requested note. On a multi-agent team this surfaced as `403 Atomic note ... belongs to a different agent` for any non-default agent. The TCVDB backend already honors `recordIds` (`documentIds`), so this aligns SQLite with TCVDB.

2. **`v2-router.ts` — `handleAtomicUpdate` falls back to creating the note**
   Previously a missing id returned `404 Atomic note not found`. L1 notes were only ever produced by the L0→L1 extraction pipeline, leaving no entry point to bulk-import or seed historical memory. Now, when the id does not exist, a new L1 note is created (upsert semantics), scoped to the request isolation triple (`team_id`/`user_id`/`agent_id`), starting at `version v1`, with an optional body `type` (`episodic` | `instruction` | `persona`) defaulting to `episodic`.

## Motivation

Migrating a legacy memory store into a fresh team required writing thousands of L1 notes across multiple agents — neither possible before (no create path, and writes to non-default agents were 403'd).

## Relation to #821

PR #821 fixes the `recordIds` lookup alone. This PR keeps the pair self-contained because the upsert path depends on correct primary-key lookup; the `recordIds` part can be dropped if #821 lands first.

## Change Type
- [x] Bug fix

## Verification
- `atomic/update` on a fresh id + non-default agent now returns `200 version v1` (previously 403/404); update of an existing note still bumps `version`.
- No other callers of `queryL1Records` / `handleAtomicUpdate` are affected (the filter addition is a superset of previous behavior for non-`recordIds` queries).

### DIFF
--- a/MemoryCore/src/core/store/sqlite.ts
+++ b/MemoryCore/src/core/store/sqlite.ts
@@ -1789,7 +1789,7 @@ export class VectorStore implements IMemoryStore {
       return [];
     }
     try {
-      const { sessionKey, sessionId, taskId, updatedAfter } = filter ?? {};
+      const { sessionKey, sessionId, taskId, updatedAfter, recordIds } = filter ?? {};
 
       let raw: Record<string, unknown>[];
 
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1093, opened 2026-08-20
feat(knowledge): add LocalSourceFetcher for local git repositories

agentlabbusiness:
## What

Implements the `LocalSourceFetcher` stub referenced in `source-fetcher/types.ts` and `registry.ts` (`// 未来：this.register(new LocalSourceFetcher())`). Code-graph `create` now accepts absolute paths and `file://` URLs pointing to local git repositories.

## Why

Teams whose repositories live only on the local machine (private monorepos, air-gapped setups, no public HTTPS remote) currently cannot use Code-Graph at all: `create` accepts the path, then the worker fails with `unsupported source type: local`.

## How

- `validate()` requires an existing directory that contains `.git` — arbitrary non-git directories are rejected.
- `fetch()` uses plain `git clone <path>` into the service-owned data dir (git does not copy hooks on clone).
- `sync()` fetches and hard-resets the service-owned clone to `origin/<branch>`; the source repository is never written to.

## Tested

- macOS (Intel), source run via tsx.
- Negative: non-git path → clear validate error; pre-patch behavior (`unsupported source type: local`) reproduced first.
- Positive: a local repo indexed end-to-end (79 files / 5,232 nodes / 23,681 edges); `callers`/`impact` answers verified against manually-read ground truth; service restart re-syncs the graph successfully.

### DIFF
--- a/MemoryKnowledge/src/source-fetcher/local-fetcher.ts
+++ b/MemoryKnowledge/src/source-fetcher/local-fetcher.ts
@@ -1,0 +1,57 @@
+/**
+ * LocalSourceFetcher — локальные git-репозитории (абсолютный путь или file://).
+ *
+ * Пилот-патч 2026-08-18 (см. types.ts: «LocalSourceFetcher / FtpSourceFetcher — 未来扩展»).
+ * Безопасность: путь обязан существовать и содержать .git — произвольные каталоги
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1094, opened 2026-08-20
fix(gateway): search TDAI_DATA_DIR when resolving the config file

nocode-ananas:
## Description | 描述

`resolveConfigPath()` searches `TDAI_GATEWAY_CONFIG`, the CWD, and the *default* data dir — but not the directory pointed to by `TDAI_DATA_DIR`. When that variable relocates the data storage, a `tdai-gateway.yaml`/`.json` kept next to the data is never found and the gateway silently runs on built-in defaults.

Observed on 1.0.1 (attach-to-Hermes setup, Linux container): `TDAI_DATA_DIR` was set and honored for data storage, the config file sat in that directory, yet the gateway kept `bm25.language: zh` — the first German record was tokenized with jieba and keyword recall returned score 0.000 until the gateway was restarted with an explicit `TDAI_GATEWAY_CONFIG`.

This PR inserts the `TDAI_DATA_DIR` location into the search order between CWD and the default data dir (mirroring the `~/` expansion used for `data.baseDir`), and updates the resolution-order doc comment.

## Related Issue | 关联 Issue

—

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过 — `vitest run`: 71/71 passing, incl. 4 new cases in `src/gateway/config.test.ts` (found via `TDAI_DATA_DIR`; ignored when unset; CWD precedence; `TDAI_GATEWAY_CONFIG` precedence)
- [x] No existing features affected | 无影响现有功能 — the new step only runs when `TDAI_DATA_DIR` is set and only *adds* a search location; all existing steps and their order are unchanged

## Additional Notes | 其他说明

Found while evaluating the 1.0.1 npm package in a sandboxed Hermes 0.20.4 deployment; the bug is still present on `main`.

### DIFF
--- a/src/gateway/config.test.ts
+++ b/src/gateway/config.test.ts
@@ -1,0 +1,85 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { loadGatewayConfig } from "./config.js";
+
+/**
+ * Config-file resolution, focused on the `TDAI_DATA_DIR` search step:
+ * a config file stored next to the relocated data dir must be picked up
+ * (previously only CWD and the *default* data dir were searched, so the
+ * gateway silently ran on defaults whenever `TDAI_DATA_DIR` was set).
+ */
+describe("loadGatewayConfig config-file resolution", () => {
+  let dataDir: string;
+  let emptyCwd: string;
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1111, opened 2026-08-21
fix(memory-core): drain L3 cascade during pipeline destroy

### LINKED ISSUE #1049: [Bug] destroy flush runs L2 but suppresses its L3 cascade

N/A — source-level reproduction in the standalone `MemoryPipelineManager`

### Plugin Version | 插件版本

Repository `feat/server_team` at `2608046`

### Operating System | 操作系统

macOS

Node.js 22.22.2

### Describe the bug | 问题描述

`MemoryPipelineManager.destroy()` can report `Pipeline flushed successfully` after running a pending L2 task while silently skipping the corresponding L3 persona generation.

The shutdown sequence is deterministic:

1. `destroy()` sets `this.destroyed = true` before `_doFlush()`.
2. `_doFlush()` explicitly flushes pending `L2_schedule` timers, so the L2 task still runs.
3. `runL2()` clears `l2_pending_l1_count`, persists the state, and calls `triggerL3()`.
4. `triggerL3()` immediately returns because `this.destroyed` is already true.
5. `_doFlush()` observes both queues idle and `destroy()` logs a successful flush.

This L3 work is not recoverable on restart: L2 already persisted `l2_pending_l1_count = 0`, and there is no persisted `l3_pending` marker.

The issue is especially visible in the `seed` command. `seed-runtime.ts` currently waits only for L1 and then calls `pipeline.destroy()`, so the latest L2 can be forced during destroy while its L3 cascade is suppressed. The file's FIXME already notes that the latest L2/L3 artifacts may be absent, but `getQueueSizes()` now exposes L2/L3 status and the current destroy path claims to flush all three layers.

There is a second ordering hazard to account for when fixing the guard: `_doFlush()` awaits `l2Queue.onIdle()` and `l3Queue.onIdle()` concurrently. If L3 is idle when the promises are created and L2 enqueues L3 later, the original L3 idle promise has already resolved. A correct full drain must observe the dependency order (L2 before the final L3 idle check).

### To Reproduce | 复现步骤

Code-level sequence:

1. Construct `MemoryPipelineManager` with an L2 runner and L3 runner that increment counters.
2. Start the manager and produce an L1 completion so an `L2_schedule` timer is pending.
3. Call `destroy()` before that timer naturally fires.
4. Observe that the L2 counter increments because `_doFlush()` calls `timers.l2Schedule.flush()`.
5. Observe that the L3 counter remains zero while the logger emits `Pipeline flushed successfully`.

The relevant paths are:

- `MemoryCore/src/utils/pipeline-manager.ts:531-605`
- `MemoryCore/src/utils/pipeline-manager.ts:916-1007`
- `MemoryCore/src/core/seed/seed-runtime.ts` module FIXME and final `pipeline.destroy()`

### Expected behavior | 预期行为

One of these shutdown contracts should be made explicit and internally consistent:

- A successful full flush drains L1 → L2 → L3 in dependency order, including L3 work produced by a flushed L2; or
- Shutdown intentionally checkpoints unfinished work, but then it must persist a recoverable L3-pending state and must not report a full successful flush.

For standalone seed, successful completion should not omit the latest derived L2/L3 artifacts.

No exception is emitted. The misleading success path is:

```text
[memory-tdai] [pipeline] Flush: triggering L2 schedule timer
[memory-tdai] [pipeline] L2 complete
[memory-tdai] [pipeline] Pipeline flushed successfully
```

There is no subsequent `L3 running` because `triggerL3()` returns while `destroyed === true`.

### Additional context | 补充信息

- This is narrower than #115: the failure does not require an LLM parse error or queue conflict; it follows directly from the destroy-state guard.
- It is separate from #505/#867, which concern the L1 idle condition and stalled seed rounds.
- No matching open issue or PR was found using `seed L3 destroy flush`, `pipeline destroy L3`, and related searches.
- I would be happy to implement a focused regression test and the smallest fix after maintainers confirm whether gateway shutdown and standalone seed should share the same timeout/flush contract.

Gout999:
## Summary

- Keep the destroy-time drain window open long enough for an already-accepted L2 task to enqueue its L3 persona cascade.
- Drain the L2 queue before checking L3 idle, so an L3 task created by L2 cannot be missed by `destroy()`.
- Add a regression test that holds L3 open and verifies `destroy()` waits for it to finish.

## Root cause

`destroy()` marks the pipeline as destroyed before flushing pending L2 timers. The L2 task still runs, but `triggerL3()` immediately returns because of the destroyed guard. In addition, `_doFlush()` checked L2 and L3 idle concurrently; an idle L3 check could resolve before L2 enqueued its cascade.

Fixes #1049.

## Validation

- Focused Vitest: `vitest run src/utils/pipeline-manager.test.ts` — 1/1 passed.
- Focused TypeScript check with strict mode — passed.

### DIFF
--- a/MemoryCore/src/utils/pipeline-manager.test.ts
+++ b/MemoryCore/src/utils/pipeline-manager.test.ts
@@ -1,0 +1,70 @@
+import { describe, expect, it, vi } from "vitest";
+
+import { MemoryPipelineManager, type PipelineConfig } from "./pipeline-manager.js";
+
+const config: PipelineConfig = {
+  everyNConversations: 1,
+  enableWarmup: false,
+  l1: { idleTimeoutSeconds: 60 },
+  l2: {
+    delayAfterL1Seconds: 60,
+    minIntervalSeconds: 60,
+    maxIntervalSeconds: 3600,
+    sessionActiveWindowHours: 24,
+  },
+};
+
+describe("MemoryPipelineManager", () => {
+  it("drains the L3 cascade produced by a flushed L2 task", async () => {
+    const l2Runs = vi.fn();
+    const l3Runs = vi.fn();
+    let releaseL3!: () => void;
+    const l3Finished = new Promise<void>((resolve) => {
+      releaseL3 = resolve;
+    });
+    const manager = new MemoryPipelineManager(config);
+
+    manager.setL1Runner(async () => ({ processedCount: 1 }));
+    manager.setL2Runner(async () => {
+      l2Runs();
+      return { latestCursor: "2026-08-19T00:00:00.000Z" };
+    });
+    manager.setL3Runner(async () => {
+      l3Runs();
+      await l3Finished;
+    });
+
+    await manager.notifyConversation("session-1", [{
+      role: "user",
+      content: "A captured message",
+      timestamp: "2026-08-19T00:00:00.000Z",
+    }]);
+
+    await vi.waitFor(() => {
+      expect(manager.getSessionState("session-1")?.l2_pending_l1_count).toBe(1);
+    });
+    expect(l2Runs).not.toHaveBeenCalled();
+    expect(l3Runs).not.toHaveBeenCalled();
+
+    let destroyFinished = false;
+    const destroyPromise = manager.destroy().then(() => {
+      destroyFinished = true;
+    });
+
+    await vi.waitFor(() => {
+      expect(l3Runs).toHaveBeenCalledTimes(1);
+    });
+    expect(destroyFinished).toBe(false);
+
+    releaseL3();
+    await destroyPromise;
+
+    expect(l2Runs).toHaveBeenCalledTimes(1);
+    expect(l3Runs).toHaveBeenCalledTimes(1);
+    expect(manager.getQueueSizes()).toMatchObject({
+      l1Idle: true,
+      l2Idle: true,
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

kvnloo:
Thanks for working on this! I ran a focused validation pass against the shutdown behavior here.

I was able to reproduce the original #1049 failure on the base revision, then verified this PR against a small scheduler-level matrix covering:

- L2 → L3 cascading work during `destroy()`
- overlapping updates while a level is already running
- rejection of genuinely new work after shutdown begins
- the existing shutdown deadline
- ordered draining so L3 work created by L2 is observed before returning

The candidate passed the matrix consistently across repeated runs. As negative controls, removing the ordered drain reproduced the early-return failure, and removing the pending-rerun handling reproduced dropped follow-up work.

One adjacent case I noticed: if `destroy()` begins while L1 itself is still running, that L1 completion can still be prevented from scheduling its downstream L2 work because shutdown has already started. I confirmed that behavior exists on the base revision as well, so I don't think it should expand the scope of this PR.

Would it make sense to treat that as a separate lifecycle question: once work has already been admitted before shutdown, should its full L1 → L2 → L3 dependency chain be considered part of the drain contract?

For the narrower #1049 case, the ordering and rerun handling here look consistent with that contract from my testing. Thanks again for the fix.

PR #1113, opened 2026-08-21
fix(proxy): surface failed L0 writes to retry wrapper

### LINKED ISSUE #1089: [Bug][Proxy] /v3/conversation/add failures are swallowed, bypassing L0 retry and observability
## Summary

## Affected flow

```text
recordTdaiTurn()
```

- non-2xx HTTP responses;

## Minimal reproduction

```ts
await withL0Retry(
  { attempts: 3, baseMs: 1 },
);
```

Expected: 3 calls to `fetch`.

## Impact

## Expected behavior

For the L0 write path:

## Related work

Gout999:
Fixes #1089.

## Summary

- Make the L0 `/v3/conversation/add` path reject HTTP failures, invalid JSON, and non-zero gateway envelopes instead of returning an empty success object.
- Preserve the existing fail-soft behavior for read/recall calls.
- Carry HTTP status and explicit retry metadata into `withL0Retry`, so 5xx/408/429 and network failures retry while 4xx and business-envelope failures stop immediately.
- Add regression tests for 503 retry, 400 no-retry, and non-zero envelope no-retry behavior.

## Validation

- `npm exec vitest run src/tdai/__tests__/client.test.ts` — 3 passed.
- Strict targeted TypeScript compile for the changed client/retry/test files — passed.

### DIFF
--- a/MemoryProxy/src/tdai/__tests__/client.test.ts
+++ b/MemoryProxy/src/tdai/__tests__/client.test.ts
@@ -1,0 +1,71 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { TdaiClient } from "../client.js";
+import { withL0Retry } from "../pending-writes.js";
+import type { TdaiIdentity, TdaiMemoryConfig } from "../types.js";
+
+const config: TdaiMemoryConfig = {
+  enabled: true,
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

Issue #1098, opened 2026-08-21
[Bug][Proxy] Missing OpenAI-compatible /models endpoint causes custom-provider validation 404


stoner-byte:
## Bug description

`MemoryProxy` supports OpenAI-compatible chat forwarding, but does not expose a model-listing endpoint. Hermes custom-provider model validation probes `<base_url>/models` (and, when applicable, `<base_url>/v1/models`) and receives HTTP 404.

With a base URL such as:

```text
http://127.0.0.1:8096/hermes/default
```

Hermes probes:

```text
GET /hermes/default/models
GET /hermes/default/v1/models
```

Both currently return `404 Not Found`.

## Evidence

- Current `MemoryProxy/src/server.ts` registers chat endpoints such as `POST /v1/chat/completions`, `POST /:agent/:spaceId/v1/chat/completions`, and the catch-all POST handler.
- The server also registers GET endpoints such as `/health` and `/whoami`, but no `GET /models`, `GET /v1/models`, or equivalent agent-scoped model-listing route.
- The MemoryProxy README documents chat/message endpoints but does not document a model-listing endpoint.
- This was reproduced against a locally running `tdai-proxy`; it is an endpoint compatibility issue, not evidence that chat forwarding or L0 persistence is failing.

## Impact

- Hermes cannot automatically verify configured model names or populate model choices from the proxy.
- Hermes accepts the model with a warning, but users see a misleading endpoint/model-listing error during model switching.
- Other OpenAI-compatible clients that expect `GET /v1/models` may have similar compatibility issues.

## Expected behavior

Expose an OpenAI-compatible model listing endpoint at the supported base URL(s), for example:

```text
GET /hermes/default/models
GET /hermes/default/v1/models
```

or document the officially supported endpoint shape and configure clients accordingly.

The response should follow the OpenAI model-list format, e.g.:

```json
{
  "object": "list",
  "data": [
    {"id": "gpt-5.6-terra", "object": "model", "owned_by": "memorydb"},
    {"id": "gpt-5.6-sol", "object": "model", "owned_by": "memorydb"},
    {"id": "gpt-5.6-luna", "object": "model", "owned_by": "memorydb"}
  ]
}
```

## Environment

- Component: `MemoryProxy`
- Local proxy port: `8096`
- Client: Hermes custom provider
- Endpoint protocol: OpenAI-compatible chat forwarding

## Scope note

[...]

PR #1114, opened 2026-08-21
fix(proxy): add OpenAI-compatible model discovery

Gout999:
Addresses #1098.

## Summary

- Add OpenAI-compatible `GET /models` and `GET /v1/models` endpoints.
- Add the same model-listing paths under `/<agent>/<spaceId>/`, which covers Hermes custom-provider probes such as `/hermes/default/models`.
- Expose the configured public `modelName` alias when present, fall back to the configured model ID, and return an empty list when no pricing table is configured.
- Document the endpoint in both MemoryProxy READMEs.

## Validation

- `npm exec vitest run src/__tests__/server-models.test.ts` — 2 passed.
- The package-wide `npm run typecheck` still reports existing errors in unrelated files (`anthropicHandler.ts`, `codexHandler.ts`, `config.ts`, and the cost-guard alias); it is not reported as a pass here.


Maxwell-Code07:

Issue #1096, opened 2026-08-20
Standalone/Hermes deployments in non-Chinese languages: bm25 default, hybrid hard-fail without embeddings, 1.0.1 packaging drift


nocode-ananas:
We evaluated the 1.0.1 npm package attached to a sandboxed Hermes 0.20.4 (Linux container, Node 26, OpenAI-compatible hosted LLM, keyword recall). The core promise held up well — capture is fire-and-forget, extraction runs entirely off the reply path (measured: `/recall` 1–8 ms without hits, 75–107 ms with hits; `/capture` HTTP 10–119 ms; an LLM-extraction timeout landing 3 minutes *after* the agent reply nicely proved the decoupling), and cross-session recall worked. Four DX findings from that run:

**1. `bm25.language` defaults to `zh`**
Our first German record was tokenized with jieba; German queries scored 0.000 in FTS and prompt injection stayed empty, while the search tool (with its fallback) did find the record. After restarting with `bm25.language: "en"` injection worked. Suggestion: log the active tokenizer language at startup, and document prominently that non-Chinese deployments must set it — or derive the default from locale.

**2. `recall.strategy` default `hybrid` hard-fails without an embedding service**
Intentional ("no degradation"), but several OpenAI-compatible hosted inference offerings expose no `/v1/embeddings`, and the error surfaces per-request rather than at startup. Suggestion: fail fast at startup with a message naming both fixes (`recall.strategy: keyword`, or configuring an embedding endpoint).

**3. 1.0.1 packaging drift**
- The packaged plugin README documents `MEMORY_TENCENTDB_LLM_*` gateway variables; the gateway reads `TDAI_LLM_*`.
- The injected tool guide references `tdai_memory_search` / `tdai_conversation_search`, while the Hermes provider registers `memory_tencentdb_memory_search` / `memory_tencentdb_conversation_search`.

Both cost real debugging time on 1.0.1. If a 1.0.2 is planned, aligning these would help.

**4. Observation: Hermes one-shots with a model override never capture**
`hermes -m <model> --provider custom -z "…"` fired `/recall` and `/session/end` but never `/capture` (2× reproduced); default invocations captured fine (2× verified). Possibly provider-side or a race on fast process exit — happy to file separately with logs if useful.

Happy to provide more detail from the evaluation notes on any of these.

Maxwell-Code07:
[...]

PR #1115, opened 2026-08-21
fix(gateway): accept Hermes LLM environment aliases

Gout999:
Partially addresses #1096, focusing on the Hermes/Gateway LLM environment-variable mismatch.

## Summary

- Accept `MEMORY_TENCENTDB_LLM_*` aliases in the Node Gateway alongside canonical `TDAI_LLM_*` variables.
- Keep `TDAI_LLM_*` precedence when both names are present.
- Cover provider, endpoint, model, max-token, and timeout aliases in the config path.
- Clarify the compatibility contract in the Hermes provider README.

## Validation

- `npm exec vitest run src/gateway/config.test.ts` — 2 passed.
- A targeted TypeScript compile also reports pre-existing repository errors in unrelated AI SDK/observability files and the existing `instanceId` typing; it is not reported as a pass here.

### DIFF
--- a/MemoryCore/hermes-plugin/memory/memory_tencentdb/README.md
+++ b/MemoryCore/hermes-plugin/memory/memory_tencentdb/README.md
@@ -231,14 +231,19 @@ eliminates a silent no-op.
 
 ### Gateway LLM (consumed by the Node sidecar, not by this provider)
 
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

Issue #1084, opened 2026-08-20
[Bug] Agent-scoped L3 persona never retriggers because L1 increments the global checkpoint


sheldonfa:
### Environment

- Deployment: standalone MemoryCore + MemoryHub
- Version: 2.0.0-beta.1
- Branch: `feat/server_team`
- Commit: `97f9465`
- OS: macOS with Docker Desktop

### Bug description

In `feat/server_team`, L1 memory extraction increments `memories_since_last_persona` in the global checkpoint, while L3 `PersonaTrigger` reads the team+agent-scoped checkpoint. As a result, the first L3 persona is generated by the cold-start condition, but subsequent threshold-based regeneration never occurs for the Agent.

Observed after importing conversations into one Agent:

- L1 records: 67
- L2 scene files: 14
- Agent-scoped `scenes_processed`: 34
- Global checkpoint `memories_since_last_persona`: 67
- Agent-scoped checkpoint `memories_since_last_persona`: 0

L3 repeatedly logs:

```text
Evaluating: total_processed=0, last_persona_at=0, memories_since=0, scenes=34
Persona generation not needed
```

Manual L3 generation succeeds and incorporates all scene files, so the failure appears to be in automatic trigger accounting rather than Persona generation itself.

### Steps to reproduce

1. Check out `feat/server_team` at `97f9465` and run MemoryCore + MemoryHub in standalone Docker mode.
2. Create a Team and an Agent.
3. Import enough conversations for that Agent to produce more than `persona.triggerEveryN` L1 memories (default: 50).
4. Wait for L1 and L2 processing to complete.
5. Compare the root `.metadata/checkpoint.json` with `profiles/<team+agent-scope>/.metadata/checkpoint.json`.
6. Observe that the root checkpoint accumulates `memories_since_last_persona`, while the Agent-scoped checkpoint remains 0.
7. Observe that the first L3 persona exists from cold start, but no later threshold-based L3 regeneration occurs.

### Expected behavior

Each stored L1 memory should increment the checkpoint counter for its own team+agent profile scope, or L3 should derive its threshold from scoped L1 records. Once the scoped count reaches `persona.triggerEveryN`, `PersonaTrigger` should regenerate that Agent persona and reset only the covered scoped counter.

### Relevant code paths

[...]

PR #1116, opened 2026-08-21
fix(memory-core): track persona counters per profile scope

Gout999:
## Summary

Addresses #1084 by separating the global L1 cursor from the profile-scoped persona
counter.

The pipeline still records the global L1 extraction cursor, but increments
`memories_since_last_persona` in the same team+agent scoped checkpoint used by L3
reads. The counter can also be updated explicitly through the checkpoint manager.

## Verification

- `npm test -- --run src/utils/checkpoint.test.ts` (2 tests passed)
- `git diff --check`


Maxwell-Code07:

Issue #889, opened 2026-08-09
[Feature] hermes agent 召回日志显示问题


muzijiang:
### Is this feature related to a problem? | 该功能需求是否与某个问题相关？请描述

hermes agent 召回日志显示杂乱，看起来是缺少换行，导致可读性非常差，不知道是因为配置问题，还是程序本来就这样，希望能提供解决方法。

<img width="1386" height="516" alt="Image" src="https://github.com/user-attachments/assets/69e473bc-b84a-429d-bf70-ae10a207908d" />

### Describe the solution you'd like | 描述你期望的解决方案

如果是程序问题，希望能改善可读性。

Maxwell-Code07:
Thanks for your feedback. We will arrange an internal evaluation as soon as possible and follow up on this matter later.

PR #1119, opened 2026-08-21
fix(hermes): preserve line breaks in recall output

Gout999:
## Summary

Addresses #889 by preserving line breaks in Hermes recall output.

Multiline L1 prefetch results, memory-search results, and conversation-search results
now indent continuation lines. This keeps the result boundaries readable in the
host's tool output without changing the underlying memory content.

## Verification

- Inline import smoke test for multiline formatting
- `python3 -m py_compile MemoryCore/hermes-plugin/memory/memory_tencentdb/__init__.py`
- `git diff --check`


Maxwell-Code07:

Issue #964, opened 2026-08-12
L0 write never attempted with Claude Code — tdai-recorder:write-l0 never logged, all memory layers stay empty


GvAiSuperAdmin:
### Summary

With the full three-in-one stack running and Claude Code pointed at the proxy,
no conversation is ever recorded. Injection works, but nothing is captured, so
memory never accumulates.

### Environment

- Upstream `4dca55c`, deployed via `deploy/global-images/start-all.sh`
- macOS 26 (Apple Silicon), OrbStack 29.4.0, images `agentmemory/*:latest`
- Upstream LLM: OpenRouter — `deepseek/deepseek-chat` (memory group),
  `anthropic/claude-sonnet-4.5` (proxy group)
- Client: Claude Code v2.1.226 via
  `ANTHROPIC_BASE_URL=http://127.0.0.1:8096/claude-code/default`

### What works

- All three containers healthy; panel reachable
- `auth` and `sessionInit` succeed; team and agent auto-selected from headers
- Injection runs every turn — `[injection-debug] … injectionEnabled=true
  injectors=["skill","knowledge","tdai-memory"] injectedSkipped=false`

### What doesn't

After several multi-turn working sessions (file reads, edits, shell commands):

- `GET /health` → `services.pipelineWorker` — every counter still `0`
  (`tasksConsumed`, `tasksCompleted`, `tasksFailed`, `deadLetterCount`)
- Panel → Chat_Memory → Agent Assets → the agent's memory block: **0 items at
  every layer, including L0**
- `docker logs tdai-proxy | grep write-l0` → no matches, ever

Since `tdai-recorder:write-l0` is logged unconditionally at the top of
`TdaiClient.addConversation`, the write is not being attempted at all.

### Ruled out

- `writeL0: true` is present in the generated `.proxy-config/config.yaml`
- `tdai.enabled` is true (`start-proxy.sh` reports `tdai=true`)
- Identity is valid — `deriveTdaiIdentity` requires only
  team/user/agent/session, and `taskId` is optional, so skipping Task binding
  is not the cause
- Proxy→core connectivity is fine (injection queries succeed over the
  `memory-core` network alias)

### Suspected cause

`recordTdaiTurn` returns early when `extractLatestUserMessage` yields null.
That helper takes the **last** `role: user` message and passes it through
`extractUserQueryText` (`MemoryProxy/src/common/user-query-extractor.ts`), which
discards content it classifies as harness noise.

[...]

PR #1120, opened 2026-08-21
test(proxy): cover Claude Code tool-result capture

Gout999:
## Summary

Adds regression coverage for #964.

The test locks the Claude Code capture rule that walks backwards past a harness
`<tool_result>` user message to find the real `<user_query>`, and returns no query when
all trailing user messages are harness output.

The current source already contains the backwards scan; this PR protects that behavior
from future regressions rather than claiming a new source rewrite.

## Verification

- `npm test -- --run src/tdai/__tests__/recorder.test.ts` (2 tests passed)
- `git diff --check`


Maxwell-Code07:

Issue #1074, opened 2026-08-20
[Discussion] dsh (non-standard presets) probes non-existent proxy API paths after session-init — how to add a one-shot 'session already initialized' notice


cxin21:
## Summary

After session-init completes, coding-agent models (observed with DeepSeek-Harness / dsh on non-standard presets, e.g. `code` / `cordis`) sometimes **probe non-existent proxy API paths** instead of using the injected tool templates — even though the session is already initialized and `<skill_tools>` / `<tdai_memory_tools>` are correctly injected with working URLs.

This is not a functional failure (binding, memory injection and skill injection all work), but it burns turns, confuses the user, and makes team-memory adoption feel broken.

## Observed behavior

Session `5bfd6fb5` (dsh, `cordis` preset, DeepSeek v4-flash):

1. session-init correctly fired → `asset_confirm` → auto-select single team/agent → `completeRegistration`
2. Proxy injected `<session_context>`, `<tdai_profile_memory>` (full L3 memory), and `<skill_tools>` (correct `http://<public-ip>:8096/skill-bridge/v3/skill/*` templates)
3. Model then tried to **guess** endpoints that do not exist:
   - `/skill-bridge/v3/team/list`
   - `/skill-bridge/v3/session/teams`
   - `/openapi.json`, `/docs`, `/skill-bridge/openapi.json`
   - `/skill-bridge/v3/session/init`, `/session/options`, `/team/options`, `/agent/options` ...
4. All returned 404, model kept probing, user saw a messy first exchange.

Proxy logs confirm the session was healthy the whole time (no proxy-side error), and the injected template explicitly lists only `skill-bridge/v3/skill/*` and `memory-bridge/v3/*` paths.

## Root-cause hypothesis (client-side model behavior, not proxy bug)

- The `<skill_tools>` block says "use curl against these paths" but does **not** state that session-init is already done and that no metadata/session endpoints are needed.
- On `code`/`cordis` presets the persona tells the agent it can "read/write the harness runtime", which encourages endpoint probing.
- Windows PowerShell also trips on `curl` (alias for `Invoke-WebRequest`) — the template uses `curl` which fails under PowerShell until the model switches to `curl.exe`.

## Discussion / proposed mitigations (not yet implemented)

We'd like maintainers' take on the best placement:

1. **A one-shot "session already initialized" notice** — only on the first request after session-init completes (per-session flag), not every turn (avoid noise/token waste).
[...]

PR #1122, opened 2026-08-21
fix(proxy): guide agents after session initialization

Gout999:
## Summary

Addresses #1074 with a small prompt-guidance improvement for session-initialized
requests.

The injected Skill and Memory tool blocks now state that session initialization has
already completed, that the listed bridge paths are the supported entry points, and
that agents should not probe unrelated metadata, OpenAPI, or session endpoints.

The guidance is static and generic; it does not change client-specific routing.

## Verification

- `npm test -- --run src/injection/injectors/__tests__/session-init-guidance.test.ts` (2 tests passed)
- `git diff --check`


Maxwell-Code07:

Issue #1028, opened 2026-08-16
[Bug][Hermes] memory_tencentdb tools stay advertised (and loop the LLM) while the Gateway is permanently down


abgutierrez:
## Summary

When the memory-tencentdb Gateway is **permanently** unhealthy (crash-looping native deps, extraction LLM without quota, missing embeddings config, …), the Hermes provider keeps advertising `memory_tencentdb_memory_search` / `memory_tencentdb_conversation_search` to the LLM anyway. Every call returns a JSON error whose `hint` says *"The Gateway may still be starting up. Try again in a moment."*, so the model keeps retrying — in our deployment it entered a multi-turn loop alternating between the built-in `memory()` tool and `memory_tencentdb_memory_search()`, burning turns and tokens until a human intervened.

This is different from the startup-timing problem fixed in #220/#423: our Gateway was not *slow to start*, it was *never going to come up* — and the tool surface never reflected that.

## Environment

- Plugin: `@tencentdb-agent-memory/memory-tencentdb@1.0.1` (npm latest; verified the tarball's `hermes-plugin/memory/memory_tencentdb/__init__.py` — functionally identical to `main` for the code paths below)
- Host: Hermes Agent v0.19.0, Linux, Python 3.13
- Config: `memory.provider: memory_tencentdb`, `MEMORY_TENCENTDB_GATEWAY_CMD` + `MEMORY_TENCENTDB_GATEWAY_PORT` set (Gateway supervised by the provider)
- Gateway state: permanently broken in our case (extraction LLM out of quota → L1 never generated; embeddings backend not configured for hybrid recall), i.e. `_gateway_available == False` for the whole session

## Symptom

Observed in our deployment: the agent alternates `memory()` (built-in) → `memory_tencentdb_memory_search()` → error JSON → `memory()` → `memory_tencentdb_memory_search()` … for many consecutive tool rounds, with no useful output. The session is effectively unusable until the operator disables the provider. We ultimately uninstalled the plugin; the retry loop was the direct cause.

## Root cause chain

All line numbers from `hermes-plugin/memory/memory_tencentdb/__init__.py` (npm 1.0.1; `main` is identical modulo the newer Windows-launcher code):

1. **`initialize()` flips `_initialized` immediately, regardless of Gateway health** (L761–763):

   ```python
   # Mark as initialized immediately so tools are registered
   # (get_tool_schemas checks _initialized, not _gateway_available).
   self._initialized = True
   ```

[...]

PR #1123, opened 2026-08-21
fix(hermes): stop retry loop after gateway breaker opens

Gout999:
## Summary

Addresses #1028 with a small fail-closed behavior improvement for the Hermes memory tool.

When the gateway circuit breaker is still starting, the response keeps the existing
temporary retry hint. Once the breaker is open, the tool now returns a terminal
unavailable response with a fallback instruction to use built-in memory and stop
retrying `memory_tencentdb_*` tools.

This does not dynamically remove Hermes tool schemas because the host snapshots the
available tools; it prevents the repeated retry loop at invocation time.

## Verification

- `python3 -m py_compile MemoryCore/hermes-plugin/memory/memory_tencentdb/__init__.py`
- `git diff --check`


Maxwell-Code07:

Issue #921, opened 2026-08-10
[Compat] ensure-hook-policy.ts allowConversationAccess auto-patch reliability


dailylights:
## Summary

`MemoryCore/src/utils/ensure-hook-policy.ts` auto-patches `openclaw.json` to add `plugins.entries.memory-tencentdb.hooks.allowConversationAccess: true` when the host version >= 2026.4.24. This is needed for non-bundled plugins to have their `agent_end` hooks fire.

## Reliability Risks

1. **SDK method dependency**: The primary path uses `runtime.config.mutateConfigFile()`, which is an internal OpenClaw API. If this method changes signature or is removed, the fallback manual file write is used instead.

2. **$include not supported**: If the user's config uses `$include` directives, the auto-patch logs a warning and does nothing — user must manually add the setting.

3. **Gateway restart required**: After patching, the user must restart the gateway. The SDK path triggers auto-restart, but the manual fallback does not.

4. **Multiple calls**: OpenClaw calls `register()` multiple times during lifecycle. The code correctly checks `hasPolicyAlready()` before each patch attempt, but this still adds overhead.

## Impact

- If the auto-patch silently fails, `agent_end` hooks never fire
- Auto-capture (L0 recording) and auto-recall metrics are silently disabled
- User sees no error, just missing memory functionality

## Suggested Actions

- Document the manual configuration step in the README
- Add a diagnostic log at startup showing whether `allowConversationAccess` is active
- Consider a runtime check that verifies the hook is firing and warns if not

## Related

- Issue #910 (compat tracking)
- `src/utils/ensure-hook-policy.ts`

Maxwell-Code07:
We really appreciate your valuable feedback! Our team will conduct an internal review and assessment promptly.

PR #1125, opened 2026-08-21
fix(openclaw): log hook policy status during startup

Gout999:
## Summary

Addresses #921 with startup diagnostics for the OpenClaw hook policy.

The helper now logs when policy setup is skipped outside gateway startup and when
`allowConversationAccess` is already active, so operators can distinguish an intended
no-op from a failed configuration patch.

## Verification

- `npm test -- --run src/utils/ensure-hook-policy.test.ts` (1 test passed)
- `git diff --check`


Maxwell-Code07:

PR #1126, opened 2026-08-21
feat: add first-class Pi agent adapter

nicwn:
## Summary

Adds a **first-class `pi` agent-source** to the proxy + a minimal client Pi extension + docs, so [Pi](https://github.com/earendil-works/pi-coding-agent) can route through the proxy for team memory (L3 persona, L2 scene index, L0 capture, on-demand L0/L1/L2 search). The ROADMAP invites new client adapters; this is the Pi one.

Pi uses OpenAI Chat Completions with **label-line** system prompts (`Available tools:`, `Guidelines:`) — neither markdown headings (Claude Code) nor XML tags (CodeBuddy) — so it gets its own `PiProfile` parser rather than reusing an existing profile.

## What this adds (5 layers, all additive)

1. **Proxy adapter** — `MemoryProxy/src/agent-adapters/pi.ts` (classifyRequest + extractUserText), + case in the factory, + `"pi"` in the `AgentKind` union.
2. **Proxy AgentProfile** — `MemoryProxy/src/injection/agents/pi/` (`PiProfile`: label-line splitter, slot map, lossless parse→rebuild, graceful degradation to coarse injection points when a label is absent). Registered in `injection/index.ts`.
3. **Client plugin** — `MemoryCore/pi-plugin/` (Pi extension: `registerProvider` to `…/pi/<spaceId>/v1` + a `before_provider_headers` hook injecting a dynamic per-session `x-conversation-id`). Carries **only routing + the dynamic header**; all memory capability is delivered server-side by the proxy.
4. **Docs** — "Using Proxy with Pi" section in `INSTALL.md` + `MemoryCore/pi-plugin/README.md`.
5. **Tests** — first test files in the repo (vitest): adapter, profile (incl. byte-for-byte lossless rebuild + all 4 anchor relations + degradation), plugin (incl. fail-fast on missing env), and a regression test pinning `extractSpaceIdFromPath` for `pi`.

## Scope: additive, no core changes

The only edits to **existing** files are ~6 lines across 4 files in `MemoryProxy/src/` (none in `core/` or `deploy/layers/`):
- `agent-adapters/index.ts` (import + `case "pi"`)
- `agent-adapters/types.ts` (`| "pi"` in the `AgentKind` union)
- `injection/index.ts` (import + `["pi", new PiProfile()]`)
[...]

PR #1129, opened 2026-08-22
fix(proxy): header-identity agents get memory on a cache miss (no more silent bypass)

nicwn:
## Problem

Header-identity agents (e.g. **Pi**, which carries \`x-team-id\` / \`x-agent-id\` / \`x-task-id\` in request headers and never uses the interactive picker) get **no memory injected at all** on a cold start or after a cache miss.

### Symptom (observed with the Pi adapter from #1126)

Every turn logs:
```
[cache] session=pi:… miss (no bindingRepo) → history-scan
[session-recover] pi:… history scan → one-shot bypass (conversation exists but no form markers found)
[session-init] session=pi:… bypassed → skipping all injection
```
Zero injection ever runs for the session (\`applyInjection\` / \`recall\` / \`persona\` never appear). The \`x-team-id\` / \`x-agent-id\` / \`x-task-id\` headers are present and valid, but ignored.

### Root cause

\`SessionStore.getOrRecover()\` runs **before** \`handleSessionInit()\` in both \`handler.ts\` (OpenAI) and \`anthropicHandler.ts\` (Anthropic). On a binding/cache miss it falls to \`tryHistoryScan()\`, which only recognizes interactive-picker **form markers** (\`<question_answer>\` / \`AskUserQuestion\`). Header-identity agents carry identity in headers, not forms, so they have no form markers → \`tryHistoryScan\` returns a \`bypassed\` state → the handler sees \`bypassed=true\` and skips **all** injection.

The header-parsed \`presetIdentity\` that *should* drive registration is computed in the handler but never passed to \`getOrRecover\`, so the recovery path never consults it.

This is a **general** bug — any agent using header-based identity (Pi today; Claude Code w/ headers similarly) is affected on a cache miss.

## Fix

Thread \`presetIdentity\` into \`getOrRecover\`'s \`RecoveryContext\`. When present, \`tryHistoryScan\` skips the bypass and returns \`undefined\`, so the handler falls through to \`handleSessionInit\` — which already handles header registration via the existing \`headerAutoSelect\` path (\`session/preset.ts\`).

- No-preset sessions (interactive-picker agents) keep the **current** behavior (history-scan bypass) — unchanged.
- Header-identity sessions now register from their headers on a cold start, then the L1 cache is populated, so only the first post-restart turn pays the cost.

## Scope

4 files, +98 lines, purely additive:
- \`MemoryProxy/src/session/store.ts\` — \`RecoveryContext.presetIdentity\` + the defer guard in \`tryHistoryScan\`.
- \`MemoryProxy/src/handler.ts\` / \`anthropicHandler.ts\` — pass \`presetIdentity\` into the \`getOrRecover\` call (1 line each).
- \`MemoryProxy/src/session/__tests__/store-recover.test.ts\` — pins both the fix (preset → \`undefined\`) and the preserved behavior (no preset → \`bypassed\`).

No core/ or deploy/ changes. No behavior change for interactive-picker agents.

## Tests

\`npx vitest run\` → 4/4 (new test). The fix is independent of #1126 (discovered via the Pi adapter, but it's a pre-existing session-recovery bug on \`feat/server_team\`).


Maxwell-Code07:

nicwn:
**Rebased onto latest `feat/server_team` (c0cf94f) + test updated per review.**

- Resolved the `session/store.ts` import conflict (kept both upstream's `isDshRuntimeContextSnapshot` import and this PR's `PresetIdentity` import). The `RecoveryContext.presetIdentity` field and the `tryHistoryScan` early-return (`preset identity present → defer to handleSessionInit`) re-applied cleanly; `anthropicHandler.ts`/`handler.ts` threading of `presetIdentity` into `getOrRecover` auto-merged.
- **Updated `store-recover.test.ts` per item 2**: upstream's `tryHistoryScan` now returns `undefined` ("treating as uninitialized") on a history-but-no-marker miss instead of a one-shot bypass, so the no-preset-no-marker test now asserts `toBeUndefined()` (was `toBeDefined()`/`bypassed===true`). The docstring now notes upstream independently defers the no-marker case, and that this PR's preset short-circuit additionally guarantees header-identity deferral regardless of history content.

Tests: **4/4 store-recover pass** (full proxy suite 12/12). Typecheck: 55 errors, identical to upstream (zero new) — all line-number shifts in `handler.ts`/`anthropicHandler.ts` are the +1 from the `presetIdentity` threading line.

GitHub reports this PR **MERGEABLE**.

PR #1131, opened 2026-08-22
feat(proxy): task-optional memory — register from team+agent, no task gate

nicwn:
## What

Memory is now captured and recalled for **every** header-identity session regardless of task state — matching the kernel's own design and the "user won't pre-create a task" reality.

## Why

The proxy over-strictly required \`team + agent + task\` **all** present to register a session from request headers:

- \`resolvePresetIdentity.canRegister = agentId && taskId && !mismatch\`
- \`completeRegistration\` bypassed when \`resolved.task_id\` was missing

This silently dropped memory for any header-identity agent whose task was absent or stale. With the Pi adapter (#1126), a stale \`x-task-id\` (the referenced task deleted/renamed) caused \`preset mismatch → fallback to form\` → Pi can't answer the form → **no memory at all**, silently.

### The kernel never required this

\`task_id\` is an **optional** business dimension in the kernel — verbatim from \`MemoryCore/src/core/store/isolation.ts\`:

> *"taskId is an optional business dimension for L0/L1 filtering. It must never replace sessionId/sessionKey because L1 extraction is session-based."*

Memory is keyed on \`(team, user, agent, session)\`. \`task_id\` is a **soft recall filter**: present → narrows recall to that task; absent → broad recall across the agent's memories. It is not a gate. The proxy was stricter than its own backend.

## Changes

- **\`preset.ts\`** — \`canRegister = agentId && !hadMismatch\` (task dropped from the requirement). A stale/unknown \`task_id\` no longer flips \`hadMismatch\` (only unknown team/agent does); it is silently dropped → \`taskId\` stays \`undefined\` → broad recall.
- **\`codebuddy/init.ts\` + \`claude-code/init.ts\` \`completeRegistration\`** — remove the \`if (!resolved.task_id) → bypass\` guard. Registration with team+agent and no task now proceeds. The interactive \"本次不关联任务\" / \`defaultTaskId\` path is **unchanged** — it already routed here with a virtual task_id; this just extends the same behavior to the header path.
- Both \`canRegister\` branches **warn** (not block) when a provided \`task_id\` is stale, so operators can re-point the client via logs.

## Behavior matrix

| Headers sent | Before | After |
|---|---|---|
| team + agent + valid task | register (narrow recall) | register (narrow recall) — **unchanged** |
| team + agent, **no** task | bypass → **no memory** | register → **broad recall** ✅ |
| team + agent + **stale** task | mismatch → form → no memory | register → **broad recall** + warning ✅ |
| team + agent (unknown) | mismatch → no register | mismatch → no register — **unchanged** (agent is a mandatory dimension) |

## Scope

3 files, +45/-41, server-side only. No \`core/\` or \`deploy/\` changes. The interactive picker is untouched.

A companion client-side relaxation (making \`TDAI_TASK_ID\` optional in the Pi plugin) belongs on the adapter PR #1126 and will follow there.

## Tests

\`npx vitest run\` → 6/6 (new \`preset.test.ts\`). Pins: task-optional (no task → register), stale-task (unknown task → register, broad recall), preserved hard-mismatch (unknown agent/team → no register).

Independent of #1126 and #1129 (discovered via the Pi adapter; it's a pre-existing over-strict gate on \`feat/server_team\`).


Maxwell-Code07:

nicwn:
**Rebased onto latest `feat/server_team` (c0cf94f).**

Resolved the two base-drift conflicts in the `completeRegistration` task gate:
- `session/claude-code/init.ts` and `session/codebuddy/init.ts`: removed the `if (!resolved.task_id) → bypass` guard (this PR's intent), taking the branch side over upstream's guard (which had gained `resetFlow`/`resetEpoch`). The stale-task warning in the `canRegister` branch and the `preset.ts` `canRegister = agentId && !hadMismatch` change auto-merged.

Tests: **6/6 preset pass**. Typecheck: 54 errors (one *fewer* than upstream's 55 — the deleted task-gate block contained a pre-existing `resetFlow` read; no new errors). End-to-end task-optional verification on the rebased code is recorded in #1126 (team + agent, no `x-task-id` → `register directly`, injection, L0).

GitHub reports this PR **MERGEABLE**.

PR #1136, opened 2026-08-22
fix(memory-proxy): implement lossless forwarding for Anthropic server tools

### LINKED ISSUE #1135: [Bug] MemoryProxy corrupts Anthropic server tools during injection, causing Claude Code web_search 400
## OpenClaw Version | OpenClaw 版本

N/A — reproduced with Claude Code 2.1.239.

## Plugin Version | 插件版本

TencentDB-Agent-Memory `v2.0.1-beta.2` (`29d609a729704ae31ff1848dc6f8acb7e712106d`).

## Operating System | 操作系统

Ubuntu 24.04, Node.js 22.23.2, source deployment.

## Describe the bug | 问题描述

When Claude Code sends an Anthropic server-side tool such as:

```json
{
  "type": "web_search_20250305",
  "name": "web_search",
  "max_uses": 5
}
```

through MemoryProxy with injection enabled, the Anthropic injection adapter parses and serializes every tool through its generic `AgentTool` representation. `parseTool()` and `serializeTool()` retain only `name`, `description`, `input_schema`, and `cache_control`.

As a result, Anthropic-specific fields such as `type: "web_search_20250305"` and `max_uses` are dropped. Because this server tool has no custom `input_schema`, the adapter also adds `input_schema: {}`. The upstream then treats `web_search` as a normal custom function and rejects the empty schema:

```text
API Error: 400 Invalid schema for function 'web_search': schema must be a JSON Schema of 'type: "object"', got 'type: null'.
```

The relevant round trip is implemented in:

- `MemoryProxy/src/injection/adapters/anthropic.ts`: `parseTool()`
- `MemoryProxy/src/injection/adapters/anthropic.ts`: `serializeTool()`

This appears to be a general lossless-forwarding issue: the proxy should not discard provider-specific or future Anthropic tool fields merely because they are not part of the internal normalized type.

### 中文摘要

Claude Code 通过 MemoryProxy 调用 Anthropic Server Tool `web_search_20250305` 时，注入适配器只保留普通自定义工具字段，导致 `type`、`max_uses` 等字段丢失，并额外生成空的 `input_schema: {}`。上游因此把它当成普通函数并返回 400。建议对未知字段和 Server Tool 定义进行无损透传。

## To Reproduce | 复现步骤

1. Check out `v2.0.1-beta.2` at commit `29d609a` and run MemoryProxy from source.
2. Configure an Anthropic-compatible upstream and enable the injection pipeline.
3. Point Claude Code 2.1.239 at MemoryProxy and start a new session.
4. Send a request that causes Claude Code to invoke its built-in Web Search tool, for example asking for today's weather.
5. Observe `Web Search(...)` fail with the 400 response shown above, while a subsequent URL fetch can still work.

The lossy transformation can also be seen directly by round-tripping the server-tool object above through `AnthropicAdapter.parseTool()` and `serializeTool()`: the output loses `type` and `max_uses` and gains an empty `input_schema`.

## Expected behavior | 预期行为

- Anthropic server-side tools retain their `type` and all provider-specific fields across the injection pipeline.
- Unknown/future tool fields are preserved unless MemoryProxy intentionally modifies them.
- Server tools without `input_schema` do not receive a fabricated empty schema.
- Existing custom tool definitions and their JSON Schema continue to work unchanged.

A regression test should cover both a normal custom tool and `web_search_20250305` through a full parse/inject/serialize round trip.

## Error Logs / Screenshots | 报错日志/截图

Sanitized proxy log excerpt:

```text
UPSTREAM_4xx status=400 body={"error":{"message":"Invalid schema for function 'web_search': schema must be a JSON Schema of 'type: \"object\"', got 'type: null'.","type":"invalid_request_error"}}
```

No raw screenshot is attached because the observability view contains experiment user and session identifiers.

## Additional context | 补充信息

Related issue: #990 reports that historical Anthropic `thinking` blocks are not passed back to the upstream after a tool-call turn. The symptoms are different, but both indicate that the Anthropic request round trip can lose provider-specific information. This issue is intentionally scoped to tool-definition preservation; additional reproduction details for thinking preservation will be added to #990 instead of opening a duplicate.

manideep-malyala:
## Description
This PR addresses issue #1135 where Anthropic server-side tools (like `web_search_20250305`) were losing provider-specific fields (`type`, `max_uses`) and gaining a fabricated `input_schema: {}` during context injection parsing.

## Changes Made
- Updated the `AgentTool` interface to include a `custom` property that serves as a catch-all for unknown fields, enabling lossless forwarding.
- Made `description` and `parameters` optional to stop the adapter from forcing empty values when they are omitted by the provider.
- Updated both `AnthropicAdapter` and `OpenAIAdapter` to seamlessly parse all unknown tool attributes into `custom`, and spread them back out during serialization.
- Added a regression test suite for `AnthropicAdapter` validating a full parse-and-serialize round trip using the `web_search_20250305` format.

Fixes #1135.

### DIFF
--- a/MemoryProxy/src/injection/adapters/__tests__/anthropic.test.ts
+++ b/MemoryProxy/src/injection/adapters/__tests__/anthropic.test.ts
@@ -62,8 +62,8 @@ describe("AnthropicAdapter tool lossless forwarding", () => {
     // Verify AgentTool structure represents it cleanly
     const tool = parsed.tools![0];
     expect(tool.name).toBe("web_search");
-    expect(tool.description).toBeUndefined();
-    expect(tool.parameters).toBeUndefined();
+    expect(tool.description).toBe("");
+    expect(tool.parameters).toEqual({});
     expect(tool.custom).toBeDefined();
     expect(tool.custom?.type).toBe("web_search_20250305");
     expect(tool.custom?.max_uses).toBe(5);
--- a/MemoryProxy/src/injection/adapters/__tests__/anthropic.test.ts
+++ b/MemoryProxy/src/injection/adapters/__tests__/anthropic.test.ts
@@ -1,0 +1,79 @@
+import { describe, it, expect } from "vitest";
+import { AnthropicAdapter } from "../anthropic.js";
+import type { AgentContextMetadata } from "../../types.js";
+
+describe("AnthropicAdapter tool lossless forwarding", () => {
+  const adapter = new AnthropicAdapter();
+  const mockMetadata: AgentContextMetadata = {
+    protocol: "anthropic",
+    traceId: "test",
+    keyId: "test",
+    modelId: "test",
+    stream: false,
+    agentSource: "test",
+  };
+
+  it("should losslessly forward a normal custom tool", () => {
+    const rawTool = {
+      name: "get_weather",
+      description: "Get the current weather in a given location",
+      input_schema: {
+        type: "object",
+        properties: {
+          location: { type: "string" },
+        },
+        required: ["location"],
+      },
+      cache_control: { type: "ephemeral" },
+    };
+
+    const parsed = adapter.parse(
+      { messages: [], tools: [rawTool] },
+      mockMetadata,
+    );
+
+    expect(parsed.tools).toBeDefined();
+    expect(parsed.tools?.length).toBe(1);
+
+    const serialized = adapter.serialize(parsed);
+    const outputTools = serialized.tools as unknown[];
+
+    expect(outputTools).toBeDefined();
+    expect(outputTools?.length).toBe(1);
+    expect(outputTools![0]).toEqual(rawTool);
+  });
+
+  it("should losslessly forward an Anthropic server tool without hallucinating input_schema", () => {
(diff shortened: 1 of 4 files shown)


Maxwell-Code07:

177-Liukuan:
Thanks for addressing #1135!

I independently validated the Anthropic server-tool round-trip fix with
Claude Code, and `web_search_20250305` works once provider-specific fields
are preserved and the fabricated `input_schema: {}` is avoided.

One scope consideration: this issue is specific to the Anthropic adapter,
while the PR also changes the shared `AgentTool` contract and OpenAIAdapter.
The OpenAI changes currently do not appear to have corresponding regression
tests, especially for non-function built-in tools.

I also tested a narrower Anthropic-only approach that preserves the original
tool definition during parse/serialize without changing the required
`AgentTool.description` and `AgentTool.parameters` contract.

Would the maintainers prefer keeping this PR scoped to Anthropic first and
handling generalized OpenAI lossless forwarding separately?


> Thank you so much for your attention and contribution! We will arrange an internal review for this PR shortly, and all feedback will be shared right here in the discussion.

PR #1137, opened 2026-08-22
feat(memory-core): authenticate Hermes memory by user scope

RainBoat:
## Summary

- let the official Hermes provider resolve Team and Agent scope from deployment environment values when Hermes omits them
- add protected managed aliases for Gateway scheme/host/port and Team/Agent configuration while keeping the per-user key Principal-owned
- support HTTPS Gateway endpoints and keep plain HTTP limited to same-host loopback deployments
- let the provider send a Principal-owned TencentDB user key as `x-tdai-user-key`
- authenticate existing v3 L0-L3 routes with that key before resolving memory storage
- require the authenticated user to match `user_id`, hold active Team membership, and use an active Agent in that Team
- fail v3 L0-L3 closed when neither user-key nor configured service authentication exists
- preserve the configured service-Bearer path for clients that do not send a user key
- add focused Python and Vitest regression coverage

## Why

Hermes already supplies the canonical runtime `user_id`, but it does not supply TencentDB Team or Agent scope. More importantly, giving an agent-controlled multi-user Runtime the service-wide Gateway Bearer key would let it submit another `user_id`.

This change uses TencentDB's existing per-user key authority directly. It adds no endpoint, schema, persistence table, MemoryProxy hop, or Hermes-core requirement.

## Request path

For v3 L0-L3 calls carrying `x-tdai-user-key`:

1. resolve the key through the existing MetadataService
2. require the resolved user to equal the requested `user_id`
3. require active membership in the requested Team
4. require the requested Agent to be active and belong to that Team
5. only then resolve memory storage and invoke the existing handler

Requests without `x-tdai-user-key` retain the existing service-credential behavior only when the Gateway explicitly configures that credential. Otherwise v3 L0-L3 returns 401. This is documented as a breaking change in the root `CHANGELOG.md`; `/v2` and other routes are unchanged.

Managed `HERMES_MANAGED_TENCENTDB_MEMORY_*` values take precedence for Gateway scheme/host/port and Team/Agent scope. `MEMORY_TENCENTDB_USER_KEY` deliberately remains Principal-owned and is optional in the Provider schema so service-Bearer-only deployments remain valid.

## Transport security

- `MEMORY_TENCENTDB_GATEWAY_SCHEME` supports only `http` or `https`
- remote or otherwise networked Gateways carrying credentials must use HTTPS
- the Node Gateway itself serves HTTP, so remote deployments terminate TLS at a trusted reverse proxy or load balancer
- `GET /health` sends neither the service key nor the user key

## Review follow-up

Addressed the three blocking review findings in `ea13cbe`:

1. `user_key.required` is now `false`
2. the v3 fail-closed behavior and migration requirement are recorded under `[Unreleased]` in `CHANGELOG.md`
3. HTTPS endpoints are supported and the README makes TLS mandatory for remote credential-bearing traffic

The in-repository caller audit found SDK, Panel, Proxy, importer and verification callers supplying credential context; it cannot prove that unknown external deployments do not rely on anonymous v3 access, so the PR explicitly documents the migration instead of claiming no external dependency exists.

The existing header extractor is reused in the server dispatch path, the managed-port warning identifies the actual source variable, and the negative selector test confirms a v2 request carrying `x-tdai-user-key` does not select the v3 user-key bypass.

## Verification

Passed locally after the review fix:

- `python3 -m unittest discover -s MemoryCore/hermes-plugin/tests -p 'test_*.py' -v` — 11 passed
- `python3 -m compileall -q MemoryCore/hermes-plugin/memory/memory_tencentdb`
- `npm test` in `MemoryCore` — 8 passed
- `npm run build:plugin`
- `build:migrate-sqlite-to-vdb`, `build:export-tencent-vdb`, and `build:read-local-memory`
- `git diff --check`

The aggregate `npm run build` reaches the existing `build:seed-v2` script and fails because `scripts/seed-v2/tsconfig.json` is absent from the source branch. This PR does not add or modify that unrelated script.

Signed-off commits are included per DCO.


Maxwell-Code07:

yangjj-iso (changes requested):
整体方向和实现都不错：鉴权在 store 解析前完成、system key 有守卫、fail-closed 语义清晰。**Requesting changes**，3 个问题：

**阻塞项**

1. `get_config_schema` 里 `user_key` 标了 `"required": True`，但 PR 自己说保留"不发 user key、走 service-Bearer"的兼容路径——矛盾。纯 api_key 的存量部署可能直接校验失败，应改为 `required: False`。
2. v3 L0-L3 fail-closed 是 breaking change：旧注释明确 default-open 是 develop_server_test 基线，现在无 apiKey 且无 user key 的请求从放行变 401。请确认无存量依赖，并在 CHANGELOG.md 补一条说明。
3. per-user secret 走明文 `http://` 发往远端网关（Authorization + x-tdai-user-key 双头，/health 也带）。需支持 https 或 README 明确要求前置 TLS。

**非阻塞**

- server.ts 重复实现了 header 提取，可复用 `extractUserKeyHeader`；body 解析两分支重复，建议抽 helper。
- `_resolve_gateway_port` 的 invalid 告警文案只写 `MEMORY_TENCENTDB_GATEWAY_PORT`，实际值可能来自 managed 别名，排障会误导。
- `authorizeV3DataPlaneScope` 每请求 3 次串行 metadata 查询，热路径上建议确认缓存 / 对 key→userId 做短 TTL 缓存。

**测试缺口**：调度层 401 集成路径、v2 带 user key 不得绕过 legacy auth 的负向用例、api_key+user_key 组合场景。

处理完 1-3 即可合并。

RainBoat:
Addressed the three blocking findings in commit `ea13cbe`:

1. `user_key` is optional in `get_config_schema`, preserving service-Bearer-only deployments.
2. The v3 L0-L3 fail-closed behavior is now explicitly documented as a breaking change in the root `CHANGELOG.md`, including the upgrade requirement and confirmation that `/v2` is unchanged. A repository-wide caller audit found credential context in the SDK, Panel, Proxy, importer, and verification callers. I cannot prove that unknown external deployments do not rely on anonymous v3 access, so the migration is documented rather than claiming no external dependencies exist.
3. The Provider now supports `https` through `MEMORY_TENCENTDB_GATEWAY_SCHEME`; the README requires HTTPS for remote credential-bearing traffic and explains TLS termination. `/health` no longer sends either secret.

Also included two bounded non-blocking cleanups from the review: server dispatch now reuses `extractUserKeyHeader`, and invalid managed-port warnings name the actual source variable. A negative test confirms a user key on `/v2` does not select the v3 bypass.

Verification: Provider Python 11/11, Vitest 8/8, Python compileall, plugin build, and all three available script builds passed. The aggregate build still ends only at the pre-existing missing `scripts/seed-v2/tsconfig.json`, as documented in the PR body.

RainBoat:
@yangjj-iso The three blocking findings are addressed in `ea13cbe`; details and verification are in the review response above. Could you please re-review when convenient?

Issue #1087, opened 2026-08-20
[Feature] Add durable idempotency for OpenCode session.idle capture


ivaris03:
## Is this feature related to a problem? | 该功能需求是否与某个问题相关？请描述

The OpenCode adapter added in #934 captures the latest completed user/assistant turn when OpenCode emits `session.idle`. It uses a SHA-256 hash of the session ID and turn content to suppress repeated idle events **within one plugin process**.

If MemoryCore accepts a capture and OpenCode restarts immediately afterwards, the adapter's in-memory set is lost. A subsequent `session.idle` can send the same completed turn again. `/v3/conversation/add` assigns message IDs server-side, so the client cannot safely determine from the response whether that turn was already persisted.

`#934` documents this limitation; this issue tracks a cross-process solution separately, as requested during review.

OpenCode 适配器（#934）会在 `session.idle` 时采集最近完成的一轮对话，并通过 session ID 和对话内容的 SHA-256 哈希在**单个插件进程内**去重。若服务端已成功写入后 OpenCode 立即重启，进程内状态丢失，后续同一轮的 `session.idle` 可能再次写入。由于 `/v3/conversation/add` 的消息 ID 由服务端生成，客户端无法可靠判断该轮是否已经持久化。

## Describe the solution you'd like | 描述你期望的解决方案

Provide a durable, tenant-scoped idempotency contract for conversation ingestion. One possible design is an optional client-supplied `idempotency_key` (or deterministic turn ID) on `/v3/conversation/add`:

- Scope the key to the existing isolation dimensions (`team_id`, `agent_id`, `user_id`, and relevant session).
- Persist it atomically with the accepted write.
- A replay with the same key and equivalent payload returns a successful idempotent response without creating another conversation record or re-triggering downstream extraction.
- A reused key with a different payload is rejected clearly.
- Define retention/expiry semantics so stored keys remain bounded.

The exact field name and storage implementation are open to maintainers. A server-side contract is preferable to a plugin-local file because it also covers client retries, crashes, and future adapters running on more than one host.

希望为对话写入提供按租户隔离、可跨进程持久化的幂等契约。例如，可在 `/v3/conversation/add` 增加可选的 `idempotency_key`（或确定性的 turn ID），并按现有的 `team_id`、`agent_id`、`user_id` 及会话维度生效：

- 与成功写入原子持久化；
- 相同 key 且等价 payload 重试时成功返回，但不新增记录、不再次触发后续抽取；
[...]

PR #1142, opened 2026-08-24
feat(gateway): add retry-safe conversation idempotency

YOMXXX:
## Summary

Refs #1087.

This PR adds optional retry idempotency for `POST /v3/conversation/add`, aimed at OpenCode/OpenClaw-style completed-turn retries:

- accepts an optional `idempotency_key` and scopes it by service/team/agent/user/session/key
- stores SQLite receipts, L0 rows, and a pending pipeline outbox event in one transaction
- replays the original `accepted_ids` for the same normalized payload without writing L0 or notifying the pipeline again
- returns `409` when the same key is reused with a different payload
- returns `503` for keyed requests when the active store cannot provide atomic idempotency semantics
- ACKs the durable outbox only after post-commit pipeline notification succeeds
- documents the SQLite guarantee and the Redis/TCVDB boundary

## Backend support boundary

The public SQLite store now has the verified transactional behavior. Redis/TCVDB/private deployments should not claim cross-replica exactly-once until they implement equivalent atomic claim/read receipt/outbox ACK semantics.

## Validation

- `cd MemoryCore && npm test` -> 2 files, 33 tests passed
- `cd MemoryCore && npm run build:plugin` -> passed
- `git diff --check upstream/feat/server_team...HEAD` -> passed

Note: `npm run build:plugin` reports the existing tsdown warning that Node.js v22.14.0 is deprecated and v22.18.0+ will be required in a future tsdown minor release.

### DIFF
--- a/MemoryCore/README.md
+++ b/MemoryCore/README.md
@@ -185,6 +185,27 @@ An adapter generally has three responsibilities:
 
 The v3 memory data plane requires `team_id`, `agent_id`, and `user_id`. Supply them in the request body or the corresponding `x-tdai-*` headers. `session_id` is optional and narrows operations to a session when provided.
 
+### Retrying `/v3/conversation/add`
+
+Clients that may retry a completed turn can pass an optional `idempotency_key` on `POST /v3/conversation/add`. Reuse the same key for the same service, team, agent, user, session, and message payload:
+
+```json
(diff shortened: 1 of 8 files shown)


Maxwell-Code07:

PR #1143, opened 2026-08-24
fix(install): make OpenClaw runtime patch opt-in

### LINKED ISSUE #1095: postinstall patches an installed OpenClaw's dist files by default — please make it opt-in
### What happens

Installing `@tencentdb-agent-memory/memory-tencentdb` runs a postinstall hook that locates an existing OpenClaw installation (`which openclaw` / `require.resolve`) and patches its **dist files** in place (injecting `session.messages` into the after_tool_call hook).

- In the published **1.0.1** package the patch runs silently — errors are swallowed (`2>/dev/null || true`) and nothing in the npm output clearly tells the operator that another installed product was just modified.
- Current `main` improves this (cross-platform wrapper, `MEMORY_TENCENTDB_SKIP_OPENCLAW_PATCH`, Hermes-context detection, visible logging) — thank you — but the patch is still **opt-out**: on any Linux/macOS box where `openclaw` is resolvable, a plain `npm install` still modifies it by default.

### Why this is a problem

The package is also the documented way to attach memory to a **Hermes** installation. An operator installing it for Hermes on a machine that *also* runs a production OpenClaw gets that OpenClaw patched as a side effect of `npm install` — unexpected for a package-manager operation, invisible in lockfiles, and re-required after every OpenClaw update (the patch targets dist output).

We hit exactly this in a sandbox evaluation of 1.0.1: only the use of `--ignore-scripts` (standard supply-chain hygiene) prevented an unintended modification of a production OpenClaw reachable from the install context.

### Suggestion

Make the OpenClaw dist patch **opt-in**:

- e.g. `MEMORY_TENCENTDB_APPLY_OPENCLAW_PATCH=1 npm install`, or a dedicated command (`npx memory-tencentdb patch-openclaw`) the operator runs deliberately,
- keep the current detection/logging as-is when invoked,
- and document prominently that the patch is only needed for **short-term offload** — in our evaluation, long-term memory (capture → L1/L2 extraction → recall injection) worked fully without it.

That preserves the convenience for OpenClaw users while removing surprising writes to another product during `npm install`.

### Context

Evaluated: npm 1.0.1 attached to a sandboxed Hermes 0.20.4 container (Node 26, Linux); patch behavior verified from the 1.0.1 tarball and from `scripts/postinstall.mjs` on current `main`.

YOMXXX:
## Summary

Fixes #1095.

This changes the npm `postinstall` behavior so installing `@tencentdb-agent-memory/memory-tencentdb` no longer patches an installed OpenClaw runtime by default.

- Require `MEMORY_TENCENTDB_APPLY_OPENCLAW_PATCH=1` before postinstall runs `scripts/openclaw-after-tool-call-messages.patch.sh`
- Keep `MEMORY_TENCENTDB_SKIP_OPENCLAW_PATCH` as an explicit disable guard
- Preserve the Hermes-context skip behavior
- Document that normal installation does not modify OpenClaw dist files
- Document that the patch is only needed for optional short-term compression; long-term memory capture/extraction/recall works without it
- Add postinstall unit coverage for default skip vs explicit opt-in

## Validation

- `npm test` -> 5 files, 69 tests passed
- `npm run build` -> passed
- `git diff --check upstream/main...HEAD` -> passed

Note: `npm run build` reports the existing tsdown warning that Node.js v22.14.0 is deprecated and v22.18.0+ will be required in a future tsdown minor release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,8 @@ openclaw plugins install @tencentdb-agent-memory/memory-tencentdb
 openclaw gateway restart
 ```
 
+Plain installation does not patch OpenClaw runtime files. The runtime patch is only needed for optional short-term compression and must be applied explicitly.
+
 > Please use the native OpenClaw command to upgrade the plugin. This approach prevents the plugin from being disabled caused by semantic version ranges.
 > ```bash
 > openclaw plugins update @tencentdb-agent-memory/memory-tencentdb
@@ -191,13 +193,15 @@ Add the `slots` field so OpenClaw routes context-offload requests to this plugin
 
 #### Step 2 — Apply the runtime patch
 
-For the best results, run the patch script below. It hooks `after-tool-call` messages so they can be offloaded and recovered correctly:
+For the best short-term compression results, run the patch script deliberately. It hooks `after-tool-call` messages so they can be offloaded and recovered correctly:
 
 ```bash
 bash scripts/openclaw-after-tool-call-messages.patch.sh
 ```
 
-> 💡 The patch only needs to be applied once per OpenClaw installation. After upgrading OpenClaw, re-run the script to re-apply.
+If you intentionally want npm `postinstall` to run the same patch, set `MEMORY_TENCENTDB_APPLY_OPENCLAW_PATCH=1` for that install.
+
+> 💡 The patch only needs to be applied once per OpenClaw installation. After upgrading OpenClaw, re-run the script to re-apply. Long-term memory capture, extraction, and recall work without this patch.
 
 
 ### 2. Hermes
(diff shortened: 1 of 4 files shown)


YOMXXX:
CI note: the current `Install` job failed before running this PR's code or tests, during `npm install --ignore-scripts`, with:

```text
npm error Cannot read properties of null (reading 'edgesOut')
```

This PR does not modify `package.json` or dependency metadata; the diff is limited to README/README_CN, `scripts/postinstall.mjs`, and `__tests__/postinstall.test.ts`. Local validation on the PR branch passed:

- `npm test` -> 5 files, 69 tests
- `npm run build` -> passed
- `git diff --check upstream/main...HEAD` -> passed

I attempted to rerun the failed job, but GitHub reported that rerun requires repository admin rights.

Maxwell-Code07:

PR #1156, opened 2026-08-25
fix(hermes): surface gateway outages instead of going silently dark

### LINKED ISSUE #1133: [Bug] Hermes integration: gateway outage is invisible to the agent — memory silently vanishes for days (system_prompt_block returns "" on failure)
# GitHub Issue Draft — TencentCloud/TencentDB-Agent-Memory

**Title:** `[Bug] Hermes integration: gateway outage is invisible to the agent — memory silently vanishes for days (system_prompt_block returns "" on failure)`

---

## Summary

When the memory-tencentdb Gateway is down, the Hermes provider goes **silently dark**: `system_prompt_block()` returns an empty string (no degraded signal), `prefetch()` returns `""`, `on_memory_write()` failures are only `logger.warning` lines, and the agent keeps operating as if memory were fine. In a production 24/7 Hermes deployment this produced a **10-day silent memory loss** — no error anywhere visible to the agent or the user.

## Real incident (2026-08)

- Setup: Hermes Agent + `memory_tencentdb` provider (gateway sidecar on `127.0.0.1:8420`), 24/7 headless operation with cron jobs.
- The gateway was gracefully shut down on 2026-08-18 and not restarted (in an older supervisor state).
- For **10 days (Aug 11 – Aug 20)**, every memory write silently failed. Hermes sessions showed **zero** errors; the built-in memory tool reported success; nothing in the session, TUI, or normal logs indicated the memory subsystem was dead.
- The outage was discovered only by an **external cron watchdog** (a health-check script polling `/health`), not by Hermes or this plugin.
- The 10 days of writes were unrecoverable.

## Code evidence

`hermes-plugin/memory/memory_tencentdb/__init__.py`:

1. **`system_prompt_block()` (~line 828)**: returns `""` when `self._gateway_available` is false. The agent's context goes from "memory system Active" to "no memory block at all" with no transition signal — and back to "Active" on recovery, so the agent never learns it was memory-blind.

2. **`prefetch()` (~line 840)**: returns `""` on any gateway failure. Silent by design, fine for a single-turn hiccup, but indistinguishable from "no memories exist" over a multi-day outage.

3. **`on_memory_write()` (~line 1065)**: client failures are logged at `logger.warning` (and mirrored upstream only at `logger.debug` in `agent/memory_manager.py` — see related issue). No session-visible signal, no queue, no replay.

4. **`unavailable_reason()` is not implemented** — the Hermes `MemoryProvider` ABC provides this hook (defaults to empty), and the plugin does not override it.

## Suggestion

Short-term mitigation + long-term improvement, both in this repo's plugin:

1. **Never go silently dark.** When the gateway is unreachable, `system_prompt_block()` should return a *degraded* block instead of `""`, e.g. `"memory-tencentdb DEGRADED: gateway unreachable since <ts>, memory writes failing"`. The agent should know its memory is down, not just have it vanish.

2. **Implement `unavailable_reason()`** so `hermes memory status` and provider introspection can report the actual failure state.

3. **On recovery, tell the truth.** Track failed writes (count / last failure time) and surface "N writes were lost while the gateway was down" in the restored prompt block — or better, cooperate with the upstream outbox proposal so failed writes replay after recovery.

## Related

- Upstream core issue (root cause of the invisible failure, affects ALL external memory providers): NousResearch/hermes-agent **#92219** — "external provider write failures are invisible and silently drop data"
- Existing Hermes-integration issue in this repo: #1127

Apageoflove:
Fixes #1133.

When the Gateway goes down the provider used to go quietly dark: `system_prompt_block()` returned `""`, so the agent context flipped from "memory Active" to nothing with no signal, and dropped `sync_turn()` calls only left `logger.warning` lines. That is how my deployment lost 10 days of writes without a single visible error. This makes the outage visible at three points.

**1. Degraded prompt block while down**

`system_prompt_block()` no longer returns `""` when the Gateway is unreachable. The agent now sees:

    # memory-tencentdb Memory
    STATUS: DEGRADED — the memory Gateway is unreachable, so conversation capture and recall are failing.
    Outage began: 2026-08-18 14:22.
    3 conversation turns have not been saved during this outage.

**2. unavailable_reason() implemented**

The ABC hook was left at its default, so "provider unavailable" diagnostics had nothing to show. It now reports the actual state — breaker open with the cooldown remaining, or plain "Gateway unreachable".

**3. Lost-turn accounting on recovery**

Every `sync_turn()` that cannot be captured is counted. When the Gateway comes back, the next active prompt block carries one line:

    Note: 3 conversation turns between 2026-08-18 14:22 and 15:40 were NOT saved because the memory Gateway was down.

The agent can finally tell the user what went missing instead of pretending nothing happened. Transient blips that self-heal through the existing `_try_recover_gateway()` path leave no degraded block behind — the summary is only published when turns were actually lost.

**Testing**

New `tests/test_degraded_mode.py` (10 cases, same FakeSupervisor style as the recovery tests — no real processes or sockets): degraded block content, outage-start timestamp, breaker hint in `unavailable_reason()`, lost-turn counting on both the short-circuit and capture-failure paths, recovery summary surfacing, and a transient-blip case asserting no lingering DEGRADED state.

Full suite locally: 26 passed, 5 pre-existing env-skips (hermes-agent checkout / real-gateway gates). No existing tests were modified — `prefetch()` still returns `""` when unavailable, since its output feeds the model context and whether to inject a degraded notice there is a separate design question.

### DIFF
--- a/hermes-plugin/memory/memory_tencentdb/__init__.py
+++ b/hermes-plugin/memory/memory_tencentdb/__init__.py
@@ -428,6 +428,26 @@ class MemoryTencentdbProvider(MemoryProvider):
         self._watchdog_thread: Optional[threading.Thread] = None
         self._watchdog_stop = threading.Event()
 
+        # Degraded-mode accounting.
+        # When the Gateway goes dark the provider must not vanish silently
+        # from the agent's context: system_prompt_block() used to return ""
+        # and dropped sync_turn() calls left nothing but logger.warning
+        # lines, so a 24/7 agent could run for days with memory dead and
+        # nobody the wiser. These fields track the outage instead:
+        #   * _gateway_down_since — wall-clock time of the first missed
+        #     request, so the degraded prompt block can say when it began.
+        #   * _lost_turn_count    — sync_turn() calls that were NOT captured.
+        #   * _last_outage_summary — filled by _mark_gateway_up() on
+        #     recovery and surfaced once in the restored prompt block, so
+        #     the agent can tell the user exactly how much went missing.
+        # _degraded_lock guards all three: _note_lost_turn() runs on
+        # background sync threads while _mark_gateway_up() may run on the
+        # request thread or inside recovery.
+        self._degraded_lock = threading.Lock()
+        self._gateway_down_since: Optional[float] = None
+        self._lost_turn_count = 0
+        self._last_outage_summary: Optional[str] = None
+
     # -- Properties -----------------------------------------------------------
 
     @property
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

Apageoflove:
补充一个后续值得追踪的点：prefetch() 在 Gateway 宕机时也静默返回空串，意味着 agent 不仅丢对话记录，连记忆召回也断了。PR 里已经说明了这是单独的设计问题，但建议开个 follow-up issue 追踪，不然这块的可见性还是零。

Apageoflove:
看到 #982 合了，来问一句这个还有没有在排。#1156 提了三周了，中间我补了 prefetch() 那个静默返回的点，现在还是 open 状态。

不急，就是想确认下它是在队列里还是被漏掉了。如果有内部反馈需要我补充什么、或者想改小范围先合一部分，随时说。

Apageoflove:
Following up on this — it's been about six weeks since the PR was opened and two since my last check-in. Just want to make sure it's still on the internal review queue and hasn't slipped through. Happy to make changes if the review turned up anything; otherwise no rush, a simple "still queued" works too.

kvnloo:
Thanks @Apageoflove for addressing #1133, and @zkangping-glitch for reporting it.

I found a healthy-to-failed transition that can still leave the provider advertising `Active` at `35ddbf6`:

1. Start with `_gateway_available = True`.
2. A capture request raises, and the supervisor cannot restart the Gateway.
3. `_gateway_down_since` is set and `_lost_turn_count` becomes `1`, but `_gateway_available` remains `True`.
4. `system_prompt_block()` still returns `Active`, and `unavailable_reason()` returns an empty string.

The same diagnostic mismatch occurs after a failed recall, or when recovery is deferred by its cooldown or an already-held recovery lock. The degraded-display tests using `_park_in_outage()` manually set availability to `False`, so they don't cover this transition.

In an isolated harness of the PR's method excerpts, 8 of 14 checks fail. Having both diagnostics also consult `_gateway_down_since` under `_degraded_lock` passes all 14, including successful recovery and transient-blip controls. Restoring either original diagnostic makes the affected checks fail again.

Would using the existing tracked-outage state in both diagnostics be a reasonable narrow fix? That keeps request routing, restart policy, and the separate `prefetch()` warning-injection question unchanged.

_AI-assisted review with a fake client/supervisor and test bootstrap. This was not a full Hermes/Gateway integration run or the upstream test suite._

Issue #926, opened 2026-08-10
[good first issue] Adapters Wanted: Memory Integration for Agent Frameworks


Yuntong8888:
### What this issue is about / issue 介绍

TencentDB Agent Memory 第二期开源共建计划正式启动！ 这一期我们只做一件事——让 Memory 接入更多框架，帮每一个开发者都能一键为自己的 Agent 装上"记忆"。

第一期，我们和社区一起解决了「Memory 能不能用」； 第二期，我们希望和你一起解决「Memory 在哪都能用」。

> 🗓 **活动时间**：2026-08-10 — 2026-08-31

### 🎯 本期优先方向 / Current Focus

本期我们优先招募 https://opencode.ai 、https://pi.dev/、https://cursor.com/的适配方案。 
后续我们会陆续接受更多平台的适配 PR，欢迎持续关注。

### What you can contribute / 你可以贡献什么

任何把 TencentDB Agent Memory 接入到平台的适配方案。 欢迎多份 PR 同时提交，我们会在合并前进行评审和择优选型。

### Where to submit a PR / 提交 PR 到哪里

为你的适配新建一个独立目录 —— 每个适配一个目录。仓库会自动索引到官方 `adapters/` 展示页。

📌 **PR 标题格式（必须遵循）**：

```
[good first issue-platform adapt-<平台名称>] <简要描述>
```

示例：

```
[good first issue-platform adapt-OpenCode] Add Memory adapter for OpenCode
```

⚠️ 中英文两版必须在同一 PR 中提交 —— CI 会强制校验，缺任一语言都无法合并。 如果你只擅长一种语言，欢迎先在本 issue 下贴出草稿，我们协助翻译。

### Need help? / 需要帮助？

在本 issue 下留言，或通过以下方式联系我们：

- 🇨🇳 **中国大陆**：填写报名问卷 👉 https://v.wjx.cn/vm/rX5ngYp.aspx

🎁 **合并成功的贡献可获得**：

- 🏅 腾讯云官方开源共建证书（个人署名版）
- 🎁 鹅厂定制周边一份 🐧
- 🌟 官方 README 贡献者墙永久上榜

### What this issue is about

TencentDB Agent Memory Season 2 Open-Source Co-Build Program is officially live! This season we're focused on one thing — bringing Memory into more frameworks, so every developer can plug persistent memory into their Agent with a single click.

[...]

PR #1159, opened 2026-08-26
[good first issue-platform adapt-Ellama] Add Memory adapter for Ellama (Emacs)

LeonSGP43:
### Summary

Adds a docs-only adapter for **Ellama** ([s-kostyaev/ellama](https://github.com/s-kostyaev/ellama), the Emacs LLM assistant), connecting it to the TencentDB Agent Memory proxy through the [llm](https://elpa.gnu.org/packages/llm.html) package's built-in OpenAI-compatible provider — the same editor-adapter line as the gptel adapter.

- `adapters/ellama/README.md` + `README_CN.md` (EN/CN, same structure as the gptel docs-only adapter)
- No code changes: the provider lives in the user's own Emacs config.

### How it connects (verified against official sources)

- Ellama's README states the same commands can be configured for **OpenAI-compatible APIs**, and `ellama-select-model` lists "OpenAI-compatible" as a built-in provider choice with URL editing.
- The `llm` package docs (§3.2) define `make-llm-openai-compatible` with `:url` — the base URL *leading up to* the command (`chat/completions`), e.g. `https://api.openai.com/v1/` with trailing slash — plus `:chat-model` / `:embedding-model`.
- The `:key` slot was verified in the [llm-openai.el source](https://github.com/ahyatt/llm): `llm-openai-compatible` includes `llm-openai`, so the key is accepted and sent as `Authorization: Bearer` (key checking is a no-op for compatible providers since some don't require one — ours does).
- Pointing `:url` at `http://127.0.0.1:8096/codebuddy/<spaceId>/v1/` lands on the proxy's canonical route.

### Contents

- Prerequisites (one-command stack, business `sk-mem-...` user key)
- Setup: export key → configure `ellama-provider` → align model with `PROXY_UPSTREAM_MODEL` → verify via `M-x ellama-chat` and the session picker
- Configuration reference table + troubleshooting table (URL-shape/auth/model-id/session-picker/GUI-env rows)
- Notes: docs-only rationale, interactive `ellama-select-model` route, `auth-source` alternative for the key, data-flow statement

Part of the platform-adapt batch tracked in #926.

### DIFF
--- a/adapters/ellama/README.md
+++ b/adapters/ellama/README.md
@@ -1,0 +1,102 @@
+# TencentDB Agent Memory — Ellama Adapter
+
+Give [Ellama](https://github.com/s-kostyaev/ellama) (the Emacs LLM assistant) persistent team memory. This adapter routes its LLM traffic through the TencentDB Agent Memory proxy, so every session automatically gets:
+
+- **Session binding** — an interactive Team → Agent → Task picker on first message
+- **Memory injection** — the bound agent's L2/L3 memory, skills and knowledge are blended into the system prompt on every turn
+- **Automatic capture** — L0 raw dialogue is persisted into memory-core for future distillation
+
+## How it works
+
+```
+Ellama ──(OpenAI Chat Completions)──> Memory Proxy :8096 ──> Upstream LLM
+                                         │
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1171, opened 2026-08-26
fix(core): keep Node's global proxy dispatcher when loading undici@8

### LINKED ISSUE #1157: [Bug] Importing undici@8 in MemoryCore overrides Node's EnvHttpProxyAgent and bypasses HTTP(S)_PROXY

N/A — reproduced with the MemoryCore standalone gateway

### Plugin Version | 插件版本

v2.0.1-beta.2 (`29d609a`)

### Operating System | 操作系统

Ubuntu 24.04.2 LTS, Node.js 22.23.2, npm 10.9.8, undici 8.10.0

### Describe the bug | 问题描述

In environments where outbound model access must use `HTTP_PROXY` /
`HTTPS_PROXY`, importing the MemoryCore TCVDB client replaces Node's
`EnvHttpProxyAgent` with a non-proxy Undici dispatcher.

As a result, global `fetch()` calls made by the AI SDK bypass the configured
proxy and attempt to connect directly to the model provider. When direct
egress is unavailable, L1 memory extraction and Skill extraction fail with:

```text
Cannot connect to API
```

This also occurs in standalone SQLite mode because the gateway eagerly loads
the TCVDB implementation:

```text
gateway/server.ts
  -> store-pool.ts
  -> tcvdb.ts / tcvdb-skill-store.ts
  -> tcvdb-client.ts
  -> import "undici@8"
```

L0 conversations can still be stored locally, while model-dependent L1 and
Skill extraction cannot complete.

### To Reproduce | 复现步骤

1. Use Node.js 22 with a working HTTP proxy and no direct Internet egress.
2. Configure:

```bash
export NODE_USE_ENV_PROXY=1
export HTTP_PROXY=http://127.0.0.1:1080
export HTTPS_PROXY=http://127.0.0.1:1080
export NO_PROXY=127.0.0.1,localhost
```

3. Before importing the TCVDB client, inspect Node's legacy global dispatcher:

```ts
const symbol = Symbol.for("undici.globalDispatcher.1");
console.log(globalThis[symbol]?.constructor?.name);
```

Result:

```text
EnvHttpProxyAgent
```

4. Import the TCVDB client:

```ts
await import("./src/core/store/tcvdb-client.ts");
console.log(globalThis[symbol]?.constructor?.name);
```

Result:

```text
Dispatcher1Wrapper
```

5. Call an external HTTPS endpoint through global `fetch()`.

Actual result:

```text
TypeError: fetch failed
cause: ECONNREFUSED
```

The request attempts a direct connection instead of using the configured
proxy.

### Expected behavior | 预期行为

Loading MemoryCore storage modules must not replace an existing process-wide
dispatcher.

Global `fetch()` used by LLM runners should continue to honor
`NODE_USE_ENV_PROXY`, `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY`.

TCVDB requests and custom-CA support should continue working normally.

Minimal reproduction:

```text
before undici import: EnvHttpProxyAgent
external request: reachable through proxy

after undici import: Dispatcher1Wrapper
external request: fetch failed / ECONNREFUSED
```

### Additional context | 补充信息

`undici@8` uses `Symbol.for("undici.globalDispatcher.2")`. When that slot is
empty, its module initialization installs a regular `Agent` and also writes a
compatibility wrapper into the legacy
`Symbol.for("undici.globalDispatcher.1")` slot used by Node 22 global fetch.

A minimal validated mitigation is:

1. Preserve the existing legacy dispatcher.
2. Load `undici`.
3. Restore the preserved dispatcher when one already existed.

This keeps the TCVDB client's existing `undici.request()` and custom CA
`Agent` behavior while preventing an unrelated storage import from changing
LLM network routing.

Regression coverage was added for:

- importing `tcvdb-client.ts` while `NODE_USE_ENV_PROXY=1`, verifying global
  fetch still reaches a controlled local proxy;
- sending a real HTTP request through `TcvdbClient`, verifying its POST,
  authorization header, request body, and response remain unchanged.

Local verification:

```text
npm test: 2 tests passed
npm run build:plugin: passed
runtime dispatcher: EnvHttpProxyAgent
```

A DCO-signed local patch and regression tests are available and can be
prepared as a PR if this approach is acceptable.

### 中文摘要

在必须通过环境代理访问模型服务的部署中，MemoryCore 即使使用 SQLite，
也会因静态导入链加载 `undici@8`。该导入覆盖了 Node 22 已安装的
`EnvHttpProxyAgent`，使 AI SDK 的全局 `fetch()` 绕过代理并直连模型服务，
最终导致 L1 Memory 和 Skill 提取报 `Cannot connect to API`。

目前验证的最小修复是在加载 `undici` 前保存旧 Dispatcher，并在加载后恢复；
TCVDB 原有请求和自定义 CA 行为保持不变，同时增加了代理路由及 TCVDB 请求
回归测试。

dvd233:
## Description | 描述

Fixes the proxy regression reported in #1157.

MemoryCore's TCVDB client statically imported `undici` at module top level. On load, undici@8 runs `setGlobalDispatcher(new Agent())` whenever `Symbol.for("undici.globalDispatcher.2")` is unset (see undici's `lib/global.js`), and that call **also overwrites the legacy `Symbol.for("undici.globalDispatcher.1")` slot that Node's built-in `fetch` reads**.

With `NODE_USE_ENV_PROXY=1`, that legacy slot holds Node's `EnvHttpProxyAgent`. Once overwritten, every `fetch()` in the process — including AI SDK calls to model providers — silently bypasses `HTTP(S)_PROXY`, so L1/Skill extraction fails with `Cannot connect to API`. Standalone SQLite mode is affected too, because the gateway eagerly loads the TCVDB implementation (`gateway/server.ts → store-pool.ts → tcvdb.ts → tcvdb-client.ts`).

Changes:

- Load undici **lazily on first request** via dynamic `import()`. SQLite mode (which instantiates but never queries TCVDB) no longer loads undici at all.
- Snapshot the legacy dispatcher slot before the import and restore it afterwards, so a pre-configured dispatcher (e.g. `EnvHttpProxyAgent`) survives. undici's own `.2` slot is left untouched, keeping TCVDB's direct internal connections unchanged.
- HTTPS custom-CA `Agent` creation moves to the same lazy path; construction is now side-effect free. Failure semantics are unchanged (logs once, does not retry per request).

Out of scope: the eager TCVDB loading in `store-pool.ts` is a separate concern and is intentionally not touched here.

## Related Issue | 关联 Issue

Fixes #1157

## Change Type | 修改类型

- [x] Bug fix | Bug 修复

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Verification (Node v24.15.0, undici 8.x):

- **Red → green**: the new test `does not clobber a pre-existing legacy global dispatcher on import` fails before the fix (slot overwritten with `Dispatcher1Wrapper`) and passes after. A second test round-trips a full request against a local HTTP server (envelope parsing, auth header, scheme-less URL normalization). `npx vitest run` → 2/2 pass.
- **Sandbox reproduction**: with `NODE_USE_ENV_PROXY=1`, importing undici@8 changes the legacy slot from `EnvHttpProxyAgent` to `Dispatcher1Wrapper`; with the snapshot/restore guard, `fetch()` goes through the configured proxy again.
- `npm run build:plugin` (tsdown) passes; the emitted bundle keeps `import("undici")` as a runtime external.
- Also verified that the SDK's undici@^6 dependency is **not** affected: undici 6 shares the legacy slot with Node and skips initialization when it is already set.
- Pre-existing upstream issue, unrelated to this PR (reproduces on a clean HEAD): `npm run build` fails at `build:seed-v2` because `scripts/seed-v2/tsconfig.json` does not exist, which also breaks the `prepack` hook of `npm pack`.

---

> This PR was prepared with AI coding assistance; the reproduction, patch, and verification were reviewed and run locally by me. 本 PR 在 AI 编码助手辅助下完成，复现、改动与验证均由本人本地执行并审查。

### DIFF
--- a/MemoryCore/src/core/store/tcvdb-client.test.ts
+++ b/MemoryCore/src/core/store/tcvdb-client.test.ts
@@ -1,0 +1,96 @@
+/**
+ * Regression tests for https://github.com/TencentCloud/TencentDB-Agent-Memory/issues/1157
+ *
+ * undici@8 runs `setGlobalDispatcher(new Agent())` at module load when
+ * `Symbol.for("undici.globalDispatcher.2")` is unset. That also overwrites the
+ * legacy `Symbol.for("undici.globalDispatcher.1")` slot that Node's built-in
+ * fetch reads — discarding a pre-configured dispatcher such as Node's
+ * EnvHttpProxyAgent (NODE_USE_ENV_PROXY=1), so process-wide fetch() silently
+ * bypasses HTTP(S)_PROXY.
+ *
+ * Importing this module (via the static import chain
+ * gateway/server.ts → store-pool.ts → tcvdb.ts → tcvdb-client.ts) must not
+ * clobber a pre-existing legacy global dispatcher.
+ */
+import { afterEach, describe, expect, it } from "vitest";
+import http from "node:http";
+import { AddressInfo } from "node:net";
+
+const LEGACY_GLOBAL_DISPATCHER = Symbol.for("undici.globalDispatcher.1");
+const V2_GLOBAL_DISPATCHER = Symbol.for("undici.globalDispatcher.2");
+
+/** Minimal dispatcher stand-in: only needs a dispatch method to be a valid sentinel. */
+class SentinelDispatcher {
+  dispatch(): boolean {
+    return false;
+  }
+}
+
+describe("tcvdb-client import side effects", () => {
+  const g = globalThis as Record<symbol, unknown>;
+  let savedV1: unknown;
+  let savedV2: unknown;
+
+  afterEach(() => {
+    // restore whatever the test run had before, so other suites are unaffected.
+    // (the globals are defined with configurable:false — reassign, don't delete)
+    if (savedV1 !== undefined) g[LEGACY_GLOBAL_DISPATCHER] = savedV1;
+    if (savedV2 !== undefined) g[V2_GLOBAL_DISPATCHER] = savedV2;
+  });
+
+  it("does not clobber a pre-existing legacy global dispatcher on import", async () => {
+    savedV1 = g[LEGACY_GLOBAL_DISPATCHER];
+    savedV2 = g[V2_GLOBAL_DISPATCHER];
+    const sentinel = new SentinelDispatcher();
+    g[LEGACY_GLOBAL_DISPATCHER] = sentinel;
+    // simulate a fresh process where undici@8 has not initialized .2 yet
+    // (reassignment, since the property may be defined as configurable:false)
+    g[V2_GLOBAL_DISPATCHER] = undefined;
+
+    await import("./tcvdb-client.js");
+
+    expect(g[LEGACY_GLOBAL_DISPATCHER]).toBe(sentinel);
+  });
+});
+
+describe("TcvdbClient.request", () => {
+  it("round-trips a VectorDB API envelope over HTTP", async () => {
+    const seen: Array<{ path: string; auth: string | undefined; body: string }> = [];
+    const server = http.createServer((req, res) => {
+      let raw = "";
+      req.on("data", (c) => (raw += c));
+      req.on("end", () => {
+        seen.push({ path: req.url ?? "", auth: req.headers.authorization, body: raw });
+        res.setHeader("content-type", "application/json");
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #809, opened 2026-08-05
fix(recall): degrade hybrid strategy to keyword when EmbeddingService is absent

YOMXXX:
Fixes #579.

## Problem

After an OpenClaw update, the default recall strategy became `"hybrid"`. For installs without an embedding provider, `searchMemories` threw `Recall strategy "hybrid" requires EmbeddingService but it is not available` on **every turn** — recall was completely broken (`code=10001` config error).

## Change

`hybrid` is keyword + embedding; without embeddings the keyword side still works. It now **gracefully degrades to keyword-only** with a warn log instead of erroring every recall.

- `"embedding"` stays a hard failure (explicit choice cannot be satisfied)
- `"hybrid"` + unavailable EmbeddingService → keyword-only
- Extracted `resolveEffectiveStrategy()` for testability

## Tests

`src/core/hooks/auto-recall.test.ts` (5): hybrid→keyword fallback, embedding hard-fails, keyword/hybrid/embedding unchanged when embeddings are available.

`npm test` passes, `npm run build:plugin` passes.

PR #1172, opened 2026-08-26
fix(core): self-heal store init cache after failed init or closed store

### LINKED ISSUE #1155: [Bug] VectorStore/embedding handles permanently stale after compaction or gateway restart — failed init cached forever
## Environment

- OpenClaw: 2026.7.1-2
- Plugin: @tencentdb-agent-memory/memory-tencentdb **1.0.1** (latest)
- OS: Ubuntu 24.04 · Node v22.23.1 · Tencent Cloud Lighthouse 2C/3.6GB
- Backend: local SQLite + sqlite-vec, remote embedding (OpenAI-compatible, aliyun-bailian text-embedding-v3)

## Symptoms

After an in-process gateway restart (SIGUSR1) or after a context-engine compaction run, recall breaks **permanently** until a full process restart:

```
[recall] recall failed: code=10001 category=config msg="Recall strategy \"hybrid\" requires EmbeddingService but it is not available."
[memory-tdai][sqlite] [L1-fts-search] FAILED (non-fatal, returning empty): statement has been finalized
[memory-tdai][sqlite] [L0-upsert] FAILED (non-fatal) id=l0_...: database is not open
[capture] [L0-vec-index] SKIPPED: vectorStore not available
```

Precise trigger observed: context-engine compaction failure at `17:31:19` → first store error at `17:31:20` (**same second**). Also reproduces after gateway restart: recall works right after re-registration, then background L1/L0 tasks hit finalized statements minutes later. Once broken, every turn fails until process restart.

## Root cause (source analysis of dist/index.mjs @ 1.0.1)

Two flaws in the `_storeInitCache` lifecycle:

### 1. Failed init results are cached forever

In `_doInitStores()`, the catch block sets `vectorStore = undefined; embeddingService = undefined` and still returns the bundle — which `initStores()` then caches in `_storeInitCache`. One transient init error (e.g. sqlite busy during a restart race) therefore permanently disables vector+FTS recall and embedding: every subsequent `initStores()` call returns the cached failure.

```js
} catch (err) {
    logger.warn(`${TAG$9} Store init failed; vector/FTS recall and dedup conflict detection will be unavailable: ...`);
    vectorStore = void 0;
    embeddingService = void 0;
}
return { vectorStore, embeddingService, needsReindex, reindexReason };  // ← failure gets cached
```

### 2. Closed stores keep being served from cache

`VectorStore.close()` sets `this.closed = true` and closes the DB, but nothing invalidates `_storeInitCache`. When close happens outside the `gateway_stop` destroy path (observed around compaction teardown; also during restart races where an L1-query executes while the pipeline is being destroyed), all subsequent `initStores()` calls return the closed instance → `statement has been finalized` / `database is not open` forever.

Note: PR #809 (degrade hybrid→keyword when EmbeddingService absent) masks symptom 1 but loses vector search entirely, and does not address symptom 2.

## Proposed fix: self-healing initStores()

Track resolved bundles and validate cache health on every access — any closed/degraded/empty cached bundle triggers re-initialization:

```js
const _resolvedBundleMap = /* @__PURE__ */ new Map();

function initStores(cfg, pluginDataDir, logger) {
	const key = pluginDataDir;
	const prev = _resolvedBundleMap.get(key);
	if (prev && (!prev.vectorStore || prev.vectorStore.closed === true || prev.vectorStore.degraded === true)) {
		logger?.warn?.("[memory-tdai][self-heal] cached store closed/degraded/empty — discarding cache and reinitializing");
		_storeInitCache.delete(key);
		_resolvedBundleMap.delete(key);
	}
	if (!_storeInitCache.has(key)) {
		const p = _doInitStores(cfg, pluginDataDir, logger).then((res) => {
			_resolvedBundleMap.set(key, res);
			return res;
		});
		_storeInitCache.set(key, p);
	}
	return _storeInitCache.get(key);
}

function resetStores(pluginDataDir) {
	if (pluginDataDir) {
		_storeInitCache.delete(pluginDataDir);
		_resolvedBundleMap.delete(pluginDataDir);
	} else {
		_storeInitCache.clear();
		_resolvedBundleMap.clear();
	}
}
```

We applied this patch locally on 1.0.1: the gateway now recovers automatically on the next turn after any close/failure event, and vector recall (vec hits) is restored without process restart.

dvd233:
## Description | 描述

Fixes the stale-store-cache failure reported in #1155.

`initStores()` caches the `_doInitStores()` promise per `pluginDataDir`. Two lifecycle gaps made that cache permanent in exactly the scenarios where a retry would have recovered:

1. **Failed init is cached forever.** The catch path in `_doInitStores()` still returns a bundle (with `vectorStore`/`embeddingService` set to `undefined`), so one transient error — e.g. `SQLITE_BUSY` during a restart race — permanently disabled vector/FTS recall and embedding for the rest of the process lifetime.
2. **Closed stores keep being served from the cache.** `VectorStore.close()` marks the store closed, but nothing invalidates `_storeInitCache`. When close happens outside the `gateway_stop → resetStores()` path (observed around compaction teardown and restart races), every later call reuses the closed instance and fails with `statement has been finalized` / `database is not open` until a full process restart.

Changes:

- `initStores()` evicts a resolved **failed** bundle from the cache (current callers still receive it — failure semantics unchanged), so the next call retries initialization.
- `initStores()` validates the cached bundle on access: if its store reports closed via the new optional `IMemoryStore.isClosed()` (implemented by the SQLite store), the bundle is transparently rebuilt.
- Eviction is guarded by an identity check (`_storeInitCache.get(key) === promise`) so concurrent rebuilds can never clobber each other; concurrent callers still share one in-flight init.
- `resetStores()` behavior is unchanged.

Intentionally out of scope: rebuilding on *degraded* stores (a much larger behavioral surface) and the complementary degrade-strategy approach in #809.

## Related Issue | 关联 Issue

Fixes #1155

## Change Type | 修改类型

- [x] Bug fix | Bug 修复

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Verification (Node v24.15.0, `npx vitest run` → 4/4 pass):

- **Red → green**: on the unpatched code, the new tests "does not cache a failed init bundle" and "re-initializes when the cached store has been closed" fail; the healthy-cache reuse and `resetStores()` contract tests pass before and after.
- The new `src/utils/pipeline-factory.test.ts` covers all four scenarios with a mocked store factory.
- `npm run build:plugin` (tsdown) passes.

---

> This PR was prepared with AI coding assistance; the reproduction analysis, patch, and verification were reviewed and run locally by me. 本 PR 在 AI 编码助手辅助下完成，分析、改动与验证均由本人本地执行并审查。

### DIFF
--- a/MemoryCore/src/core/store/sqlite.ts
+++ b/MemoryCore/src/core/store/sqlite.ts
@@ -3832,4 +3832,9 @@ export class VectorStore implements IMemoryStore {
       );
     }
   }
+
+  /** Whether close() has been called (see IMemoryStore.isClosed). */
+  isClosed(): boolean {
+    return this.closed;
+  }
 }
--- a/MemoryCore/src/core/store/types.ts
+++ b/MemoryCore/src/core/store/types.ts
@@ -566,6 +566,13 @@ export interface IMemoryStore extends MemoryPromptStore, MemoryGenerationRefStor
   isDegraded(): boolean;
   getCapabilities(): StoreCapabilities;
   close(): void;
+  /**
+   * Whether close() has been called. Optional: only stores holding closable
+   * resources (e.g. SQLite) implement it. Used by initStores() to detect a
+   * cached store that was closed outside the normal reset path (restart /
+   * compaction races) and re-initialize instead of reusing a dead handle.
+   */
+  isClosed?(): boolean;
 
   // ── L1 Write ─────────────────────────────────────────────
 
--- a/MemoryCore/src/utils/pipeline-factory.test.ts
+++ b/MemoryCore/src/utils/pipeline-factory.test.ts
@@ -1,0 +1,136 @@
+/**
+ * Regression tests for https://github.com/TencentCloud/TencentDB-Agent-Memory/issues/1155
+ *
+ * initStores() caches the promise returned by _doInitStores() per dataDir.
+ * Two lifecycle flaws made that cache permanent in failure scenarios:
+ *
+ * 1. A failed init (e.g. sqlite busy during a restart race) resolves to a
+ *    bundle with vectorStore/embeddingService = undefined — and that failure
+ *    bundle stayed cached forever, permanently disabling vector/FTS recall.
+ * 2. A store that was closed later (compaction teardown / restart race outside
+ *    the gateway_stop reset path) kept being served from the cache, producing
+ *    "statement has been finalized" / "database is not open" on every turn.
+ *
+ * initStores() must not cache failed bundles, and must re-initialize when the
+ * cached bundle's store has been closed.
+ */
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { IMemoryStore } from "../core/store/types.js";
+
+const createStoreBundleMock = vi.fn();
+
+vi.mock("../core/store/factory.js", () => ({
+  createStoreBundle: (...args: unknown[]) => createStoreBundleMock(...args),
+}));
+
+// Keep the manifest side effects (fs) out of these unit tests.
+vi.mock("./manifest.js", () => ({
+  readManifest: () => undefined,
+  writeManifest: () => undefined,
+  buildStoreInfo: () => ({}),
+  diffStoreBinding: () => [],
+}));
+
+import { initStores, resetStores } from "./pipeline-factory.js";
+
+const logger = {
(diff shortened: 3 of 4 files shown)


Maxwell-Code07:

PR #1173, opened 2026-08-27
fix(store): honor recordIds filter in sqlite queryL1Records

### LINKED ISSUE #1027: sqlite store: queryL1Records ignores filter.recordIds → /v3/atomic/update returns 403 'belongs to a different agent'
## Summary
On the **sqlite** store backend, `VectorStore.queryL1Records()` ignores `filter.recordIds`. `handleAtomicUpdate` (gateway/v2-router.ts) calls `store.queryL1Records({ recordIds: [id] })` and takes `existing[0]`, so it reads an arbitrary row from the whole table instead of the requested note. If that row belongs to a different agent/user, `/v3/atomic/update` fails with:

```
403 Atomic note m_... belongs to a different agent
```

The tcvdb backend honours `recordIds` (`store/tcvdb.ts` ~L743 → `documentIds`), so this only bites standalone/sqlite deployments. Same root cause likely affects `handleAtomicDelete` if it goes through the same path.

## Repro (standalone, storeBackend: sqlite, v3 API)
1. Two agents in the same team each produce ≥1 L1 atomic note (via `/v3/conversation/add`).
2. As agent B, call `/v3/atomic/update` on one of B's own note ids with `team_id`, `agent_id`, `user_id`, `session_id` all set correctly.
3. Gateway log shows: `[L1-query] filter={sessionKey=(all), sessionId=(all), teamId=(all), userId=(all), agentId=(all), ...}, returned N record(s)` (N = whole table), then 403 if `existing[0]` is A's note.

## Fix (verified locally)
`MemoryCore/src/core/store/sqlite.ts`, in `queryL1Records()` after the in-memory isolation filters:

```ts
if (filter?.recordIds && filter.recordIds.length > 0) {
  const want = new Set(filter.recordIds);
  rows = rows.filter((r) => want.has(r.record_id));
}
```

After this, `/v3/atomic/update` on the correct note returns 200 and bumps `version`.

Base: `feat/server_team` @ `fe3230f`. Happy to open a PR if wanted.

dvd233:
## Description | 描述

Fixes #1027.

The sqlite backend's `queryL1Records()` ignored `filter.recordIds`: the prepared-statement matrix only handles session/time predicates, and the in-memory isolation filters only cover `teamId` / `userId` / `agentId` / `taskId`.

`POST /v3/atomic/update` loads the target note via `store.queryL1Records({ recordIds: [id] })` and takes `existing[0]`. With the filter dropped, that is an **arbitrary row from the whole table**, so on multi-agent sqlite deployments the ownership check can trip a spurious `403 "belongs to a different agent"` — or, worse, the update could inherit `user_id` / `agent_id` from an unrelated row.

Change: apply the `recordIds` primary-key filter in memory alongside the existing isolation-dimension filters, matching both the tcvdb backend's `documentIds` path and the documented `L1QueryFilter.recordIds` contract.

Verified while investigating: `handleAtomicDelete` is **not** affected — it deletes through `deleteL1`'s primary-key path, not via `queryL1Records`.

## Related Issue | 关联 Issue

Fixes #1027

## Change Type | 修改类型

- [x] Bug fix | Bug 修复

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

Verification (Node v24.15.0, `node:sqlite`):

- **Red → green**: added `src/core/store/sqlite.test.ts` with three cases against a real on-disk sqlite store — `recordIds` selection, non-existent ids, and `recordIds` combined with an isolation dimension. The two filter-only cases fail before the fix (whole table returned) and pass after; the combined case guards the interaction with isolation filters. `npx vitest run` → 3/3 pass.
- `npm run build:plugin` (tsdown) passes.
- The filter is applied in memory (same pattern as the existing isolation dimensions), keeping the prepared-statement matrix unchanged.

---

> This PR was prepared with AI coding assistance; the reproduction analysis, patch, and verification were reviewed and run locally by me. 本 PR 在 AI 编码助手辅助下完成，分析、改动与验证均由本人本地执行并审查。

### DIFF
--- a/MemoryCore/src/core/store/sqlite.test.ts
+++ b/MemoryCore/src/core/store/sqlite.test.ts
@@ -1,0 +1,85 @@
+/**
+ * Regression tests for https://github.com/TencentCloud/TencentDB-Agent-Memory/issues/1027
+ *
+ * The sqlite store's queryL1Records() ignored `filter.recordIds`, returning
+ * rows from the whole table instead of the requested primary keys. The v3
+ * `/atomic/update` handler loads the target note via
+ * `queryL1Records({ recordIds: [id] })` and takes `existing[0]`, so with more
+ * than one row present it could read another agent's note and fail ownership
+ * checks with a spurious 403.
+ */
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+
+import { VectorStore } from "./sqlite.js";
+import type { MemoryRecord } from "../record/l1-writer.js";
+
+function makeRecord(id: string, agentId: string): MemoryRecord {
+  const now = new Date().toISOString();
+  return {
+    id,
+    content: `note for ${id}`,
+    type: "episodic",
+    priority: 50,
+    scene_name: "default",
+    source_message_ids: [],
+    metadata: {},
+    timestamps: [now],
+    createdAt: now,
+    updatedAt: now,
+    version: 1,
+    sessionKey: "sess-1",
+    sessionId: "sess-1",
+    teamId: "team-1",
+    userId: "user-1",
+    agentId,
+  };
+}
+
+describe("VectorStore.queryL1Records recordIds filter", () => {
+  let dir: string;
+  let store: VectorStore;
+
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1180, opened 2026-08-27
fix(proxy): handle selected single-agent Team for OpenCode

### LINKED ISSUE #1176: [Bug] OpenCode session init fails for a single-agent Team when multiple Teams exist
## OpenClaw Version | OpenClaw 版本

N/A. This is reproduced through the MemoryProxy OpenCode integration, not OpenClaw.

## Plugin Version | 插件版本

TencentDB Agent Memory v2.0.1 deployment/configuration. The running MemoryProxy container reports `context-proxy` version `0.1.0` (`agentmemory/memory-proxy:latest`, digest `sha256:c8de30142787a5df7937c02c167f2ee37f00505b79036357653a6ce78a29fba5`).

## Operating System | 操作系统

Ubuntu 22.04.5 LTS, Docker Compose deployment.

## Describe the bug | 问题描述

When using the v2.0.1 OpenCode integration with interactive session initialization, selecting a Team that has exactly one Agent fails before the Agent selection form can be returned if the user belongs to more than one Team.

The proxy enters `pending_agent_select`, then throws because the OpenCode form builder rejects a page with one option. After the configured retries, it abandons the flow but records the session as `initialized` with no Agent or Task. Later requests are bypassed, so no memory context is injected and no L0 conversation is written.

This is not an OpenCode user-input issue. The same setup works when `x-team-id`, `x-agent-id`, and `x-task-id` are sent explicitly, which bypasses the interactive form.

## To Reproduce | 复现步骤

1. Deploy TencentDB Agent Memory based on v2.0.1 with the OpenCode provider endpoint `/opencode/default/v1`.
2. Enable `sessionInit.enabled: true` and set `headerAutoSelect.onMismatch: form`.
3. Create two Teams for the same user.
4. In one Team, create exactly one Agent and at least one Task.
5. Start a new OpenCode session without `x-team-id`, `x-agent-id`, or `x-task-id` headers.
6. In the session-init questions, choose to associate Team assets, then choose the single-Agent Team.
7. Observe the proxy logs. The flow reaches `pending_agent_select`, repeatedly fails, then abandons the binding.

## Expected behavior | 预期行为

After the Team is chosen, the only Agent should be auto-selected and the user should continue to Task selection (or the one-option Agent form should be rendered normally). The completed session should be bound to Team/Agent/Task, inject the configured memory context, and write L0 conversations.

## Error Logs / Screenshots | 报错日志/截图

```text
[session-init:cb] session=opencode:<session> team=<team> -> pending_agent_select
[session-init] Error in handleSessionInit: [oc form] agent page 0 has 1 option(s); pagination.ts should have avoided a solo last page.
[session-recover] opencode:<session> L2a hit status=pending_agent_select (agent=-, task=-)
[session-init:cb] session=opencode:<session> agent_select max retries, abandoning
[session-recover] opencode:<session> L2a hit status=initialized (agent=-, task=-)
[session-init] session=<session> bypassed -> skipping all injection
```

After the failure, searching the Team/Agent/User isolation bucket found no L0 records for that OpenCode session, and the proxy logged no `tdai-recorder:write-l0` event for it.

## Additional context | 补充信息

The failing path is visible in the v2.0.1 source shipped in the proxy image:

- `MemoryProxy/src/session/opencode/form.ts` computes a page from `team.agents` and throws when `combinedOptions.length < 2`.
- `MemoryProxy/src/session/codebuddy/init.ts` auto-selects a sole Agent only inside the `teams.length === 1` branch. Therefore, the combination of multiple Teams plus a selected Team with one Agent reaches the OpenCode form with exactly one option and fails.

Possible fixes:

1. Auto-select `team.agents[0]` whenever the selected Team has exactly one Agent, regardless of the total Team count; or
2. Allow a one-option OpenCode Agent/Task form.

Current workaround: configure OpenCode model headers with `x-team-id`, `x-agent-id`, and `x-task-id`, which takes the header auto-select path and avoids the form.

Related: #833 proposes OpenCode native integration but does not cover this interactive-form failure.

YOMXXX:
## Summary

- auto-select the sole Agent after a user selects a Team, including the multi-Team flow
- apply the same cascade when a request preselects only a Team through `x-team-id`
- register directly for one Task or render `task_select` for multiple Tasks
- add OpenCode regression coverage for both interactive and header-preselection paths

## Root cause

The shared session-init state machine only cascaded a sole Agent in the single-Team interactive path. Both selecting a Team from a multi-Team form and resolving only a Team from request headers could otherwise transition OpenCode to `pending_agent_select` with one Agent. OpenCode rejects a one-option `question` page, so session initialization failed before binding a valid identity.

## Validation

- `npx vitest run src/session/__tests__/opencode-single-agent-team.test.ts`
- `npm test` — 2 files, 12 tests passed
- `git diff --check` and `git show --check` — clean
- `npx tsc --noEmit` remains blocked by existing repository-wide type errors outside this change (including pre-existing `resetFlow` result typing errors); this change adds no new diagnostic

Fixes #1176

### DIFF
--- a/MemoryProxy/src/session/__tests__/opencode-single-agent-team.test.ts
+++ b/MemoryProxy/src/session/__tests__/opencode-single-agent-team.test.ts
@@ -1,0 +1,126 @@
+import { describe, expect, it } from "vitest";
+
+import type { SessionInitConfig } from "../../types.js";
+import { handleSessionInit } from "../index.js";
+import { SessionStore } from "../store.js";
+import type { SessionInitState, TeamOption } from "../types.js";
+
+class InMemorySessionStore {
+  readonly states = new Map<string, SessionInitState>();
+
+  get(key: string): SessionInitState | undefined {
+    return this.states.get(key);
+  }
+
+  async set(key: string, state: SessionInitState): Promise<void> {
+    this.states.set(key, state);
+  }
+}
+
+const config: SessionInitConfig = {
+  enabled: true,
+  maxRetries: 3,
+};
+
+function sessionStore(initialState: SessionInitState): InMemorySessionStore {
+  const store = new InMemorySessionStore();
+  store.states.set("opencode:session-1", initialState);
+  return store;
+}
+
+function teams(tasksInSelectedTeam: TeamOption["tasks"]): TeamOption[] {
+  return [
+    {
+      team_id: "team-selected-0001",
+      team_name: "Selected Team",
+      agents: [{ agent_id: "agent-only-0001", agent_name: "Only Agent" }],
+      tasks: tasksInSelectedTeam,
+    },
+    {
+      team_id: "team-other-000002",
+      team_name: "Other Team",
+      agents: [{ agent_id: "agent-other-0002", agent_name: "Other Agent" }],
+      tasks: [{ task_id: "task-other-00002", task_name: "Other Task" }],
+    },
+  ];
+}
+
+function pendingTeamSelection(cachedTeams: TeamOption[]): SessionInitState {
+  return {
+    status: "pending_team_select",
+    keyId: "session-1",
+    startedAt: 1,
+    attemptCount: 0,
+    userId: "user-1",
+    cachedTeams,
+  };
+}
+
+function selectedTeamAnswer(): Record<string, unknown>[] {
+  return [{ role: "user", content: "Selected Team" }];
+}
+
+describe("OpenCode session initialization for a selected single-agent Team", () => {
+  it("skips agent selection and renders task selection when the selected Team has multiple Tasks", async () => {
+    const cachedTeams = teams([
+      { task_id: "task-first-00001", task_name: "First Task" },
+      { task_id: "task-second-0002", task_name: "Second Task" },
+    ]);
+    const store = sessionStore(pendingTeamSelection(cachedTeams));
+
+    const result = await handleSessionInit(
+      "session-1",
+      "user-1",
+      selectedTeamAnswer(),
+      config,
+      store as unknown as SessionStore,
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

YOMXXX:
## Follow-up fix: Team-only header preselection

This PR now also covers the header-driven entry point where a request supplies `x-team-id` but no `x-agent-id`.

### Root cause

After resolving only the Team, the shared state machine unconditionally entered `pending_agent_select` for OpenCode. When that Team has exactly one Agent, the OpenCode `question` renderer rejects the resulting single-option Agent page and throws before the session can bind. This was the same single-option invariant behind the original multi-Team interactive failure, but it remained reachable through preset identity.

### Change

The resolved Team now follows the existing cascade when it has one Agent:

1. **0 Tasks** → preserve the existing bypass behavior.
2. **1 Task** → register the session directly with the sole Agent and Task.
3. **2+ Tasks** → persist `pending_task_select` and render only the Task question.
4. **2+ Agents** → preserve the prior `pending_agent_select` / legacy `pending_agent_task` behavior unchanged.

### Regression coverage

Added OpenCode integration-level state-machine tests for Team-only preset identity:

- one Agent + multiple Tasks → `task_select`, with the Agent pre-bound;
- one Agent + one Task → direct registration, with no Agent form rendered.

### Validation

- `npm test` → **2 files, 12 tests passed**.
- `git diff --check` and `git show --check` → clean.
- `npx tsc --noEmit` is still blocked by existing repository-wide diagnostics (for example the pre-existing `SessionInitResult.resetFlow` mismatch and unrelated handler/storage errors). This change adds no new type-check diagnostic.

PR #1186, opened 2026-08-27
feat(pi): interactive Team→Agent→Task picker (parity with CC/Codex/OpenCode)

nicwn:
## Why this PR exists

PR #1126 added first-class `pi` support but shipped only the **static header-preselect** path — set `TDAI_TEAM_ID`/`TDAI_AGENT_ID` env vars → proxy `register directly`, no form — the same UX tier as Hermes/OpenClaw. The stated goal for Pi was parity with the **interactive** clients (Claude Code / CodeBuddy / Codex / OpenCode), where every new session pops a Team → Agent → Task picker on turn 1.

That picker was designed and built (RFC: \`docs/rfc-pi-interactive-picker.md\`, with a live spike against a running proxy) but was sequenced as a follow-up PR that was never opened; the static half merged alone. **This PR closes that gap** — and was verified end-to-end by dogfooding a real interactive Pi session against a running stack before opening.

## What's in this PR

**Interactive picker (client side, \`MemoryCore/pi-plugin/\`):**
- \`form.ts\` — parse the proxy's double-encoded \`ask_followup_question\` arguments; format the answer as \`<question_answer>\` XML (per-question \`id\` split so the proxy's extractor handles the combined agent+task round without collision).
- \`picker.ts\` — render a Pi-native TUI menu via \`ctx.ui.custom\` (↑↓ navigate • Enter select • Esc cancel), one menu per question in order.
- \`index.ts\` — register a custom \`ask_followup_question\` tool (Pi has no built-in form tool) so the proxy's forged tool-call renders as a real picker; identity env vars now **optional** (interactive default, static opt-in).

**First-class agent-pack:**
- \`agents/pi/README.md\` — per-agent doc mirroring \`agents/opencode/README.md\`.
- \`agents/README.md\` + \`INSTALL.md\` — Pi added to the comparison/Session-ID/config tables and the "After deploy" agent list; "Using Proxy with Pi" rewritten (interactive as default, static as opt-in).
- \`agents/setup-proxy.sh\` — \`pi\` in the wizard: symlinks the extension into \`~/.pi/agent/extensions/\`, writes a sourced env file (chmod 600), interactive by default / static preset when \`--team-id\`/\`--agent-id\` are passed.

**Tests restored:** the #1126 squash merge dropped the three proxy test files (31 tests). Restored unchanged from the original branch — all 39 proxy tests pass (8 existing + 31 restored).

## How it works (the key mechanism)

Pi has no native form tool, so the plugin registers \`ask_followup_question\`. The proxy forges a tool_call with that name (id prefix \`call_session_init_\`); Pi's agent loop executes the registered tool, which renders the TUI menu and returns the answer as \`<question_answer>\` XML. Pi sends that back as a \`role:tool\` message with the proxy's \`tool_call_id\` preserved; the proxy's CodeBuddy extractor matches the id and advances the state machine. **Zero proxy changes** — \`agentSource=pi\` already falls into the CodeBuddy state machine's default branch in \`session/index.ts\`, reusing the battle-tested \`ask_followup_question\` form path unchanged.

State machine reused: \`asset_confirm → team_select → agent_task_select → initialized\`. The \`asset_confirm\` yes/no gate runs first; the agent+task round carries two questions in one form. The binding is session-scoped, so the picker only fires on turn 1 of a new Pi process (matching CC's "every new session triggers the picker; subsequent turns reuse the binding").

## Verified

Dogfooded against a running stack (rebuilt \`tdai-proxy\` image + memory-core + memory-hub, a business user with 1 Team + 2 Agents + 2 Tasks):
- New Pi session, no \`TDAI_TEAM_ID\`/\`TDAI_AGENT_ID\` → picker fires on turn 1 (asset_confirm → team → agent+task), selection binds, first message forwarded with memory injected. ✅
- Single-agent team → proxy auto-selects the agent, only asks the task question. ✅
- Multiple agents → full Team → Agent → Task picker. ✅
- Static preset (env vars set) → picker skipped, \`register directly\`. ✅ (the existing path, preserved as opt-in)

Tests: \`cd MemoryCore/pi-plugin && npx vitest run\` → 19/19. \`cd MemoryProxy && npx vitest run\` → 39/39. Plugin \`tsc --noEmit\` clean.

## Non-goals / follow-ups

- No proxy behavior changes (only restored test files + the agent-pack docs/wizard).
- Pi-specific \`asset-import\` scanner (Pi skills/sessions on disk) — separate PR.
- The picker requires TUI mode (\`ctx.mode === "tui"\`); \`pi -p\` / RPC fall back to the static env preset with a clear message.


Maxwell-Code07:

PR #1192, opened 2026-08-28
fix(global-images): quote model variable in LLM check

### LINKED ISSUE #1191: fix(global-images): quote model variable before Chinese punctuation
## Problem

`deploy/global-images/start-all.sh` fails during the interactive LLM connectivity check when the OpenAI `/models` endpoint returns 200 but does not explicitly list the configured model.

In `deploy/global-images/_lib.sh`, the success-message branch writes `$model` immediately before a full-width Chinese comma. Under a UTF-8 locale and `set -u`, Bash parses the punctuation as part of the variable name and exits with `model�: unbound variable`.

## Expected behavior

The setup flow should report the successful connectivity check and continue starting the services.

## Proposed fix

Use `${model}` in the message, matching the equivalent implementation in `verify.sh`.

## Reproduction

1. Configure the memory LLM with a model ID not returned by `/models`.
2. Run `deploy/global-images/start-all.sh` in a UTF-8 locale.
3. The script exits from `check_llm_openai` before any container startup.

cjb1234567:
## Summary

Fixes the interactive `start-all.sh` LLM connectivity check under UTF-8 locales by using `${model}` before full-width Chinese punctuation.

Without braces, Bash with `set -u` can parse the punctuation as part of the variable name when the configured model is absent from the OpenAI `/models` response, terminating the startup flow with an unbound-variable error.

## Validation

- Reproduced the original failure using a mocked HTTP 200 `/models` response without the configured model.
- Re-ran the same reproduction after the change; it completes successfully.
- Ran `bash -n deploy/global-images/_lib.sh deploy/global-images/start-all.sh`.

Closes #1191.

### DIFF
--- a/deploy/global-images/_lib.sh
+++ b/deploy/global-images/_lib.sh
@@ -178,7 +178,7 @@ check_llm_openai() {
     if grep -q "\"$model\"" "$body_file" 2>/dev/null; then
       ok "$label OpenAI 协议通路 OK（$model 在 /models 列表内）"
     else
-      ok "$label OpenAI 协议通路 OK（未在 /models 里显式列出 $model，业务侧仍可能可用）"
+      ok "$label OpenAI 协议通路 OK（未在 /models 里显式列出 ${model}，业务侧仍可能可用）"
     fi
   elif [[ "$code" == "401" || "$code" == "403" ]]; then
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:

PR #1062, opened 2026-08-19
feat(proxy): add pi agent support

san-tian:
## Summary

Adds native Proxy support for [pi](https://github.com/earendil-works/pi-mono), an open-source coding agent harness that speaks the standard OpenAI Chat Completions protocol.

## Changes

- **`MemoryProxy/src/routes/whitelist.ts`**: register `pi` in `AGENT_PREFIX_RE` so `/pi/{spaceId}/v1/chat/completions` requests are recognized and prefix-stripped.
- **`MemoryProxy/src/agent-adapters/pi.ts`**: new adapter, reuses the conservative default behaviour (classify `main`, join text blocks), with `agentKind: "pi"`.
- **`MemoryProxy/src/agent-adapters/types.ts`**: add `"pi"` to the `AgentKind` union.
- **`MemoryProxy/src/agent-adapters/index.ts`**: route `"pi"` → `piAdapter`.
- **`INSTALL.md`**: add "Using Proxy with pi" section.

## Verification

Deployed the full stack (memory-core + memory-hub + proxy) and pointed a pi client at `http://<proxy>:8096/pi/default/v1` with `x-team-id` / `x-agent-id` / `x-task-id` headers plus a per-session `x-conversation-id` injected via a pi extension's `before_provider_headers` hook. Confirmed:

- sessionInit completes via header auto-select (no interactive form)
- L2/L3 profile memory + memory tools injected into the system prompt
- L0 conversation recorded, L1 extraction runs

## Notes

pi's provider headers are static, so the per-session `x-conversation-id` needs the small extension shown in the INSTALL section; this matches the existing Hermes/OpenClaw pattern where the client manages that header.

PR #1195, opened 2026-08-28
fix(proxy): register pi in AGENT_PREFIX_RE

san-tian:
## Summary

Follow-up to #1126 (first-class Pi adapter). That PR added the `pi` agent adapter but left `AGENT_PREFIX_RE` without `pi`, so `/pi/{spaceId}/v1/...` requests are not prefix-stripped and cost-guard / analyse markers on `/pi` paths are not recognized (whitelist matching misses).

This PR is the minimal change:

- Add `pi` to the `AGENT_PREFIX_RE` alternation (one line)
- Add vitest tests covering:
  - `/pi/v1/messages` and `/pi/{spaceId}/v1/...` normalization
  - cost-guard / analyse markers on `/pi` paths
  - whitelist endpoint matching for `/pi` paths
  - regression for existing agent prefixes (claude-code, codebuddy)

## Test

```
cd MemoryProxy && npx vitest run src/routes/__tests__/whitelist.test.ts
# 13 passed
```

Supersedes #1062.


Maxwell-Code07:

L2ncE (approved):
LGTM. Verified the 13 tests fail on the pre-fix regex, and /ping-style prefixes can't false-positive. Thanks for splitting just the gap out of #1062.

PR #1197, opened 2026-08-29
fix(proxy): CC main-dialog agentic loop misclassified as fork — L0 writes silently skipped for all sessions

### LINKED ISSUE #1089: [Bug][Proxy] /v3/conversation/add failures are swallowed, bypassing L0 retry and observability

kristhianmanue1:
## Summary

The `cache_control` marker-position heuristic in `cc-request-classifier.ts` classifies **every real Claude Code main-dialog session as `fork`** once the conversation grows past 2 messages. When `kind=fork`, the proxy skips both the L0 memory write and the skill buffer — so **no conversation is ever recorded** for Claude Code users doing actual agentic work (tool use), which is most real work.

This is likely the root cause behind #964 ("L0 write never attempted with Claude Code"), which reports the same symptom from a different angle.

## Root cause

The original rule:

```ts
// messages[n-2] → FORK
if (markerIdx === n - 2) return "fork";
```

...was validated against captured fork traffic (title/recap/compact prompts with `skipCacheWrite=true`). But in a **main-dialog agentic loop**, Claude Code places the incremental prompt-cache breakpoint on the last **assistant** message (the one carrying `tool_use`). When the turn ends with a `tool_result` (role=user), that marked assistant message sits at `n-2` — and gets classified as fork.

**Observed in production logs** (full 3-in-1 local stack, Claude Code → proxy):

```
[cc-routing] session=74289ba8 kind=main msgs=1
[cc-routing] session=74289ba8 kind=main msgs=2
[cc-routing] session=74289ba8 kind=fork msgs=3    ← first tool use
[cc-routing] session=74289ba8 kind=fork msgs=5
...
[cc-routing] session=74289ba8 kind=fork msgs=89
[cc-routing] skip L0 write (stream) for kind=fork ...
```

Every active session showed the same pattern. After the fix, the same session classifies `main` at `msgs=267/269` and L0 writes resume immediately (`tdai-recorder:write-l0` present in logs, records landing in the kernel's conversations store).

## Fix

`n-2` only means **fork** when the marked message is `role=user` (internal fork prompts are user-terminated). If the marked message at `n-2` is `role=assistant`, it's the main dialog's agentic loop → `main`:

```ts
if (markerIdx === n - 2 && roleAt(msgs, markerIdx) === "user") return "fork";
return "main";
```

Legitimate forks (title/recap/compact, which reuse the main cache and end in a user message) keep classifying as `fork` — covered by tests.

## Tests

15 new cases in `cc-request-classifier.test.ts`:
- the incident pattern (marker on assistant at n-2, msgs=3..89) → main
- legitimate forks (marker on user at n-2 with trailing user prompt) → fork
- sidequery fallback, classic main regression, role-filtering, defensive defaults

Full suite: **28/28 green**.

## Deployment evidence

Built from this fix and deployed locally (`agentmemory/memory-proxy`, docker, 3-in-1 stack): session re-linked, classification flipped to `main`, L0 writes confirmed landing in the memory kernel within the first minute. The local deployment also includes two earlier fixes on the same branch (task_id search scoping in the memory bridge, `externalGatewayUrl` emission in the regenerated proxy config) — kept in the merge for completeness.

## Notes

- A stronger long-term signal would be an explicit sidechain flag in the request body (e.g. `isSidechain` metadata) rather than position inference; left as future work.
- Upstream issue #964's hypothesis (extractLatestUserMessage discarding tool_results) may be an additional independent failure layer; this PR does not touch that path.

## Relation to #1203 (stacked)

This branch (`integration/server-team`) **contains** the branch of #1203 (`fix/search-task-scope`) — the PRs share diff today because this one is a superset. Merge order matters:

1. **Merge #1203 first** (strict L0 write: `/v3/conversation/add` failures must throw → retry instead of being swallowed). Fixes #1089.
2. **Then #1197** (this one): the fork-classifier fix. Its base already includes #1203, so merging it after is conflict-free.

Independent value: #1197 fixes *when* L0 writes are attempted (fork misclassification skipped them all); #1203 fixes *knowing whether they worked*. Complementary, both needed for "no silent data loss" on the L0 path.


kristhianmanue1:
Friendly ping for review when maintainers have bandwidth. This fixes the CC main-dialog agentic loop being misclassified as a fork — for all Claude Code users, that classification silently skips L0 writes, so entire sessions never reach memory (this is the root cause we traced after losing session data; the write-side strict-mode companion is #1203, which should merge first). Happy to iterate on any feedback. 🤖

Maxwell-Code07:

kristhianmanue1:
Heads-up for reviewers: we updated the PR head (merge of upstream `feat/server_team` @ 220af62) to resolve the merge conflict introduced by recent upstream commits (v2.0.2-beta.1, Pi adapter, task-optional memory). The only conflict was a one-line UI-text comment in two session init files — resolved by adopting the upstream wording verbatim. The PR diff now shows only this PR's actual contribution (11 files, +1057/−10). All MemoryProxy tests pass (33/33). Mergeable again — ready for the internal review whenever maintainers have bandwidth.

PR #1203, opened 2026-08-29
fix(proxy): L0 write failures must not be swallowed — strict mode for /v3/conversation/add

### LINKED ISSUE #1089: [Bug][Proxy] /v3/conversation/add failures are swallowed, bypassing L0 retry and observability

kristhianmanue1:
Fixes #1089.

## Summary

`postForCtx` (`MemoryProxy/src/tdai/client.ts`) swallowed every failure path — HTTP non-2xx, envelope `code !== 0`, network errors, timeouts — by returning `{} as T`. That is the right semantics for the *read* paths (`searchL1`, `listL2`, `readL2`, `readL3`: injection degrades silently, and the injection pipeline has its own per-hook catch), but for the **L0 write** it defeats the existing retry scaffolding:

```ts
await withL0Retry(() => client.addConversation(identity, messages), { attempts: 3 });
```

`withL0Retry` only retries on rejection. With errors swallowed, a failed write resolved as if the chat turn had succeeded — exactly the "write was skipped" vs "write was attempted and failed" indistinguishability this issue describes.

## Change (minimal, path-directed)

- `postForCtx` gains `options.strict`: in strict mode it throws a typed `TdaiWriteError` (carrying HTTP status + envelope detail) instead of returning `{}`
- `addConversation` (`/v3/conversation/add`) is the **only** strict caller — read paths keep their silent-degrade semantics untouched
- `TdaiWriteError.message` preserves the `tdai POST <path> HTTP <code>` format that `isRetryable()` in `pending-writes.ts` already parses — so 5xx/408/429/network retry, 4xx fail fast, no retry-logic changes needed

Note: the codebase already documents this exact asymmetry as intended for `checkAcl` (fail-closed vs silent-degrade, see the comment above `checkAcl`). This PR extends the same reasoning to the write path — writes deserve fail-loud, not fail-silent.

## Tests (`tdai-l0-write-errors.test.ts`, 5 cases)

- **the issue's exact repro**: mocked 503 → `withL0Retry` makes **3 fetch calls**, then rejects with `TdaiWriteError`
- HTTP 200 + envelope `code=40101` → rejects (kernel-level rejection is a real failure)
- HTTP 400 → single attempt, rejects immediately (non-retryable)
- happy path unchanged
- regression guard: `searchL1` with a down kernel still returns `[]` (read paths keep silent degrade)

Full suite: **33/33 green**.

## Relation to #1197

Same subsystem, complementary story: #1197 fixes *when* L0 writes are attempted (fork misclassification skipped them all); this fixes *knowing whether they worked*. Both are necessary for "no silent data loss" on the L0 path.

### DIFF
--- a/MemoryProxy/src/__tests__/cc-request-classifier.test.ts
+++ b/MemoryProxy/src/__tests__/cc-request-classifier.test.ts
@@ -1,0 +1,221 @@
+import { describe, it, expect } from "vitest";
+
+import { classifyCcRequest, findLastCacheControlIndex } from "../common/cc-request-classifier.js";
+
+/**
+ * Gate del fix de clasificación main/fork (incidente 2026-08-28/29):
+ *
+ * Síntoma: TODAS las sesiones claude-code se clasificaban kind=fork a partir
+ * de msgs>=3 → el proxy saltaba L0 write + skill buffer (proxy_kv = 0 filas).
+ *
+ * Causa raíz: la regla original "marker en n-2 → fork" asume que fork reutiliza
+ * el caché del main dejando el marker en el penúltimo mensaje. Pero en el loop
+ * agéntico REAL del main dialog, Claude Code coloca el breakpoint incremental
+ * del prompt-caching sobre el último mensaje assistant (tool_use) que queda en
+ * n-2 cuando el turno termina en un tool_result (user). Resultado: la sesión
+ * main viva era "fork" en cada request agéntico → 0 writes L0.
+ *
+ * Fix: n-2 SOLO es fork si el mensaje marcado es role=user (los prompts internos
+ * de fork — title/recap/compact — terminan en user). Si el mensaje marcado es
+ * assistant, es el loop agéntico del main → MAIN.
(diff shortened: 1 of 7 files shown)


kristhianmanue1:
Friendly ping for review when maintainers have bandwidth. This fixes silent data loss: currently any L0 write failure in /v3/conversation/add is swallowed (HTTP 200 with the error in the response envelope only), so clients believe conversations were persisted when they were not — we lost real session data to this in production before diagnosing it (details in the PR body, closes #1089). Merge order note: merging this first makes #1197 apply cleanly on top. Happy to iterate. 🤖

Maxwell-Code07:

PR #1211, opened 2026-08-30
fix(proxy): send Authorization Bearer on auth/verify when gateway key is configured

kristhianmanue1:
## Problem

MemoryProxy calls `/v3/meta/auth/verify` **without** an Authorization header (MemoryProxy/src/auth.ts — `fetchOpts.headers` only carries `content-type` and `x-tdai-service-id`). The core gateway's Layer-1 Bearer gate (`checkAuthForV2` → `verifyAuth`) applies to all /v3 routes, so if an operator sets `TDAI_GATEWAY_API_KEY` (or `server.apiKey`), **every proxy auth verification fails with 401 and session-init breaks**.

This forces the gateway key to stay empty, which is the documented fail-open default: `verifyAuth` returns "ok" without inspecting the request when no key is configured (MemoryCore/src/gateway/server.ts:1118), leaving all data routes reachable with any Bearer value (verified at runtime: read/write/delete with an arbitrary token).

The two halves of this incompatibility are mutually reinforcing: the missing Bearer makes the empty key the only working state, and the empty key is the fail-open.

## Fix

- `auth.ts`: attach `Authorization: Bearer <gatewayApiKey>` to the verify request when `auth.gatewayApiKey` is configured. Optional — empty keeps the legacy behaviour (no header), fully backwards compatible.
- `types.ts`: add `gatewayApiKey` to `AuthConfig` and the raw yaml auth section.
- `config.ts`: resolve precedence `TDAI_PROXY_GATEWAY_API_KEY` (env) > `auth.gatewayApiKey` yaml (with ${VAR} expansion) > default "", matching the existing `admin.apiKey` and `TDAI_GATEWAY_API_KEY` ops habit.

## Verification

- typecheck: no new errors (54 pre-existing on this branch, none in touched files)
- vitest: suite green
- integration probe with a mock verify server: no key → no Authorization header; key set → `Bearer <key>` sent; user_id resolved in both cases

## Operator-facing result

With this change the deployment can run the full chain with the gateway key enabled:

```yaml
auth:
  enabled: true
  url: http://tdai-memory-core:8420
  gatewayApiKey: ${TDAI_GATEWAY_API_KEY}
```

…closing the fail-open data-route exposure by configuration instead of by tolerated risk.

### DIFF
--- a/MemoryProxy/src/__tests__/auth-gateway-token.test.ts
+++ b/MemoryProxy/src/__tests__/auth-gateway-token.test.ts
@@ -1,0 +1,64 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { initAuth, verifyUserKey } from "../auth.js";
+
+const validResponse = () =>
+  new Response(
+    JSON.stringify({
+      code: 0,
+      data: { valid: true, user: { user_id: "usr-test" } },
+    }),
+    { status: 200, headers: { "content-type": "application/json" } },
+  );
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+  initAuth({ enabled: false, url: "", serviceToken: "", timeoutMs: 0 });
+});
+
+describe("auth gateway service token", () => {
+  it("sends the internal token as Bearer without replacing the user_key body", async () => {
+    const fetchMock = vi.fn<(url: string, options: RequestInit) => Promise<Response>>(
+      async () => validResponse(),
+    );
+    vi.stubGlobal("fetch", fetchMock);
+    initAuth({
+      enabled: true,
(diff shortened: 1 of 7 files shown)


kristhianmanue1:
Friendly ping for review when maintainers have bandwidth. Context: this unblocks enabling `TDAI_GATEWAY_API_KEY` in production deployments — currently the proxy's auth/verify calls (without Bearer) force the gateway key to stay empty, which keeps the documented fail-open data-route exposure active by design. The fix is small (+32/-4, 3 files) and fully backwards compatible (empty key → no header, legacy behavior). Happy to iterate on any feedback. 🤖

Maxwell-Code07:

PR #1219, opened 2026-08-30
fix(panel): allow agent deletion without SkillCore

### LINKED ISSUE #1218: [Bug] Agent deletion is blocked when SkillCore is disabled
## Environment

- TencentDB Agent Memory: current `feat/server_team`
- MemoryPanel with a MemoryCore deployment where SkillCore is not enabled
- Operating system: Linux (container deployment)

## Root cause

`POST /api/v1/agent/delete-cascade` always calls `/v3/skill/list` before archiving an Agent. When SkillCore is not configured, MemoryCore returns the explicit capability-absence envelope `code=404`, `message="Skill module not enabled"`. MemoryPanel forwards that response and never reaches `/v3/meta/agent/archive`, even for an Agent with no Skills.

## Reproduction

1. Run MemoryPanel against a MemoryCore instance without SkillCore enabled.
2. Create an Agent owned by the caller and do not attach any Skills.
3. Call `POST /api/v1/agent/delete-cascade` with that Agent ID.
4. Observe `404 Skill module not enabled`; the Agent remains active.

Directly calling the retained `/api/v1/meta/agent/archive` escape hatch succeeds, which confirms the failure is in the cascade control route.

## Expected behavior

The cascade route should treat only the explicit Skill capability-absence response (`404` plus `Skill module not enabled`) as an empty Skill collection and continue to archive the Agent.

It must continue to fail closed when:

- `skill/list` returns any other 404 or error;
- SkillCore is available and `skill/delete` fails.

When Skills exist, they must still be deleted before the Agent is archived.

## Scope and acceptance evidence

- Add route-level regression coverage for disabled SkillCore, unrelated list failures, normal Skill deletion ordering, and Skill deletion failure.
- Make the smallest Control-layer error-classification change; do not enable SkillCore or change MemoryCore persistence/authorization.
- Run MemoryPanel tests, typecheck/build, and the repository secret scan.

## Risk

The main risk is accidentally swallowing unrelated upstream 404/errors and orphaning active Skills. The fix therefore needs an exact capability-absence match and tests that preserve fail-closed behavior.

RainBoat:
## Description | 描述

Fixes #1218.

### Root cause

`POST /api/v1/agent/delete-cascade` unconditionally calls `skill/list`. A deployment without SkillCore returns the explicit capability-absence envelope `404 Skill module not enabled`, which MemoryPanel previously forwarded before reaching `meta/agent/archive`. This blocked deletion even when the Agent had no Skills.

### Behavior change

- Treat only the exact `code=404` plus `message=Skill module not enabled` response as an empty Skill collection, then continue to archive the Agent.
- Preserve fail-closed behavior for every other `skill/list` error.
- Preserve the existing ordering when SkillCore is available: list Skills, delete every Skill, then archive the Agent.
- Preserve the existing failure behavior: if any `skill/delete` fails, do not archive the Agent.

### Non-goals

- Do not enable or deploy SkillCore.
- Do not change MemoryCore, persistence, authorization, or Agent ownership.
- Do not treat generic 404 or upstream failures as an empty Skill collection.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Verification | 验证

- [x] Regression test first reproduced the old `404` behavior.
- [x] `pnpm test` — 4/4 route-level tests passed.
- [x] `pnpm typecheck`
- [x] `pnpm build`
- [x] Standalone strict TypeScript check for the new test.
- [x] Scoped strict secret scan for all changed files.
- [ ] Full `bash scripts/secret-scan.sh --strict` is currently blocked by three pre-existing placeholder hits in `MemoryPanel/web/src/pages/GuidePage/index.tsx` at lines 88, 106, and 141; this PR does not change that file.
- [ ] Live E2E was not run; the regression is covered at the public Hono route with the MemoryCore boundaries stubbed.

Covered cases:

1. SkillCore explicitly disabled: empty Agent is archived.
2. Unrelated Skill list errors still block archive.
3. Existing Skills are deleted before archive.
4. Skill deletion failure prevents archive.

## Compatibility and rollback | 兼容与回滚

No request/response schema or durable state changes. SkillCore-enabled deployments retain their current behavior. Rollback is a single commit revert; that restores the prior failure for SkillCore-disabled deployments without requiring data migration.

## Self-test Checklist | 自测清单

### DIFF
--- a/MemoryPanel/panel-api-doc.md
+++ b/MemoryPanel/panel-api-doc.md
@@ -765,7 +765,7 @@ L0/L1 列表批量删除。**仅资产 Owner**。
 | deleted_skill_count | number | 已删 skill 数 |
 | deleted_skill_ids | string[] | 已删 skill ID 列表 |
 
-**错误**：`MISSING_AGENT_ID`、`INVALID_USER_KEY`、`AGENT_NOT_FOUND`、`NOT_YOUR_AGENT`；任一 skill 删除失败返回 `500 SKILL_DELETE_FAILED`（含 `failed_skill_id`、`deleted_skill_ids`），此时 agent 不会 archive。
+**错误**：`MISSING_AGENT_ID`、`INVALID_USER_KEY`、`AGENT_NOT_FOUND`、`NOT_YOUR_AGENT`；任一 skill 删除失败返回 `500 SKILL_DELETE_FAILED`（含 `failed_skill_id`、`deleted_skill_ids`），此时 agent 不会 archive。若内核明确返回 `404 Skill module not enabled`，Control 会将其视为没有可级联的 skill 并继续归档 Agent；其它 `skill/list` 错误仍原样阻断归档。
 
 **示例**
 
--- a/MemoryPanel/src/panel/http/routes/agent-lifecycle.ts
+++ b/MemoryPanel/src/panel/http/routes/agent-lifecycle.ts
@@ -10,7 +10,8 @@
  * 本路由的做法（业务级联收口在 control 层，不改内核）：
  *   1. auth/verify 反查 caller
  *   2. agent/get 拿到 agent，强校验 owner_user_id === caller（本期不允许 admin 代删）
- *   3. skill/list 按 owner_agent_id + active 分页拉全
+ *   3. skill/list 按 owner_agent_id + active 分页拉全；只有内核明确返回
+ *      404 "Skill module not enabled" 时，才把 capability 缺失视为没有 skill
  *   4. 逐条 skill/delete —— 任一失败立即中断，返回 500 + 已删列表 + 失败 skill_id
  *      + 内核错误 message；此时 agent/archive 不会被调用，caller 需要修复后重试
  *   5. 全部 skill 成功归档后调 meta/agent/archive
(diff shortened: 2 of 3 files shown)


Maxwell-Code07:

pengpengyi92:
Thanks for the focused fix. I ran the four route tests, typecheck, and build locally; all pass. I also reproduced one pagination boundary that should remain fail-closed: if page 1 returns 100 active Skills with total=101, then page 2 returns code=404/message='Skill module not enabled', listAgentSkills currently returns { ok: true, items: [] }. The route then archives the Agent without deleting the 100 Skills already observed. A temporary hostile regression test expected HTTP 404/no agent:archive but received HTTP 200 and agent:archive. Suggested minimal guard: treat the exact capability-absence envelope as an empty collection only before any successful page has been accumulated (for example offset===0 && all.length===0); after pagination has begun, return the envelope unchanged. Adding this case to agent-lifecycle.test.ts would preserve the intended disabled-SkillCore behavior while preventing orphaned Skills if capability availability changes mid-pagination.

HXYreality666:
I saw pengpengyi92's pagination boundary finding. I'd like to add a regression test and the suggested guard (only treat the capability-absence envelope as empty when offset===0 && all.length===0). Happy to coordinate if you prefer to include it in this PR.

PR #1220, opened 2026-08-30
fix(sanitize): repair unescaped ASCII quotes inside JSON string values from LLM output

karls0r:
## Description | 描述

Adds a third, conservative repair pass to `sanitizeJsonForParse()` in
`MemoryCore/src/utils/sanitize.ts` that recovers a specific LLM JSON-output
failure mode: the model opens a typographic quote inside a string value
(e.g. a German opening quote, U+201E „) but closes it with a literal ASCII
`"` instead of the matching typographic glyph or an escaped `\"`. Because
the enclosing JSON string is itself delimited by `"`, that stray character
ends the string early and `JSON.parse` fails a few tokens later — a shape
the existing two phases (control-character escaping only) cannot catch,
since neither ever inspects ASCII-quote placement.

The new phase 3 walks the JSON string literals and, on encountering an
unescaped ASCII `"`, peeks at the next non-whitespace character: a JSON
structural continuation character (`,` `}` `]` `:`) or end-of-text means
it's the real terminator and is left untouched; anything else means it's a
stray quote embedded in content, which gets escaped to `\"` while the
string stays open. It deliberately runs on the *original* raw text rather
than the phase-1/2 output, because phase 1's control-char escaper toggles
its string-open/closed state on every unescaped quote with no lookahead —
a single stray quote would desync that state for the rest of the document
(turning real structural newlines into literal `\n` sequences further
down). The lookahead-based fix restores correct quote pairing first;
phases 1 and 2 (unchanged) are then reapplied to the corrected text so
control characters inside the now-correctly-delimited strings are still
handled.

No changes to the exported signature (`sanitizeJsonForParse(raw: string):
string`) or its call site (`l1-extractor.ts:543`). Phase 3 only runs when
phases 1 and 2 both still fail to produce parseable JSON, so well-formed
output is byte-for-byte unaffected.

## Related Issue | 关联 Issue

Same error class as #110 (L1 extraction JSON parse failure) — that report
covers a different model (deepseek-v4-flash) producing unescaped Chinese
characters and trailing text after the JSON object. This PR covers a
quote-mismatch variant of the same underlying problem (non-strict LLM JSON
output breaking a naive `JSON.parse` in the L1 extraction pipeline). Not a
duplicate of #110 — a narrower, additive fix for a specific failure shape
that report doesn't cover and the current sanitizer doesn't handle.

## Reproduction

Observed in production against a locally hosted OpenAI-compatible model
(Qwen3.8-27B) used for L1 memory extraction, with a German-language system
prompt. Across 3 real extraction failures (verified byte-for-byte against
the raw LLM output captured in application logs), all 3 hit the identical
pattern — the model opens a quote with „ (U+201E) and closes it with an
ASCII `"` (U+0022) instead of the typographic closing glyph or `\"`, e.g.
(anonymized, shape-preserving example):

    "content":"The reviewer replied: „Looks good to me"."
                                                        ^ stray ASCII quote ends the string here

`JSON.parse` then fails a few tokens later (`Expecting ',' delimiter` or
similar). `sanitizeJsonForParse()`'s existing phases 1–2 pass this input
through unchanged, since no control character is involved — the extraction
result was silently dropped (`extracted=0`) for every one of these 3
failures before this fix.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

**Unit tests added** (`MemoryCore/src/utils/sanitize.test.ts`, 5 cases, all
passing):

- 3 synthetic fixtures reproducing the exact quote-mismatch shape
  (end-of-sentence, a quoted named entity mid-sentence, a quoted phrase
  followed by an em dash) — all repaired and parse correctly, with the
  stray quote preserved as a literal character in the output (no data
  loss)
- 2 idempotency cases: an already-escaped `\"` and a correctly paired
  typographic quote (`„…“`) both stay byte-identical through
  `sanitizeJsonForParse()`

```
$ npm test

 Test Files  1 passed (1)
      Tests  5 passed (5)
```

**Offline validation against the real failures that motivated this fix:**
replayed the 3 real production payloads that originally broke L1
extraction (reconstructed byte-for-byte from application logs) directly
through the patched `sanitizeJsonForParse()` — all 3 now parse (0/3 before
this fix). 3 previously-valid extraction outputs re-checked for
idempotency — all 3 stay unchanged and still parse.

**Field validation:** after deploying this fix — together with an
unrelated, complementary prompt-side mitigation that reduces how often the
model produces this quote shape in the first place, and is not part of
this PR — a 67-minute live reimport through the L1 pipeline produced 74
extracted records with 1 residual parse failure (down from 3/3 failures on
comparable content before). The residual case marks this heuristic's
documented limit: a stray quote directly followed by a natural-language
comma is indistinguishable from a real string terminator without language
understanding, so phase 3 correctly leaves it untouched; the pipeline's
existing retry regenerates the output, which resolves such cases in
practice.

## Additional Notes | 其他说明

Scope is intentionally narrow: one new private helper
(`escapeStrayQuotesInJsonStrings`), one new call site inside
`sanitizeJsonForParse()`'s existing phase-3 fallback, one colocated test
file. No dependency, config, or public-API changes.

---

*This PR was researched and written by Ren, an AI agent operated by @karls0r, with his review.*

### DIFF
--- a/MemoryCore/src/utils/sanitize.test.ts
+++ b/MemoryCore/src/utils/sanitize.test.ts
@@ -1,0 +1,72 @@
+import { describe, expect, it } from "vitest";
+import { sanitizeJsonForParse } from "./sanitize.js";
+
+/**
+ * Regression tests for the stray-quote repair pass (phase 3) in
+ * `sanitizeJsonForParse`.
+ *
+ * All three failure fixtures below reproduce the same error class: the LLM
+ * opens a typographic quote (e.g. German „) inside a JSON string value but
+ * closes it with a literal ASCII `"` instead of the matching typographic
+ * glyph or an escaped `\"`. That stray `"` ends the JSON string early, and
+ * `JSON.parse` then throws a few tokens later (typically
+ * `Expecting ',' delimiter` or similar). The fixtures use neutral,
+ * synthetic content — not real production data — but reproduce the exact
+ * quote-mismatch shape observed in the wild.
+ */
+describe("sanitizeJsonForParse — stray ASCII quote inside a JSON string value", () => {
+  it("repairs a stray quote at the end of a sentence, followed by trailing punctuation", () => {
+    const broken =
+      '[{"scene_name":"Chat","message_ids":["m1"],"memories":[{"content":"The reviewer replied: „Looks good to me".","type":"episodic","priority":50,"source_message_ids":["m1"],"metadata":{}}]}]';
+
+    const sanitized = sanitizeJsonForParse(broken);
+    expect(() => JSON.parse(sanitized)).not.toThrow();
+
+    const parsed = JSON.parse(sanitized);
+    // The stray quote must be preserved as a literal character in the
+    // repaired content — no data loss, just correct escaping.
+    expect(parsed[0].memories[0].content).toBe('The reviewer replied: „Looks good to me".');
+  });
+
+  it("repairs a stray quote wrapping a named entity mid-sentence", () => {
+    const broken =
+      '[{"scene_name":"Planning","message_ids":["m2"],"memories":[{"content":"The user renamed the project to „Aurora" during the call.","type":"episodic","priority":50,"source_message_ids":["m2"],"metadata":{}}]}]';
+
+    const sanitized = sanitizeJsonForParse(broken);
+    expect(() => JSON.parse(sanitized)).not.toThrow();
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1230, opened 2026-09-01
fix: preserve accurate agent memory layer counts

lixiaoming0536:
## Summary
- preserve real L0–L3 counts in MemoryPanel instead of returning fabricated zero placeholders
- query agent aggregate L0/L1 data by team + agent scope without incorrectly filtering by control-plane asset-owner user ID
- retain explicit user filtering when a user ID is supplied

## Validation
- MemoryCore query-isolation regression tests passed
- MemoryPanel typecheck passed
- MemoryPanel chat-memory regression tests passed

## Scope
No configuration, credentials, or data migration changes.


Maxwell-Code07:

PR #1243, opened 2026-09-02
fix(sdk-py): reject malformed falsey response data

MohammedThowfiq:
## Summary

The Python v3 transport promises object-valued response data, but `envelope.get("data") or {}` converts `[]`, `false`, `0`, and `""` into a successful empty result before checking their types. This hides malformed server responses from both synchronous and asynchronous SDK callers.

Only missing or null data now defaults to an empty object. Other non-object values raise the existing `TDAMError` with the response trace ID. Object payloads and optional data retain their existing behavior.

## Verification

From `sdk/memory-core/python/`, using Python 3.12.13 and httpx 0.28.1:

- `python -m unittest discover -s tests -v` — 2 tests passed, covering 24 synchronous/asynchronous subcases. Before the fix, the 8 falsey-data cases failed because no exception was raised.
- `python -m build` — source distribution and wheel built successfully.

Tests use `httpx.MockTransport` and make no external requests. This change is confined to v3 response decoding; v2 behavior is unchanged.

### DIFF
--- a/sdk/memory-core/python/tencentdb_agent_memory/_v3_http.py
+++ b/sdk/memory-core/python/tencentdb_agent_memory/_v3_http.py
@@ -62,7 +62,9 @@ def _decode_response(resp: httpx.Response) -> dict:
             details=details,
         )
 
-    result = envelope.get("data") or {}
+    result = envelope.get("data")
+    if result is None:
+        result = {}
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1244, opened 2026-09-03
feat: add expected-version CAS for atomic memory updates

### LINKED ISSUE #1236: [Proposal] Prevent lost updates in editable memories with expected_version
## Problem

The v2.0.1 roadmap adds editable L1-L3 memories, but the current L1 update path has no stale-write precondition.

- [`atomicUpdateRequestSchema`](https://github.com/TencentCloud/TencentDB-Agent-Memory/blob/3efcd317b84146d6a08518ac0f7ee7c8a8d200ec/MemoryCore/src/gateway/generated/schemas.ts#L166-L170) accepts only `id`, `content`, and optional `background` (plus isolation fields).
- [`handleAtomicUpdate`](https://github.com/TencentCloud/TencentDB-Agent-Memory/blob/3efcd317b84146d6a08518ac0f7ee7c8a8d200ec/MemoryCore/src/gateway/v2-router.ts#L1044-L1110) reads the current row, computes `record.version + 1`, then calls `upsertL1`.
- [MemoryPanel currently submits](https://github.com/TencentCloud/TencentDB-Agent-Memory/blob/3efcd317b84146d6a08518ac0f7ee7c8a8d200ec/MemoryPanel/src/panel/http/routes/chat-memory.ts#L1644-L1649) `{ id, content }` without the version that was displayed to the editor.

Two clients can therefore read the same version, both submit an edit, and both receive success while the later upsert silently overwrites the earlier one.

## Proposed semantics

1. Add `expected_version` to memory mutation requests.
2. Perform compare-and-swap in the store adapter, not as another router-level read-then-write check.
3. If the current version differs, return HTTP `409` with a stable error code and the current version.
4. Have MemoryPanel submit the version originally loaded by the editor and show a conflict instead of silently overwriting.
5. A rejected stale update must not increment the version or append an `applied` audit event.

For backward compatibility, the field could initially be optional for non-panel clients, while Panel always sends it. Alternatively, a new strict mutation endpoint/version could require it. I would appreciate maintainer guidance on the preferred compatibility policy before implementing this across store adapters.

## Acceptance criteria

- Two concurrent updates with the same `expected_version`: exactly one succeeds; the other receives `409`.
- A stale update changes neither content nor version.
- A successful update increments the version exactly once.
- The conflict response exposes the current version but no memory content.
- Regression tests cover success, stale conflict, and concurrent submissions.
- The behaviour of callers that omit `expected_version` is documented.

## Related work / non-duplicate

- #1025 addresses **semantic staleness and source-message provenance** for memories. This proposal addresses **concurrent human edits and lost updates**.
- I searched open and closed issues/PRs for `expected_version`, `expected version`, `optimistic locking`, `lost update`, `concurrent edit`, and `stale version`; I found no matching implementation proposal as of 2026-09-01.

If this direction fits the editable-memory roadmap, I can follow up with a narrowly scoped implementation plan and regression tests.

zealxz:
## Description | 描述

> Draft implementation for maintainer review following the internal-review acknowledgement in #1236. This does not assume the proposal has been accepted.

Add an optional `expected_version` guard to `/v3/atomic/update` so human-authored L1 edits can reject stale writes instead of silently overwriting a newer revision.

- Preserve the existing last-write-wins behavior when `expected_version` is omitted.
- Perform atomic compare-and-swap updates in both SQLite (`UPDATE ... WHERE version = ?` inside `BEGIN IMMEDIATE`) and Tencent VectorDB (one server-side version-filtered document update).
- Disable transport retries for the conditional Tencent VectorDB update so a lost success response cannot become a false conflict on retry.
- Return `409 ATOMIC_VERSION_CONFLICT` with `expected_version` and `current_version` when the stored version has advanced.
- Keep L1 isolation fields in the atomic predicate and reject cross-team/user/agent/task update attempts.
- Expose L1 versions through query/search paths and pass them from MemoryPanel editors into subsequent updates.
- Add `expected_version` to both TypeScript SDK update APIs while retaining compatibility with older callers and servers.

## Related Issue | 关联 Issue

Closes #1236

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [x] New feature | 新功能

## Self-test Checklist | 自测清单

Verification performed with Node.js v22.22.2:

- [x] MemoryCore CAS regression suite: 4 files, 13 tests passed
  - SQLite winner/stale-writer and not-found behavior
  - Tencent VectorDB version-filtered update, conflict, and not-found behavior
  - Tencent VectorDB conditional-update no-retry policy
  - Gateway success, legacy no-guard behavior, 409/404/500/503 mapping, and unsafe-version validation
- [x] MemoryCore plugin build: `npm run build:plugin`
- [x] MemoryPanel route suite: 4 tests passed (forwarding, 409 propagation, unsafe/non-L1 rejection)
- [x] MemoryPanel server typecheck: `npm run typecheck`
- [x] MemoryPanel web lint: 0 errors (32 pre-existing warnings)
- [x] MemoryPanel web production build: `npm run build`
- [x] TypeScript SDK suite: 2 tests passed (guard included/omitted)
- [x] TypeScript SDK build: `npm run build`

## Additional Notes | 其他说明

- This is opt-in concurrency control: clients that do not send `expected_version` retain the current behavior.
- Durable `operation_id` replay receipts are intentionally out of scope for this patch. That broader retry contract should be designed separately and aligned with the idempotency work in #1087 / #1142.
- The gateway uses a local request-schema override because the generated schema source is not present in this branch; generated files are intentionally left untouched.
- The update response still serializes the new version using the existing `vN` wire format. Query/search items expose the persisted numeric version used as the next `expected_version`.
- MemoryPanel preserves a stale edit draft on 409 and requires the user to reload/reconcile before retrying; it does not silently advance the editor to the server version.
- Please review the Tencent VectorDB `/document/update` predicate shape in `TcvdbMemoryStore.compareAndSwapL1`; it uses a single document ID plus version and isolation filters, not a read-then-upsert sequence.

### DIFF
--- a/MemoryCore/src/core/store/sqlite-cas.test.ts
+++ b/MemoryCore/src/core/store/sqlite-cas.test.ts
@@ -1,0 +1,77 @@
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+import type { MemoryRecord } from "../record/l1-writer.js";
+import { VectorStore } from "./sqlite.js";
+
+const opened: VectorStore[] = [];
+const dirs: string[] = [];
+
+afterEach(() => {
+  for (const store of opened.splice(0)) store.close();
+  for (const dir of dirs.splice(0)) rmSync(dir, { recursive: true, force: true });
+});
+
+function makeRecord(overrides: Partial<MemoryRecord> = {}): MemoryRecord {
+  return {
+    id: "memory-1",
+    content: "original",
+    type: "persona",
+    priority: 50,
+    scene_name: "default",
+    source_message_ids: [],
+    metadata: {},
+    timestamps: [],
+    createdAt: "2026-09-01T00:00:00.000Z",
+    updatedAt: "2026-09-01T00:00:00.000Z",
+    version: 2,
+    sessionKey: "session-key",
+    sessionId: "session-id",
+    teamId: "team-1",
+    userId: "user-1",
+    agentId: "agent-1",
+    ...overrides,
+  };
+}
+
+function makeStore(): VectorStore {
+  const dir = mkdtempSync(path.join(tmpdir(), "tdai-cas-"));
+  dirs.push(dir);
+  const store = new VectorStore(path.join(dir, "memory.sqlite"), 0);
+  const init = store.init();
+  expect(store.isDegraded(), init.reason).toBe(false);
+  opened.push(store);
+  return store;
+}
+
+describe("VectorStore.compareAndSwapL1", () => {
+  it("updates exactly once and rejects a stale writer without overwriting content", async () => {
+    const store = makeStore();
+    expect(await store.upsertL1(makeRecord(), undefined)).toBe(true);
+
+    const first = await store.compareAndSwapL1(
+      makeRecord({ content: "first edit", version: 3 }),
+      2,
+      undefined,
+    );
+    expect(first).toEqual({ status: "updated" });
+
+    const stale = await store.compareAndSwapL1(
+      makeRecord({ content: "stale edit", version: 3 }),
+      2,
+      undefined,
+    );
+    expect(stale).toEqual({ status: "conflict", currentVersion: 3 });
+
+    const [persisted] = await store.queryL1Records({ recordIds: ["memory-1"] });
+    expect(persisted.content).toBe("first edit");
+    expect(persisted.version).toBe(3);
+  });
+
+  it("distinguishes a deleted record from a version conflict", async () => {
+    const store = makeStore();
+    const result = await store.compareAndSwapL1(makeRecord({ id: "missing" }), 2, undefined);
+    expect(result).toEqual({ status: "not_found" });
+  });
+});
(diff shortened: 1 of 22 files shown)


Maxwell-Code07:

kvnloo:
Thanks @zealxz for working on this and keeping the compare-and-swap check in storage.

I found a mixed-client compatibility edge at `22fc2f6`. A request that omits `expected_version` calculates its next version from the initial read, then later calls the unconditional upsert. SQLite's existing `version=excluded.version` assignment can therefore move the counter backwards:

1. A legacy update reads version `2`, then pauses during embedding.
2. Two guarded updates succeed in sequence: `2 → 3 → 4`.
3. The legacy update resumes and writes its precomputed version `3`.
4. An old request with `expected_version: 3` now succeeds again, even though that revision was already superseded.

I reproduced `2 → 3 → 4 → 3 → 4` using the extracted handler and the PR's metadata SQL against an in-memory SQLite database. This isn't a regression in the new conditional SQL; it's an existing legacy-write behavior that matters to the new version-token contract.

Should omitting `expected_version` preserve last-write-wins for **content**, while still allocating a strictly increasing version in storage?

One detail from a counterfactual test: changing only the SQL increment prevented token reuse, but left the handler's response and audit at its precomputed `v3` while storage held `v5`. The store-assigned version would also need to flow back to those outputs.

AI-assisted review and isolated reproduction; schema, embedding, and audit dependencies were stubbed. I haven't run full-project CI or a live VectorDB instance.

PR #1138, opened 2026-08-23
[good first issue-platform adapt-Cursor] Add Memory adapter for Cursor

li6667:
# Cursor Proxy Adapter

## Summary

This revision changes #1138 from the original Cursor Hooks + MCP sidecar experiment to the maintainer-requested proxy-based integration. Cursor now shares MemoryProxy's session initialization, system-prompt injection, Skill / Memory / Knowledge assets, conversation archiving, mem commands, cost-guard routing, and Langfuse observability path with the existing clients.

The superseded sidecar implementation has been removed from the proposed upstream diff. Its design remains available in the PR discussion history.

## Protocol reconnaissance

Five sanitized real Cursor requests are included under `MemoryProxy/docs/cursor-recon/proxy-inbound/`.

Confirmed findings:

- Cursor's custom endpoint uses OpenAI Chat Completions with `messages[]`, `tools[]`, streaming, and `stream_options.include_usage`.
- No stable conversation-specific HTTP header was observed. The adapter derives a deterministic fallback from the account-scoped root `user` plus the first content-block user message.
- That fallback cannot distinguish two fresh conversations from the same account when their complete first user turns are identical. This is documented as a protocol limitation, not guaranteed isolation.
- Cursor's native UI tool is `AskQuestion` with `questions[].{id,prompt,options,allow_multiple}` and `options[].{id,label}`.
- Cursor displayed ten explicit options plus `Other...`; pagination is not required for the current four-stage form.
- Cursor replays assistant tool calls without DeepSeek's non-standard `reasoning_content`; the adapter repairs the missing field without overwriting real reasoning content.

The Free IDE-side Connect/Protobuf capture is documented as non-authoritative for Proxy ingress.

## Implementation

- Adds the Cursor adapter, type/factory registration, and Cursor route variants.
- Reuses the shared CodeBuddy state machine for `asset_confirm -> team -> agent -> task`, with Cursor branches gated by `agentSource === "cursor"`.
- Renders Cursor-native `AskQuestion` tool calls.
- Adds Cursor first-frame metadata filtering, headless bypass, and unknown-shape fail-open passthrough.
- Repairs missing `reasoning_content` on Cursor assistant tool-call replay.
- Supports Cursor replay ordering and multi-block content for mem commands.
- Adds Cursor bridge session-key lookup while preserving existing client precedence.
- Injects PowerShell bridge guidance only for Cursor; existing clients retain their previous instructions.
- Removes the unrelated deployment-script workaround used during local E2E troubleshooting.

[...]

PR #1247, opened 2026-09-03
fix(proxy): derive agent URL prefixes from AgentKind registry

### LINKED ISSUE #1246: [Bug] MemoryProxy: agent URL prefix list drifts from the AgentKind registry — dsh/opencode/workbuddy aux endpoints 404
**OpenClaw Version | OpenClaw 版本**
2026.3.13

**Plugin Version | 插件版本**
0.1.0 (MemoryProxy, feat/server_team @ 2ee2239)

**Operating System | 操作系统**
macOS 15 (reproduced from source, OS-independent)

**System Specification | 系统配置**
—

**Describe the bug | 问题描述**
MemoryProxy keeps two parallel agent name lists that must be kept in sync by hand:

1. The `AgentKind` registry (`agent-adapters/types.ts`) — where a new agent is actually registered
2. `AGENT_PREFIX_RE` (`routes/whitelist.ts:175`) — a hardcoded regex alternation that the whitelist path normalizer depends on

Nothing links the two. When an agent lands (adapter + `resolveAgentAdapter` case + `server.ts` routes all added, TypeScript all green), the regex silently misses it. Requests reach the router, but `normalizeWhitelistRequestPath` fails to strip `/{agent}/{spaceId}`, `matchWhitelistEndpoint` misses, and the aux handler 404s — its 404 branch (auxiliaryHandler.ts:170) is literally documented as the alarm for "route/whitelist inconsistency".

This is not a one-off: pi had the exact same gap and was just patched in #1195 (one more name in the regex). dsh / opencode / workbuddy are still broken on main.

The 404 is latent (no confirmed aux traffic; primary endpoints survive via the `?? "/chat/completions"` fallback in handler.ts:1288, coincidentally correct). But the failure mode is structural: **every future agent will miss again** unless the person who adds it happens to remember a regex in a different module.

**To Reproduce | 复现步骤**
On feat/server_team @ 2ee2239, local proxy, auth disabled:

```
curl -X POST http://127.0.0.1:8096/dsh/team-abc/v1/embeddings -H 'content-type: application/json' -d '{"input":"x"}'
curl -X POST http://127.0.0.1:8096/opencode/team-abc/v1/moderations -H 'content-type: application/json' -d '{}'
curl -X POST http://127.0.0.1:8096/workbuddy/team-abc/v1/embeddings -H 'content-type: application/json' -d '{"input":"x"}'
```

All return 404 `{"error":"Unregistered endpoint"}`. Same paths under `/pi/...` work only because #1195 just patched that one name in.

**Expected behavior | 预期行为**
The whitelist prefix list should not be a second hand-maintained registry. It should be derived from (or mechanically tied to) `AgentKind`, so that adding an agent updates the whitelist automatically or fails loudly — not 404 in production.

I have a fix ready (regex derived from the `AgentKind` registry + a regression test that iterates every kind) — will open the PR right after this issue.

**Error Logs / Screenshots | 报错日志/截图**
—

**Additional context | 补充信息**
Related: #1126 (pi adapter), #1195 (pi prefix gap).

L2ncE:
AGENT_PREFIX_RE kept its own hardcoded agent list, separate from the `AgentKind` registry. Nothing linked the two, so every time an agent landed, the regex silently missed it. Reported in #1246: dsh/opencode/workbuddy aux endpoints 404 on main (same class as the pi gap fixed in #1195).

**The gap (reproducible on main)**

```
POST /dsh/team-abc/v1/embeddings      → 404 "Unregistered endpoint"
POST /dsh/team-abc/v1/moderations     → 404
POST /opencode/team-abc/v1/embeddings → 404
POST /workbuddy/team-abc/v1/embeddings → 404
```

Routes are registered in `server.ts`, but the whitelist regex didn't strip these prefixes, so `matchWhitelistEndpoint` missed and the aux handler 404'd. Latent rather than live — no confirmed traffic on these aux endpoints, and primary endpoints survived via the `?? "/chat/completions"` fallback, which happens to be correct for these clients.

**Changes**

- `agent-adapters/types.ts`: `AgentKind` now derives from a runtime `AGENT_KINDS` array (type expands to the identical union). The array is the single source of truth.
- `routes/whitelist.ts`: `AGENT_PREFIX_RE` is constructed from `AGENT_URL_PREFIXES` = all `AGENT_KINDS` (minus `unknown`) + protocol prefixes (`anthropic`/`openai`), with metachar escaping. Registering a new `AgentKind` now updates the whitelist automatically — the #1195 class of fix is no longer needed one name at a time.
- `routes/__tests__/whitelist.test.ts`: one tripwire test iterating `AGENT_KINDS`, asserting each prefix actually strips — a new agent missing from the whitelist fails here.

Notes on deliberate non-changes:

- `cursor` is dropped from the prefix list: no adapter, no docs, injection profile commented out; the open Cursor adapter PR (#1138) adds it to `AgentKind`, which auto-includes it here once merged. Primary `/cursor/...` paths are unaffected (fallback is protocol-correct); only aux endpoints change, and only until #1138 lands.
- workbuddy's primary endpoints are served by `workbuddyHandler`, which strips its own prefix before calling `joinUrl`, so the whitelist never sees a workbuddy prefix on those paths — adding it to the regex cannot double-strip or break them. Only the generic aux routes (the ones this PR fixes) are affected.
- `credit-reporter.ts` keeps its own list (different semantics: header-identity clients hermes/openclaw) — candidate for separate consolidation.

**Verification**

- `npx vitest run` — 9 passed (8 pre-existing + 1 new)
- `npm run typecheck` — 54 pre-existing errors, no new
- End-to-end against a local mock upstream, before → after: dsh/opencode/workbuddy aux endpoints 404 → 200; pi/claude-code primary paths unchanged; forward paths confirmed hitting `/embeddings`, `/moderations` on the upstream.

Fixes #1246

### DIFF
--- a/MemoryProxy/src/agent-adapters/types.ts
+++ b/MemoryProxy/src/agent-adapters/types.ts
@@ -20,7 +20,19 @@
  * 只是**取用户输入的规则**和**分类规则**按 agent 适配。
  */
 
-export type AgentKind = "claude-code" | "codebuddy" | "codex" | "workbuddy" | "dsh" | "opencode" | "pi" | "unknown";
+/** AgentKind 全员名单：type 与下游名单（如 URL 前缀正则）均从此派生。 */
+export const AGENT_KINDS = [
+  "claude-code",
+  "codebuddy",
+  "codex",
+  "workbuddy",
+  "dsh",
+  "opencode",
+  "pi",
+  "unknown",
+] as const;
+
+export type AgentKind = (typeof AGENT_KINDS)[number];
 
 export type RequestKind = "main" | "fork" | "sidequery" | "auxiliary";
 
--- a/MemoryProxy/src/routes/__tests__/whitelist.test.ts
+++ b/MemoryProxy/src/routes/__tests__/whitelist.test.ts
@@ -1,0 +1,15 @@
+import { describe, expect, it } from "vitest";
+import { AGENT_KINDS } from "../../agent-adapters/types.js";
+import { normalizeWhitelistRequestPath } from "../whitelist.js";
+
+describe("AGENT_PREFIX_RE follows AGENT_KINDS", () => {
+  it("strips /{kind} and /{kind}/{spaceId} for every AgentKind except unknown", () => {
+    for (const kind of AGENT_KINDS) {
+      if (kind === "unknown") continue;
+      expect(normalizeWhitelistRequestPath(`/${kind}/team-abc/v1/messages`)).toBe(
+        "/v1/messages",
+      );
+      expect(normalizeWhitelistRequestPath(`/${kind}/v1/messages`)).toBe("/v1/messages");
+    }
+  });
+});
--- a/MemoryProxy/src/routes/__tests__/whitelist.test.ts
+++ b/MemoryProxy/src/routes/__tests__/whitelist.test.ts
@@ -12,4 +12,9 @@ describe("AGENT_PREFIX_RE follows AGENT_KINDS", () => {
       expect(normalizeWhitelistRequestPath(`/${kind}/v1/messages`)).toBe("/v1/messages");
     }
   });
+
+  it("strips /cursor prefix pending #1138 adapter", () => {
+    expect(normalizeWhitelistRequestPath("/cursor/team-abc/v1/embeddings")).toBe("/v1/embeddings");
+    expect(normalizeWhitelistRequestPath("/cursor/v1/completions")).toBe("/v1/completions");
+  });
 });
--- a/MemoryProxy/src/routes/whitelist.ts
+++ b/MemoryProxy/src/routes/whitelist.ts
@@ -9,6 +9,8 @@
  * 新增端点时，只需在 `WHITELIST_ENDPOINTS` 增加一条记录即可，无需散点修改。
  */
 
+import { AGENT_KINDS } from "../agent-adapters/types.js";
+
 /** 白名单端点元数据。 */
 export interface WhitelistEndpoint {
   /**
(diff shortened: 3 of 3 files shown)


Maxwell-Code07:

yangjj-iso (changes requested):
整体方案没问题，dsh、opencode、workbuddy 的 404 已修复。
但新实现删除了原本支持的 `cursor` 前缀，导致 `/cursor/:spaceId/v1/embeddings` 等辅助接口从正常转发变为 404。原因是 `cursor` 不在 `AGENT_KINDS` 中，现有测试也因此覆盖不到。
建议在 `AGENT_URL_PREFIXES` 中暂时保留 `"cursor"`，并补一个 Cursor 回归测试后再合入。其余未发现阻塞问题。

L2ncE:
> 整体方案没问题，dsh、opencode、workbuddy 的 404 已修复。 但新实现删除了原本支持的 `cursor` 前缀，导致 `/cursor/:spaceId/v1/embeddings` 等辅助接口从正常转发变为 404。原因是 `cursor` 不在 `AGENT_KINDS` 中，现有测试也因此覆盖不到。 建议在 `AGENT_URL_PREFIXES` 中暂时保留 `"cursor"`，并补一个 Cursor 回归测试后再合入。其余未发现阻塞问题。

Fixed, PTAL

yangjj-iso (approved):
cursor 已重新保留在 AGENT_URL_PREFIXES，并新增 /cursor/.../embeddings、/cursor/.../completions 回归测试。
dsh、opencode、workbuddy、pi 的辅助端点均能正确归一化；用 mock upstream 验证请求返回 200，且分别转发到 /embeddings、/moderations。
PR 分支测试：10 个测试全部通过；基线分支为 8 个测试通过。
npm run typecheck 当前和基线均为相同的 54 个既有错误，没有新增类型错误。
git diff --check 无问题。

L2ncE:
@Maxwell-Code07 PTAL~

PR #1252, opened 2026-09-03
fix: make services start and work correctly on Windows

interestingyong:
Three cross-platform bugs that only reproduce on Windows (POSIX CI never sees them):

1. MemoryKnowledge never starts as an entry script The bootstrap guard compared import.meta.url against `file://${process.argv[1]}`. On Windows argv[1] is a backslash path (`D:\...\server.ts`), so the URL never matches and the module silently does nothing when run via `node --import tsx
   src/server.ts`. Build the comparison URL with pathToFileURL()   instead, handling both absolute and relative argv paths.

2. Wiki uploads rejected with 400 on Windows wiki-service.ts joined raw/page paths with manual `dir + / +
   name` concatenation and compared them against path.resolve()   output, which uses `\` separators on Windows. Every upload was
   misclassified as a path traversal attempt. Use path.join() so
   separators match the platform.

3. Gateway config.yaml errors were silently swallowed When tdai-gateway.yaml failed to parse, the catch block fell back to env-only config without printing anything, leaving no clue why the file was ignored. Log a warning with the path and parse error to stderr before falling back.

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

### DIFF
--- a/MemoryCore/src/gateway/config.ts
+++ b/MemoryCore/src/gateway/config.ts
@@ -412,8 +412,18 @@ export function loadGatewayConfig(overrides?: Partial<GatewayConfig>): GatewayCo
           : {};
       }
       fileConfig = expandEnvVars(fileConfig) as Record<string, unknown>;
-    } catch {
+    } catch (err) {
       // Config file is optional — malformed files fall back to env-only config.
+      // But never fail silently: an invalid file means EVERY yaml setting
+      // (skill module, llm model, pipeline timings, ...) is ignored and the
+      // gateway runs on pure defaults with no visible trace. Logger is not
+      // available this early, so write straight to stderr.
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

omerfraz:
[install_tencentdb_claude_memory.py](https://github.com/user-attachments/files/31855054/install_tencentdb_claude_memory.py)

omerfraz:
> [install_tencentdb_claude_memory.py](https://github.com/user-attachments/files/31855054/install_tencentdb_claude_memory.py)

Right now this script just enables tencentdb with deepseek api from opencode-go ultimately working for claude code as well.

PR #1257, opened 2026-09-03
chore(deps): bump better-sqlite3 to ^13.0.3 in knowledge & proxy

yonikremer:
## What

Bump **better-sqlite3** `^11.x → ^13.0.3` in **MemoryKnowledge** and **MemoryProxy**, plus the changes a clean install of 13.x requires:

- `MemoryKnowledge/package.json`, `MemoryProxy/package.json` — dependency range bump
- `MemoryKnowledge/start.sh` — Step 4 no longer compiles 11.x from a hard-coded `.pnpm/better-sqlite3@11.10.0` path; it now verifies the packaged prebuild loads, with source compile kept only as a fallback
- `MemoryProxy/pnpm-lock.yaml` + `package-lock.json` — regenerated to 13.0.3

## Why

better-sqlite3 **v13 is the first N-API release**: prebuilt `.node` binaries ship **inside the npm tarball** (`prebuilds/`, sqlite 3.53.4), the `prebuild-install` GitHub-releases download is gone, and the same binary loads across Node majors `>= 22`.

For 11.x, air-gapped / offline installs had to either fetch the prebuilt binary from GitHub releases or compile from source and pin the exact Node ABI. With 13.x, better-sqlite3 installs from an internal npm mirror with zero network and no ABI pin — which is the deployment path this repo targets.

No public API used by either module changed between 11.x and 13.x (constructor, `pragma`, `prepare().run/get/all`, `transaction`, `close`). Embedded SQLite goes 3.45 → 3.53 (feature-additive for the WAL / FTS5 / UPSERT usage in this repo).

## Verification

- `docker build` of both module images (`node:22-slim`) installs 13.0.3 from the tarball — **no local compile** (`better_sqlite3.node` not produced; gyp configure is a no-op via the `prebuild_exists` guard)
- Confirmed the packaged `prebuilds/linux-x64.node` loads under **Node 22** (sqlite 3.53.4, WAL ok)
- Tests green on Node 22 linux (fresh `npm install`): **MemoryKnowledge 12 passed**, **MemoryProxy 8 passed**
- Runtime: Knowledge image boots and `POST /v3/wiki/create|get|delete` round-trips through the DB; Proxy image boots with `storage.backend: sqlite` activated (no degrade) and creates `proxy.db`

## Notes

- Committed with DCO sign-off (`Signed-off-by`).
- The MemoryProxy lockfiles were already stale relative to `package.json` (e.g. missing `cos-nodejs-sdk-v5`); regeneration reflects the correct closure along with the 13.0.3 bump.

### DIFF
--- a/MemoryKnowledge/package.json
+++ b/MemoryKnowledge/package.json
@@ -35,7 +35,7 @@
     "@opentelemetry/api": "^1.9.1",
     "@opentelemetry/sdk-node": "^0.219.0",
     "ai": "^6.0.164",
-    "better-sqlite3": "^11.10.0",
+    "better-sqlite3": "^13.0.3",
     "crc-32": "^1.2.2",
     "dayjs": "^1.11.20",
     "dotenv": "^16.4.0",
--- a/MemoryKnowledge/start.sh
+++ b/MemoryKnowledge/start.sh
@@ -41,15 +41,26 @@ if [ ! -d "node_modules" ]; then
   PYTHON="${PYTHON_BIN}" npm_config_python="${PYTHON_BIN}" pnpm install --ignore-scripts
 fi
 
-# ── Step 4: 编译 better-sqlite3 ──
-echo "[4/5] 编译 better-sqlite3..."
-BETTER_SQLITE3_DIR="node_modules/.pnpm/better-sqlite3@11.10.0/node_modules/better-sqlite3"
-BINDING_GYP="${BETTER_SQLITE3_DIR}/binding.gyp"
-BUILD_NODE="${BETTER_SQLITE3_DIR}/build/Release/better_sqlite3.node"
+# ── Step 4: 验证/准备 better-sqlite3 ──
+# v13 起 better-sqlite3 在 npm 包内自带预编译二进制（N-API，跨 Node 大版本），
+# 安装后通常可直接加载，无需本机编译；仅当平台/glibc 无匹配预编译时才回退源码编译。
+echo "[4/5] 验证 better-sqlite3..."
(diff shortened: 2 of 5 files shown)


yonikremer:
### Heads-up for reviewers: the MemoryProxy `pnpm-lock.yaml` churn includes one pre-existing repair beyond better-sqlite3

The regenerated `MemoryProxy/pnpm-lock.yaml` drops a **dangling workspace entry**:

```yaml
-  '@context-proxy/cost-guard@file:packages/cost-guard':
-    resolution: {directory: packages/cost-guard, type: directory}
```

That entry referenced `file:packages/cost-guard` — a directory that **does not exist** in this repo, and `@context-proxy/cost-guard` is **not declared** in `MemoryProxy/package.json`. The lock could not be frozen-installed with it, so regeneration (required to record better-sqlite3 13.0.3) removed it. It also adds `cos-nodejs-sdk-v5@3.0.0`, which `package.json` already requires but the old lock was missing.

Two things this PR does **not** change, both pre-existing on `feat/server_team`:

1. `MemoryProxy/src/storage/factory.ts` still contains an `import`/require of `@context-proxy/cost-guard` (the cos-backend "CostGuard" routing), which resolves to nothing today.
2. That feature was already non-functional upstream — this PR only stops the lockfile from pointing at a package directory that isn't there.

Happy to split the lock repair into a separate PR if preferred; otherwise it's an unavoidable consequence of re-locking to bump better-sqlite3. The only better-sqlite3-relevant lock deltas are: `better-sqlite3@11.x → 13.0.3`, removal of the `prebuild-install`/`bindings` transitive closure, and the new `node-addon-api` dependency.

Maxwell-Code07:

PR #1258, opened 2026-09-03
fix(deploy): honor PROXY_UPSTREAM_PROTOCOL in LLM connectivity checks

Liubing-Yu:
## Summary
- `verify.sh` and `_lib.sh`'s `interactive_llm_setup` both hardcoded `openai` as the protocol for the proxy group's upstream precheck, ignoring `PROXY_UPSTREAM_PROTOCOL` from `.env` — this produced a spurious `GET /models 404` warning on every start for anthropic-protocol upstreams (e.g. `PROXY_UPSTREAM_URL=https://api.anthropic.com`).
- `interactive_llm_setup` also never persisted `PROXY_UPSTREAM_PROTOCOL` back to `.env`, so it silently reverted to the `openai` default on every interactive run.
- `check_llm_from_container`'s anthropic branch unconditionally appended `/v1/messages`, even when the base URL already ended in `/v1`, producing a double `/v1/v1` 404 during the container-reachability check.

## Test plan
- [x] `./verify.sh` with `PROXY_UPSTREAM_PROTOCOL=anthropic` and `PROXY_UPSTREAM_URL=https://api.anthropic.com/v1` now reports `协议=anthropic` for the proxy group and passes with no warnings, both from the host and from-container checks.
- [x] `./start-all.sh`'s interactive flow now correctly prompts/reuses the anthropic protocol for the proxy group and writes `PROXY_UPSTREAM_PROTOCOL` back to `.env`.
- [x] End-to-end: a real `/v1/messages` call through the running `tdai-proxy` container against an Anthropic upstream returns `200` with a valid completion.

### DIFF
--- a/deploy/global-images/_lib.sh
+++ b/deploy/global-images/_lib.sh
@@ -368,17 +368,20 @@ interactive_llm_setup() {
     PROXY_UPSTREAM_URL="$MEMORY_LLM_BASE_URL"
     PROXY_UPSTREAM_API_KEY="$MEMORY_LLM_API_KEY"
     PROXY_UPSTREAM_MODEL="$MEMORY_LLM_MODEL"
+    PROXY_UPSTREAM_PROTOCOL="$MEMORY_LLM_PROTOCOL"
     ok "proxy 组复用 memory 组配置，跳过重复检查"
   else
     while true; do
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1260, opened 2026-09-04
fix(proxy): bridge session resolution + make Pi a first-class client

nicwn:
## Problem

The memory-bridge and skill-bridge read path is broken for any agent source other than `codebuddy` and `claude-code`. When an agent (pi, dsh, codex, workbuddy, opencode) tries to search its L1/L0 memory or load skills via the bridge, it receives `40101 "session not initialized"` — even though the session was successfully initialized and the agent has persona/skills injected.

This affects **every default-install user** (the 3-container `start-all.sh` deploy with no Redis and no `storage:` section), not just custom deployments.

### Bug 1: BindingRepo never activated in default install

The default install generates `redis: enabled: false` and no `storage:` section. The eager activation block in `server.ts` only calls `tryActivateStorage` and `tryActivateRedis` — both silently no-op when their config flags are false.

`ensureBindingRepoPersistent` (the fs fallback) is only called inside `buildPipelineBundle`, which runs **lazily on the first main request** — after `completeRegistration` → `store.set()` already skipped the L2b binding write because `this.bindingRepo` was still `null`.

Result: the L2b binding (designed as the bridge's L2 backstop, keyed by `(spaceId, sessionId)`) is never written. The bridge L2 lookup always misses.

### Bug 2: Bridge L1 lookup uses a hardcoded agent-source prefix list

`loadSessionIdsL1` in both `memory-bridge.ts` and `skill-bridge.ts` tries these candidate keys:
```typescript
const candidates = sessionId.includes(":")
  ? [sessionId]
  : [sessionId, `codebuddy:${sessionId}`, `claude-code:${sessionId}`];
```

The `pi:` prefix (and `dsh:`, `codex:`, `workbuddy:`, `opencode:`) is missing. The L1 state IS in the `states` Map under `pi:sessionId`, but the bridge never looks there.

### Fix 1: Eagerly activate BindingRepo at startup

Add `ensureBindingRepoPersistent(config)` to the `server.ts` eager block, right after `tryActivateStorage`/`tryActivateRedis`. It is idempotent (no-ops if a BindingRepo was already set by storage/redis).

This ensures the BindingRepo is ready before any request can trigger `completeRegistration`, so the L2b binding write actually happens.

### Fix 2: Enumerate L1 keys by suffix instead of hardcoded prefixes

Add `SessionStore.keysWithSuffix(sessionId)` — returns all L1 keys matching `:${sessionId}` (or the bare sessionId). Both bridges use it in `loadSessionIdsL1`:

```typescript
function loadSessionIdsL1(sessionId: string): SessionIdFields | null {
  const store = getSessionStore();
  const bare = store.get(sessionId);
  if (bare) { /* ... */ }
  if (!sessionId.includes(":")) {
    for (const k of store.keysWithSuffix(sessionId)) {
      const state = store.get(k);
      if (state) { /* ... */ }
    }
  }
  return null;
}
```

This works for **any** current or future agentSource without maintaining a hardcoded list.

## Impact

- **No breaking changes.** The L2b binding write and L1 lookup are additive — existing codebuddy/claude-code sessions continue to work exactly as before.
- **No new dependencies.** Uses the existing `SessionStore.states` Map.
- **Default install now functional.** The 3-container deploy (no Redis, no storage) now has a working bridge read path.

## Testing

6 new tests in `src/__tests__/bridge-session-resolution.test.ts`:
- `keysWithSuffix`: bare match, multi-agent-source, no-match, substring-vs-suffix
- Integration: pi-prefixed and dsh-prefixed sessions found with bare sessionId

All 14 tests pass (8 existing + 6 new).

## Files Changed

| File | Change |
|------|--------|
| `src/server.ts` | Add `ensureBindingRepoPersistent(config)` to eager activation |
| `src/session/store.ts` | Add `keysWithSuffix(sessionId)` method |
| `src/memory/memory-bridge.ts` | Use `keysWithSuffix` in `loadSessionIdsL1` |
| `src/skill/skill-bridge.ts` | Use `keysWithSuffix` in `loadSessionIdsL1` |
| `src/__tests__/bridge-session-resolution.test.ts` | New test file

### DIFF
--- a/MemoryCore/pi-plugin/README.md
+++ b/MemoryCore/pi-plugin/README.md
@@ -5,11 +5,22 @@ through the [TencentDB Agent Memory v2](https://github.com/TencentCloud/TencentD
 proxy for team memory: L3 persona, L2 scene index, L0 conversation capture, and
 on-demand L0/L1/L2 search.
 
-The plugin carries **only routing + a dynamic per-session `x-conversation-id`
-header**. All memory capability is delivered server-side by the proxy
-(injected into the system prompt and captured from the response). This keeps
-the extension minimal and keeps memory logic in one place. (Scope: routing +
-header; no client-side recall/capture.)
+The plugin carries **routing + a dynamic per-session `x-conversation-id` header + an interactive Team/Agent/Task picker**. All memory capability is delivered server-side by the proxy (injected into the system prompt and captured from the response). This keeps the extension minimal and keeps memory logic in one place. (Scope: routing + header + picker; no client-side recall/capture.)
+
+## Interactive session init (default)
+
+On the **first turn of each new Pi session**, if no preset identity is set (see
+below), the proxy sends a Team → Agent → Task form and this plugin renders it as
+a Pi-native TUI menu: `↑↓` to navigate, `Enter` to select, `Esc` to cancel. Pick
+once and the session is bound for its lifetime — the chosen team's memory (L3
+persona, L2 scene index) is injected and the conversation is captured (L0).
+
+This matches the Claude Code / CodeBuddy UX. The picking is **manual**, by
+design — which team/agent a piece of work belongs to is a human-intent decision,
+not auto-inferred from your prompt.
+
+Requires interactive mode (the default `pi` TUI). In `pi -p` / RPC / non-TUI
+modes the picker cannot render; set the preset env vars instead (below).
 
 ## Prerequisites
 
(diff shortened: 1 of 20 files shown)


Maxwell-Code07:

Movtrength (commented):
## Review (local verification)

Verified on top of current `feat/server_team` (merge is clean; PR was 3 commits behind: #1195 / #1222 / #1052).

### What looks good
- **Bug 1 fix** — eager `ensureBindingRepoPersistent(config)` in `server.ts` is the right place; idempotent and matches the lazy path in `buildPipelineBundle`.
- **Bug 2 fix** — `SessionStore.keysWithSuffix` + bridge `loadSessionIdsL1` correctly drops the hardcoded `codebuddy`/`claude-code` prefix list. This is the durable fix vs forever extending the candidate list.
- **Pi first-class surface** — docs / `setup-proxy.sh` / asset-import / pi-plugin tests are coherent with the earlier Pi adapter work.
- **Tests** — pi-plugin: 30/30 pass. Original bridge unit tests: 6/6 pass after install.

### Nits / follow-ups before un-draft
1. **Please merge/rebase onto latest `feat/server_team`** — no conflicts; needed for #1195 (`pi` in `AGENT_PREFIX_RE`).
2. **Bridge “integration” tests don’t call the bridge** — they only exercise `keysWithSuffix`, and `session_id: ""` would make real `toIdFields` return `null`. I have a local patch that drives `createMemoryBridgeHandler` with a mocked fetcher (pi/dsh hit → 200 upstream, miss → 40101). Happy to push it if you add write access / cherry-pick, or I can paste a follow-up PR once there’s a fork to target.
3. **Ambiguous multi-prefix** — if both `pi:abc` and `dsh:abc` exist, first Map hit wins. Probably fine for real traffic (one agent source per conversation), but worth a one-line comment.
4. **Scope** — bridge fix + Pi onboarding are independently reviewable; splitting would speed merge of the bugfix if Pi docs need more bikeshedding.

### Ready?
Approach LGTM for the bridge bugs. After syncing with `feat/server_team` (+ ideally the real handler-level tests), this can leave draft.

Movtrength:
Pushed a companion with write access on my fork: https://github.com/TencentCloud/TencentDB-Agent-Memory/pull/1265

Includes your two commits + merge of current `feat/server_team` + handler-level bridge tests (`087d4de`). Cherry-pick that commit here if you want this draft to stay the landing PR.

nicwn:
Thanks @Movtrength — cherry-picked your handler-level test commit (`53b0da5`, from your #1265 companion) and added the one-line multi-prefix first-hit-wins note at both enumeration sites (`6542ecb`). Verified against current `feat/server_team` (0 commits behind, clean merge base). Full suite: 15/15 passing.

On nit 4 (splitting): keeping this unified — the bridge fix and the Pi onboarding landed as one first-class-client effort and the maintainer's internal review is already scoped to this PR as a unit. Happy to split later if the review stalls on the docs half. Companion #1265 can be closed whenever you like — this PR carries your test commit forward with credit.

PR #737, opened 2026-08-03
fix(knowledge): wake stopped auto-sync workers

RerankerGuo:
## Description | 描述

The v2.0.0 auto-sync scheduler keeps its idle workers alive by awaiting short `setTimeout` sleeps. `stop()` clears those timers, but the associated Promises are never resolved. The worker loops therefore never exit or decrement `workerCount`, and repeated stop/start cycles accumulate suspended workers.

This PR stores a wake callback with each idle timer and resolves cancelled sleeps during shutdown. Each scheduler run also receives a generation token, preventing an old worker from joining a new pool when callers execute `stop(); start()` before the old continuation resumes.

Fake-timer tests cover:

- a normal stop waking all idle workers and clearing every timer;
- an immediate stop/start retaining exactly the configured worker count, followed by a clean final stop.

## Related Issue | 关联 Issue

No existing issue. Discovered during a v2.0.0 default-branch Knowledge scheduler lifecycle audit.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [ ] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

## Verification | 验证

- `pnpm vitest run src/store/auto-sync-scheduler.test.ts` - 1 file, 2 tests passed
- `pnpm test` - 1 file, 2 tests passed
- strict changed-file TypeScript check - passed
- `pnpm build` - 11 output files built
- `git diff --check` - passed

The full-package typecheck still reaches the unchanged `src/middleware/response-envelope.ts:47` base-branch error, which is fixed separately by #730. The changed scheduler and test pass an isolated strict TypeScript check.

## Additional Notes | 其他说明

- In-progress sync operations still finish naturally after stop, matching the existing contract.
- Queue contents, `inFlight` deduplication, scan cadence, and max concurrency are unchanged.
- Full Open PR title/body and changed-file scans found no active implementation touching the auto-sync scheduler.

Signed-off-by: RerankerGuo <121015044+RerankerGuo@users.noreply.github.com>

PR #1262, opened 2026-09-05
fix(knowledge): enforce auto-sync build concurrency

zealxz:
## Description | 描述

`AutoSyncScheduler.maxConcurrentSyncs` currently limits scheduler workers, but `CodeGraphService.sync()` returns immediately after enqueueing a per-asset `BuildQueue` job. The scheduler therefore releases its slot before the actual clone/index build finishes, and multiple real builds can exceed the configured limit.

This change waits for `CodeGraphService.onIdle(code_graph_id)` after a successful enqueue, so a scheduler slot remains occupied until that asset's build completes. It also adds a fake-timer regression test for configured limits 1, 2, and 3.

## Related Issue | 关联 Issue

N/A — no exact existing issue was found for this enqueue-vs-completion concurrency gap. Existing #737 concerns waking stopped auto-sync workers and is a different failure mode.

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Verification

- Baseline regression: with six ready repositories, configured limits 1/2/3 all started six builds.
- Patched regression: peak build concurrency never exceeded the configured limit; 3/3 tests passed.
- `MemoryKnowledge`: `pnpm run build` passed.
- `git diff --cached --check` passed.
- Commit includes DCO sign-off.

### DIFF
--- a/MemoryKnowledge/src/store/auto-sync-scheduler.test.ts
+++ b/MemoryKnowledge/src/store/auto-sync-scheduler.test.ts
@@ -1,0 +1,73 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { AutoSyncScheduler } from "./auto-sync-scheduler.js";
+import { CodeGraphService } from "./code-graph-service.js";
+import type { CodeGraphRow, IKnowledgeStore } from "./types.js";
+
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #542, opened 2026-07-20
feat(adapters): add Codex, Claude Code, and OpenCode integrations

Szkkhh:
## Description | 描述

This PR completes all four progressive acceptance stages in #235. It documents the existing memory architecture, adds Gateway-backed integrations for Codex, Claude Code, and OpenCode, and introduces a public lifecycle Adapter SDK so a new platform only needs to implement one `PlatformAdapter<TBindings>` interface.

本 PR 完成 #235 的四档渐进式验收：补充现有记忆架构说明，为 Codex、Claude Code 与 OpenCode 增加 Gateway-backed 接入，并公开统一生命周期 Adapter SDK，使新平台只需实现一个 `PlatformAdapter<TBindings>` 接口。

### Core architecture and documentation | 核心架构与文档

- Documents `TdaiCore` recall, capture, search, session-end, and lifecycle responsibilities.
- Compares OpenClaw in-process integration with Hermes through the HTTP Gateway.
- Adds Mermaid diagrams for automatic recall, active search, and the L0 -> L1 -> L2 -> L3 memory pipeline.
- Defines the responsibility boundaries of `HostAdapter`, `LLMRunnerFactory`, Gateway, `IMemoryStore`, MCP, and lifecycle platform adapters.

### Shared MCP adapter | 共享 MCP adapter

Adds a Gateway-backed stdio MCP server with the following tools:

| Tool | Gateway endpoint | Purpose |
| --- | --- | --- |
| `tdai_memory_recall` | `POST /recall` | Recall relevant long-term memory |
| `tdai_memory_capture` | `POST /capture` | Capture a completed user/assistant turn |
| `tdai_session_end` | `POST /session/end` | Flush queued work for one session |
| `tdai_memory_search` | `POST /search/memories` | Search L1 structured memory |
| `tdai_conversation_search` | `POST /search/conversations` | Search L0 conversation history |

The MCP adapter remains the model-facing protocol entry point. It reuses the existing Gateway and does not introduce another memory core or store. Gateway URL and Bearer token are configurable through `TDAI_GATEWAY_URL` and `TDAI_GATEWAY_API_KEY`.

One Gateway instance currently represents one memory namespace. This PR intentionally does not expose a `TDAI_USER_ID` option because the existing Gateway does not provide user-level namespace isolation across recall, capture, search, and session-end.

### Lifecycle Adapter SDK | 生命周期 Adapter SDK

Adds the public npm subpath:

```ts
import {
  createAdapterRuntime,
[...]

PR #1267, opened 2026-09-06
feat(adapters): wiki tools for the MCP adapter + transcript capture for Claude Code (on top of #542)

plainwuatlig:
## Description | 描述

Two additions to the platform adapters introduced in #542. This PR is stacked on #542: the first four commits are that PR's adapter tree; the last two are this PR's own change (`07e5ffe`, `5bc9912`).

**1. MCP adapter: Knowledge Service (wiki) tools**

The stdio MCP server exposed only the memory Gateway, so agents on hosts such as Claude Code could not reach the team wiki without hand-rolled HTTP. This adds an organisation-agnostic Knowledge Service client and five tools, named the same as in the pi extension so behaviour is portable across hosts:

- `tdai_wiki_list`, `tdai_wiki_search`, `tdai_wiki_pages`, `tdai_wiki_read`, `tdai_wiki_write` → `POST /v3/wiki/{list,search,page/ls,page/read,page/write}`
- Wiki ids are always tool parameters. Service and tenant identity come from `TDAI_KNOWLEDGE_URL`, `TDAI_SERVICE_ID`, `TDAI_TEAM_ID`, `TDAI_USER_ID`, `TDAI_AGENT_ID`; `TDAI_KNOWLEDGE_API_KEY` falls back to `TDAI_GATEWAY_API_KEY`.
- Responses are unwrapped from the `{ code, message, data }` envelope; a non-zero `code` becomes a tool error. Read/write batches are capped at 20 per call.
- `createMemoryMcpServer({ knowledge: false })` keeps the previous five-tool surface; `createKnowledgeTools()` is exported for platform adapters.

**2. Claude Code adapter: capture the session transcript, not only prompt + reply**

The Stop hook sent one turn as prompt + final assistant message, so tool calls, tool results and intermediate assistant text never reached L0, and other agents could not see what a Claude Code session actually did. Both `Stop` and `SessionEnd` now read the transcript named by `transcript_path` in the hook payload (fallback: the Claude Code projects directory) and send the entries after a per-session marker as L0 messages:

- `Stop` sends the finished turn as one capture (prompt, tool calls, tool results, intermediate text, final message) with the prompt and final message as `user_content` / `assistant_content`. `SessionEnd` sends whatever is left. Per-turn sending matters: a 100-message Gateway batch takes about 5 s and a session's SessionEnd fires once, so an end-only capture loses a long session's tail (measured: 4 of 5 batches inside a 20 s budget on a 434-message session).
- Policy mirrors the pi extension: thinking and images dropped; `tool_use` / `tool_result` quoted as prefixed text with size caps; 8192-char chunks; 100 messages per batch.
- Credential-shaped substrings are redacted before sending (private keys, bearer tokens, `sk-` keys, GitHub/GitLab/Slack/AWS/Google/npm tokens, `password=` values). Tool traffic routinely carries env dumps and config files.
- Turns captured by the plain path contribute only their tool traffic later; the Gateway records what it is sent and does not deduplicate by message id. A failed batch leaves the marker at the last batch that landed and the turn unclaimed, so `SessionEnd` sends the rest.
- Transcript batches use their own Gateway client (15 s timeout); recall keeps 3 s. Kill switch `TDAI_CLAUDE_CODE_TRANSCRIPT_CAPTURE=off`. Sample hook timeouts raised to 15 s (Stop) and 30 s (SessionEnd) so a batch in flight can write its marker.

Docs updated in `docs/mcp.md`, `docs/claude-code.md` (both with `_CN`), `integrations/claude-code/hooks.json`, `CHANGELOG.md`. Both commits carry a DCO sign-off.

## Related Issue | 关联 Issue

Depends on #542 (stacked on it). Also opened against the #542 author's fork as Szkkhh/TencentDB-Agent-Memory#1 and #2, in case it is easier to fold in there first.

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [x] New feature | 新功能

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过 — `vitest run`: 176 tests pass (new: `knowledge.test.ts`, `transcript.test.ts`, `transcript-capture.test.ts`, MCP round-trips in `server.test.ts`); `tsdown` build passes.
- [x] No existing features affected | 无影响现有功能 — the five Gateway tools and their tests are unchanged; `knowledge: false` and `TDAI_CLAUDE_CODE_TRANSCRIPT_CAPTURE=off` restore the previous behaviour exactly.

## Additional Notes | 其他说明

Live verification against a running Gateway and Knowledge Service: the stdio server started by Claude Code lists 10 tools and wiki list/search/read/pages return real data; a Stop run completed in 5.3 s with the marker advanced, and tool calls plus `[redacted:…]` markers are searchable in L0 with no raw credential in the hits.

One observation for maintainers, not changed here: a 100-message `/capture` costs roughly 5 s on the Gateway, which is why the transcript is sent per turn rather than at session end.


plainwuatlig:
Superseded by #1268, which re-implements the same Claude Code integration as a native v3 client (`MemoryCore/claude-code-plugin/`) against the current default branch `feat/server_team`, in the shape of the OpenClaw client adapter. This PR targeted the v1 `main` line and depended on #542; closing it to keep one open change per topic.

PR #1290, opened 2026-09-08
feat(proxy): add native [OI]-compatible /v1/models endpoint

### LINKED ISSUE #1287: feat(MemoryProxy): add native OpenAI-compatible /v1/models endpoint
## Problem

`MemoryProxy` currently forwards the main chat endpoints, but it does not expose a native model-list endpoint. OpenAI-compatible clients commonly call `GET /v1/models` during startup or model selection, so clients configured to use MemoryProxy cannot discover available models through the proxy.

The current endpoint documentation lists `/v1/chat/completions` and `/v1/messages`, but no `/models` endpoint. The existing `/direct/*` passthrough can be used as a diagnostic workaround, for example:

```bash
curl http://localhost:8096/direct/v1/models \
  -H "Authorization: Bearer sk-mem-xxx"
```

However, `/direct/*` intentionally bypasses session initialization, memory injection, conversation write-back, and related MemoryProxy processing. It should remain a test/diagnostic channel rather than the normal client route.

## Proposed solution

Add a native OpenAI-compatible model-list route to `MemoryProxy`:

```http
GET /proxy/<spaceId>/v1/models
```

The route should forward the request to the configured upstream model-list endpoint and return the upstream response without applying chat-specific memory injection or conversation extraction.

Please also consider whether the following compatibility route should be supported:

```http
GET /v1/models
```

If supported, its behavior should be documented clearly because it has no `spaceId` and therefore cannot provide the same memory-scoped context as `/proxy/<spaceId>/...` routes.

## Expected behavior

- Preserve the upstream HTTP status, response body, and relevant headers.
- Forward the configured upstream API key/authentication according to the existing Proxy configuration rules.
- Support the standard OpenAI response shape, including `object: "list"` and `data` model entries.
- Do not route the request through the chat completion or Anthropic message handlers.
- Document the endpoint and an example client/curl invocation in `MemoryProxy/README.md` and `MemoryProxy/README_CN.md`.
- Add route/forwarding tests, including upstream errors and authentication behavior.

## Alternatives considered

1. Add an Nginx reverse proxy in front of MemoryProxy: no application code changes, but adds an extra deployment/network layer.
2. Require clients to use `/direct/v1/models`: available today, but changes the client path and bypasses the normal Proxy routing model.
3. Connect clients directly to the upstream LLM: simplest for model discovery, but does not provide a unified MemoryProxy endpoint.

## Context

The implementation likely belongs near the route registration in `MemoryProxy/src/server.ts`, reusing the existing upstream forwarding behavior where possible. The `/direct/*` path handling is implemented in `MemoryProxy/src/directHandler.ts` and is useful as a reference for transparent forwarding, but should not be treated as the recommended user-facing memory route.

cjb1234567:
Closes #1287

## Summary

Adds a native model-list endpoint so [OI]-compatible clients configured against MemoryProxy can discover available models through the proxy, without switching their base URL to the `/direct/*` diagnostic channel.

## Changes

- **`src/modelsHandler.ts`** (new): lightweight `GET` passthrough that injects the configured upstream credentials (replacing the client proxy-layer key) and forwards upstream status / headers / body verbatim. Skips session-init, context injection, conversation write-back, routing and credit reporting — a model list is not a conversation turn.
- **`src/routes/whitelist.ts`**: add `/v1/models` → `/models` mapping, and switch the agent-prefix regex from a hardcoded allowlist to a **reserved-word exclusion** (`v1` / `proxy` / `skill-bridge` / `memory-bridge`), matching the `agentFromPath` logic in `handler.ts` / `anthropicHandler.ts`. This also fixes `/v1/models` and auxiliary endpoints (`embeddings` / `completions` / `moderations` / `count_tokens`) for agents missing from the old allowlist (`dsh` / `opencode` / `workbuddy` / `pi` / `hermes` / `openclaw`) and keeps future agents working without touching this regex.
- **`src/server.ts`**: register `GET /v1/models` plus the `/:agent/:spaceId/v1/models`, `/:agent/v1/models` and `/proxy/:spaceId/v1/models` variants.
- **`README.md` / `README_CN.md`**: document the new endpoint.

## Supported paths

```http
GET /v1/models
GET /proxy/<spaceId>/v1/models
GET /{agent}/{spaceId}/v1/models
GET /{agent}/v1/models
```

## Tests

- New `src/routes/__tests__/whitelist.test.ts` (11 cases): whitelist matching for all agent prefixes, reserved-word non-stripping, unknown-agent support, and `joinUrl` mapping to `/models` (not the `/chat/completions` fallback).
- Full suite: 19 tests passing.

## Verification

- `npm run typecheck` — no new errors in touched files.
- Deployed to a live instance against a real `[OI]` upstream; all supported agent prefixes return `200` with the model list.


Maxwell-Code07:

PR #999, opened 2026-08-15
[good first issue-platform adapt-OpenCode] Add Memory adapter for OpenCode

LeonSGP43:
## What / 做了什么

Adds `adapters/opencode/` — a TencentDB Agent Memory adapter for [OpenCode](https://opencode.ai), per #926.

新增 `adapters/opencode/` 目录：OpenCode 的 TencentDB Agent Memory 适配器（对应 #926 本期优先方向）。

## How it works / 工作原理

OpenCode connects through the proxy's OpenAI-compatible endpoint (`/codebuddy/<spaceId>`) using a custom `@ai-sdk/openai-compatible` provider — configuration only, no plugin code required. Each session then gets:

- **Session binding** — Team → Agent → Task picker on first message
- **Memory injection** — bound agent's L2/L3 memory, skills and knowledge blended into the system prompt every turn
- **Automatic capture** — L0 raw dialogue persisted into memory-core

OpenCode 通过自定义 `@ai-sdk/openai-compatible` provider 接入代理的 OpenAI 兼容端点，纯配置接入、无需插件代码。每个会话自动获得：会话绑定（Team → Agent → Task 选择器）、L2/L3 记忆注入、L0 原始对话沉淀。

## Files / 文件

| File | Purpose |
|---|---|
| `opencode.json` | Provider config (baseURL, service header, model mapping) |
| `README.md` | English setup guide, config reference, troubleshooting |
| `README_CN.md` | 中文配置指南（同一 PR 双语，满足 CI 校验） |

## Notes / 说明

- The model ID under `models` must match the proxy's `PROXY_UPSTREAM_MODEL`; documented in both READMEs.
- When a dedicated `/opencode/<spaceId>` endpoint lands upstream, only `options.baseURL` needs to change (noted in README).
- Tested with OpenCode ≥ 0.6 custom-provider config path.

Closes #926 (OpenCode direction)

---
- [x] Both EN and CN docs included in this PR / 中英文档同一 PR 提交
- [x] Standalone directory per adapter / 适配器独立目录
- [x] PR title follows the required format / PR 标题符合格式要求
- [x] DCO signed / 已 DCO 签名

PR #1291, opened 2026-09-08
fix(proxy): register opencode in AGENT_PREFIX_RE

kristhianmanue1:
## Summary

The first-class opencode adapter (`agent-adapters/opencode.ts`, present since v2.0.1) handles `/opencode/{spaceId}/v1/...` requests, but `AGENT_PREFIX_RE` in `routes/whitelist.ts` still lacks `opencode`. Same gap as #1195 fixed for `pi`: cost-guard / analyse markers on `/opencode` paths are not recognized, and whitelist matching misses after prefix normalization.

## Change

Add `opencode` to the agent alternation in `AGENT_PREFIX_RE` (1 line) + vitest regression coverage (normalization ×4, markers ×2, whitelist matching ×1), mirroring the pi coverage pattern from #1195.

## Live verification

Verified E2E against a running proxy (v2.0.1 image, upstream `feat/server_team` line):

- `POST /opencode/{spaceId}/v1/messages` → session-init intercepts with the native opencode `question` tool (`call_oc_session_init_*`), completing all four stages (asset → team → agent → task): final state `initialized agent=agt-... team=team-...`
- Injection hooks fire for the opencode agent source (skill + tdai-memory + profile blocks)
- L0 conversation rows recorded for the opencode session

One observation from that live run, related but out of scope here: an auth/verify timeout mid-form resets the session-init state machine and consumes `maxRetries`, leaving the session silently bypassed (no injection / no registration). Happy to file a separate issue if maintainers agree it is a real failure mode.

## Tests

`npx vitest run` → 7/7 passing (new file). `tsc --noEmit` delta vs upstream tip `220af62`: zero new errors.

### DIFF
--- a/MemoryProxy/src/__tests__/memory-bridge-identity.test.ts
+++ b/MemoryProxy/src/__tests__/memory-bridge-identity.test.ts
@@ -1,0 +1,292 @@
+/**
+ * Fresh-context spec tests — memory-bridge identity resolution.
+ *
+ * Written against the endpoint SPEC (not the implementation) by a fresh
+ * subagent (Corolario C, ronda 2026-09-09); caso (d) de colisión proviene
+ * del hallazgo HIGH del crítico codex.
+ *
+ * The session store singleton keys state as `${agentSource}:${sessionId}`
+ * (e.g. `opencode:ses_abc`, `claude-code:ses_abc`). The handler must resolve
+ * the caller's identity from a seeded session under either prefix when the
+ * bare sid arrives, reject with 40101 when it exists under no key, and
+ * REJECT (40901) when a bare id matches several distinct identities.
+ *
+ * Deliberately does NOT go through createApp() (server.ts) so the route layer
+ * is out of scope; the handler is mounted exactly as server.ts registers it.
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi, type Mock } from "vitest";
+import { Hono } from "hono";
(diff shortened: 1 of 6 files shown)


Maxwell-Code07:

kristhianmanue1:
Hi @Maxwell-Code07 — following up with a field report related to #999 (merged): it routes opencode users to `/opencode/<spaceId>/v1/...` on this proxy.

We tested both client variants behind the proxy:
- **Anthropic-compatible** (`@ai-sdk/anthropic` → `/opencode/<spaceId>/v1/messages`): works end-to-end.
- **OpenAI-compatible** (opencode's `openai`-mode baseURL pointing at the same `/opencode/<spaceId>/v1` prefix): accepted by the proxy, but with an Anthropic per-agent upstream the chat-completions route 404s; the Anthropic-variant path is the one that works.

#1291 (whitelisting `opencode` in `AGENT_PREFIX_RE`) is what lets session identity resolution (memory/skill bridges) resolve for sessions arriving via either variant — without it, `/opencode/...` sessions fall outside the agent-source whitelist and bridge reads fail with `session not initialized`. With #999 merged, opencode users hit this path by default, so in our testing #1291 becomes a natural companion. Happy to adjust per internal review.

PS: the related proxy→core auth fix (internal gateway Bearer on `auth/verify`, with test + deploy wiring) now lives in #1211.

PR #1312, opened 2026-09-09
feat(l2): deterministic scene maintenance (semantic dedup + heat decay + cold eviction)

kevinyuan:
## Problem

The scene (L2) set is bounded only by the LLM's **reactive** MERGE at the hard `persona.maxScenes` cap ("merge the coldest / most similar"). `heat` is a monotonic hit counter with **no time decay**, and there is **no eviction** — so the set skews toward whatever is recalled most and collapses topic diversity. Because the persona (L3) is generated from scenes, it then narrows to those few topics. Raising `maxScenes` only postpones the collapse.

## Change

Add a deterministic maintenance pass (`runSceneMaintenance`) that runs after each L2 extract, using the store's embedding service (local model → no extra remote cost):

- **semantic dedup**: near-duplicate scenes (cosine ≥ `SCENE_DEDUP_THRESHOLD`) collapse into the highest-priority anchor, which absorbs the losers' heat; losers are archived under `.archive/scene_blocks/` (not deleted).
- **heat decay**: rank by `decayedHeat = heat * 0.5^(ageDays / halfLife)` so a scene untouched for long cools even if once hot (stored heat unchanged; an UPDATE refreshes `updated`, i.e. recency).
- **cold eviction**: if survivors still exceed `SCENE_SOFT_CAP` (default 80% of `maxScenes`, leaving headroom for the LLM to CREATE diverse scenes), archive the lowest-`decayedHeat` scenes down to the cap.

It runs unconditionally after extract (reads the on-disk scene set; safe even when the L2 LLM call failed) and only rewrites the index when something changed. When no embedding service is available it degrades to decay + eviction only.

Tunables via env: `SCENE_DEDUP_THRESHOLD` (0.88), `SCENE_HEAT_HALFLIFE_DAYS` (14), `SCENE_SOFT_CAP`.

## Testing

Verified live: on a 15-scene set with aggressive settings (`SCENE_DEDUP_THRESHOLD=0.75`, `SCENE_SOFT_CAP=8`) the pass reported `deduped=1, evicted=6, remaining=8` with losers archived; at production defaults (0.88 / softCap=32) a 15-scene set is untouched. After enabling it, a cleared scene set rebuilt into a diverse, bounded set and the regenerated persona captured previously-dropped topics.

### DIFF
--- a/MemoryCore/src/utils/pipeline-factory.ts
+++ b/MemoryCore/src/utils/pipeline-factory.ts
@@ -700,6 +700,168 @@ export function createPersister(
  *
  * Used by both `index.ts` (live runtime) and `seed-runtime.ts` (seed CLI).
  */
+
+// ────────────────────────────────────────────────────────────────────────────
+// Scene maintenance — semantic dedup + heat decay + cold eviction.
+//
+// The stock L2 flow bounds the scene set only via the LLM's *reactive* MERGE at
+// the hard `persona.maxScenes` cap ("merge the coldest / most similar"). `heat`
+// is a monotonic hit counter with no time decay, and there is no eviction, so
+// the set skews toward whatever is recalled most and loses topic diversity —
+// and since the persona (L3) is built from scenes, it narrows to those topics.
+// Raising maxScenes only delays this.
+//
+// This deterministic pass runs after each L2 extract and keeps the scene set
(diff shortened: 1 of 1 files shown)


Maxwell-Code07:

PR #1320, opened 2026-09-10
fix(memory-core): compute L1 dedup timestamps in code instead of trusting the model

### LINKED ISSUE #1318: [Bug] L1 dedup omits new-memory source timestamps and accepts unvalidated `merged_timestamps`
The L1 conflict prompts require `merged_timestamps` to contain the sorted, deduplicated union of timestamps from the new memory and selected existing memories. The current data flow omits the new memory’s source timestamps and does not enforce that union before writing.

Reporter-verified against the default branch `feat/server_team` at commit `c387ea4534d08f3204d50a137ef55206d7c50301`; the subsequent independent review did not reproduce this baseline check.

## Observed code behavior

- `ConversationMessage.timestamp` contains epoch milliseconds.
- `ExtractedMemory` has `source_message_ids` but no timestamp field; flattening in `l1-extractor.ts` does not resolve source timestamps.
- `formatBatchConflictPrompt()` includes candidate `timestamps` but omits timestamps from each new-memory object.
- `parseBatchResult()` accepts a model-provided `merged_timestamps` array after string coercion, without checking provenance or completeness.
- For `merge`/`update`, `writeMemory()` uses that array, or processing time when it is absent.

## Minimal reproduction

Evaluate this TypeScript fixture from `MemoryCore`; it requires no LLM or persistent store:

```ts
import assert from "node:assert/strict";
import {
  formatBatchConflictPrompt,
  type CandidateMatch,
} from "./src/core/prompts/l1-dedup.js";

const oldTime = "2026-09-08T04:00:00.000Z";
const newTime = "2026-09-09T04:00:00.000Z";
const source = {
  id: "msg-1",
  timestamp: Date.parse(newTime),
};

const newMemory = {
  record_id: "new-1",
  content: "Use UTC for reports.",
  type: "instruction" as const,
  priority: 70,
  source_message_ids: [source.id],
  metadata: {},
  scene_name: "Reporting",
};

const matches: CandidateMatch[] = [{
  newMemory,
  candidates: [{
    ...newMemory,
    id: "old-1",
    content: "Use local time for reports.",
    source_message_ids: [],
    timestamps: [oldTime],
    createdAt: oldTime,
    updatedAt: oldTime,
    sessionKey: "example",
    sessionId: "example",
  }],
}];

const prompt = formatBatchConflictPrompt(matches);

// Current behavior:
assert.equal(prompt.includes(oldTime), true);
assert.equal(prompt.includes(newTime), false);
```

The fixture mirrors the extracted-memory shape: `source_message_ids` identifies the source, but its timestamp is not carried into conflict detection.

Separately, source inspection shows that an `update` decision containing only the old timestamp—or an invented timestamp such as `2099-01-01T00:00:00.000Z`—can pass through parsing and become the written record’s `timestamps`.

## Expected behavior and impact

Conflict detection should receive the new memory’s source timestamps. Merge/update should preserve available timestamps from that memory and the selected targets without accepting unsupported model-generated additions.

Currently, the timestamp trail can omit new evidence or contain invented values. This is a provenance-integrity risk; no production frequency or specific downstream ranking failure is claimed. Source-message time is also distinct from event time and from `createdAt`/`updatedAt`.

## Possible implementation

One approach is to derive prepared-memory `timestamps` from validated `source_message_ids`, include them in `formatBatchConflictPrompt()`, and compute the final union in code after validating `target_ids`. The model would still choose the dedup action and merged content.

An experimental implementation exposed policy questions that should be resolved before adoption:

- Falling back to every scene message can attribute unrelated timestamps to a memory. Missing and partially resolved sources need an explicit policy.
- Throwing during extraction flattening aborts the extraction call. Prompt/decision validation errors instead use the existing batch-wide `store` fallback.
- `store` writes currently use processing time, including conservative fallback.
- Candidate recall reconstructs timestamps from `timestamp_str`; unioning those inputs cannot recover historical timestamps that recall does not expose.
- Target validation should explicitly define whether the allowed set is the per-memory candidate list or the unified batch pool.

These are implementation considerations, not prerequisites for recognizing the prompt/parser defect.

## Validation

A reporter-verified local experimental patch produced:

- targeted regression tests: **6/6 passed**;
- local full suite: **76/76 passed**;
- local full build: **passed**;
- `git apply --check` against `c387ea4534d08f3204d50a137ef55206d7c50301`: **passed**.

These execution and applicability results were not independently rerun during the subsequent read-only review.

The six added tests cover prompt inclusion/rejection, source conversion/fallback/rejection, and one selected-candidate timestamp union. Further coverage is needed for batch fallback, persistence, partial source resolution, and repeated merges.

## Related scope

Issue #1025 proposes proactive L1 maintenance and identifies the separate loss of `source_message_ids` during SQLite persistence. This report isolates the current write-time timestamp omission and unchecked timestamp union. The topics overlap in provenance, but #1025 does not describe this dedup prompt/parser defect.

Historical repair, periodic maintenance redesign, and general retry/idempotency behavior are outside this report’s scope.

LovePlayCode:
## Description | 描述
L1 conflict detection omitted new-memory source timestamps and wrote model-supplied `merged_timestamps`. Merge/update now unions source-message times with selected candidate timestamps in code, and ignores invented model values.

Unsafe merge/update decisions fall back to `store` (original extracted content, no target deletes): missing source timestamps, empty targets, all-invalid IDs, and mixed valid/invalid IDs. `writeMemory` applies the same guard so a raw decision cannot delete existing records without timestamps or targets.

## Related Issue | 关联 Issue
Fixes #1318

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明
- `cd MemoryCore && npx vitest run src/core/record/l1-dedup-timestamps.test.ts` → 34 passed
- Missing `source_message_ids` and out-of-range epoch values are skipped (no throw, no fallback to every scene message)
- Invalid / mixed / empty `target_ids` on merge/update fall back to `store`; blank ids are ignored, not treated as mixed
- Search hits contribute `timestamp_str` / `timestamp_start` / `timestamp_end` to the union
- Merge/update never stamps processing time; `writeMemory` also coerces empty-timestamp or empty-target replace actions to `store`

### DIFF
--- a/MemoryCore/src/core/prompts/l1-dedup.ts
+++ b/MemoryCore/src/core/prompts/l1-dedup.ts
@@ -43,8 +43,8 @@ export const CONFLICT_DETECTION_SYSTEM_PROMPT = `你是记忆冲突检测器。�
    - 跨类型示例：一条 episodic "用户在 2018 年开始做播客" + 一条 persona "用户有播客制作经验" → 可 merge 为一条 persona 或 episodic（取决于信息侧重）
 
 5. **timestamp 处理**：
-   - merge / update 时，merged_timestamps 应包含**所有相关记忆的时间戳并集**（去重排序）
-   - 这样可以保留事件发生的完整时间线
+   - 每条新记忆和候选记忆都已带 \`timestamps\`（来源消息时间，ISO 8601）。
+   - merge / update 时系统会用「新记忆 timestamps + 你选中的 target_ids 对应候选 timestamps」做并集，**不要编造时间戳**。\`merged_timestamps\` 可省略，填写也会被忽略。
 
 ## 输出格式
 
@@ -57,17 +57,15 @@ export const CONFLICT_DETECTION_SYSTEM_PROMPT = `你是记忆冲突检测器。�
     "target_ids": ["要删除的候选记忆 record_id 1", "record_id 2"],
     "merged_content": "合并/更新后的记忆内容（merge/update 时必填）",
     "merged_type": "合并后的最佳 type：persona|episodic|instruction|work_fact|work_task|work_method|work_artifact（merge/update 时必填）",
-    "merged_priority": 85,
-    "merged_timestamps": ["合并后的时间戳数组，包含所有新旧记忆时间戳的并集（merge/update 时必填）"]
+    "merged_priority": 85
   }
 ]
 
 字段说明：
-- target_ids：要删除替换的旧记忆 ID **数组**（可以 1 条或多条）。store/skip 时省略或为空。
+- target_ids：要删除替换的旧记忆 ID **数组**（可以 1 条或多条）。必须来自统一候选记忆池。store/skip 时省略或为空。
 - merged_content：merge/update 时的最终记忆文本。store/skip 时省略。
 - merged_type：merge/update 后记忆应归属的 type。根据合并后内容本质判断。
-- merged_priority：merge/update 后的新优先级（0-100 整数，merge/update 时必填）。合并后信息更完整、更确定，通常应**酌情提升** priority（例如两条 priority 70 的记忆合并后可提升到 80）。参考标准：80-100（核心特质/重要事件），60-79（一般偏好/普通活动），<60（次要信息）。
-- merged_timestamps：合并后的时间戳数组。收集新记忆 + 所有被合并旧记忆的时间戳，去重排序。`;
+- merged_priority：merge/update 后的新优先级（0-100 整数，merge/update 时必填）。合并后信息更完整、更确定，通常应**酌情提升** priority（例如两条 priority 70 的记忆合并后可提升到 80）。参考标准：80-100（核心特质/重要事件），60-79（一般偏好/普通活动），<60（次要信息）。`;
 
 export const WORK_CONFLICT_DETECTION_SYSTEM_PROMPT = `你是团队工作记忆冲突检测器。批量比较多条【新记忆】与【统一候选记忆池】中的已有记忆，逐条决定如何处理。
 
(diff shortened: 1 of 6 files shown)


Maxwell-Code07:

K1nG11 (commented):
Thanks for the fix. I reproduced three edge cases worth addressing:

1. **Missing source timestamps still allow merge/update.** With new `timestamps: []` and candidate timestamps `[t1]`, the merge keeps only `[t1]`. If both are empty, the writer uses processing time. To preserve source provenance, please fall back for that decision to storing the original memory or retrying, rather than replacing existing records.

2. **Invalid targets are dropped, but merged content is kept.** An unknown-only target list becomes `[]`, yet the writer still uses `merged_content`. Mixed valid/invalid targets also change the target set without revisiting the content. Please reject/re-judge the affected decision or fall back to `store` with the original content. The unified candidate pool itself is fine.

3. **An out-of-range timestamp can abort extraction.** This throws `RangeError: Invalid time value`:
   ```ts
   resolveSourceTimestamps(["m"], [{ id: "m", timestamp: 1e20 }]);
   ```
   `Number.isFinite` is not enough to guarantee a valid Date. Please guard `toISOString()` and skip invalid values; this call is outside the dedup catch.

Please add regression tests for missing/all-empty provenance, all-invalid/mixed targets (including resulting action/content), and mixed valid/out-of-range timestamps.

Checked at `75702936` using isolated Node helper repros and static writer-path inspection; full suite/E2E storage were not run.

LovePlayCode:
Thanks for reproducing these cases. Addressed in e334e86:

1. **Missing / all-empty source timestamps.** Merge and update now fall back to `store` with the original extracted memory when the new record has no source timestamps. Existing records are not deleted, and we no longer stamp processing time onto a replacement.

2. **Invalid / mixed targets.** If every target is outside the unified candidate pool, the target list is empty, or valid and invalid IDs are mixed, the decision also falls back to `store` with original content (`merged_content` is dropped). Blank target IDs are ignored rather than treated as mixed. `writeMemory` has a last-line guard for empty timestamps / empty targets when it is called with a raw decision.

3. **Out-of-range timestamps.** `toISOString()` is guarded (`epochMsToIso`). Finite but out-of-range values such as `1e20` / `-1e20` are skipped instead of throwing. The same helper is used when formatting extraction prompts, so a bad L0 timestamp cannot abort the batch.

Regression coverage is in `l1-dedup-timestamps.test.ts` (36 passing), including empty provenance, all-invalid / mixed / empty targets (action + content), mixed valid/out-of-range timestamps, and the writer/prompt guards.

PR #1332, opened 2026-09-10
fix(proxy): preserve WorkBuddy memory routing

sogakeji:
## Description | 描述

WorkBuddy renders an `AskUserQuestion` result as a single line containing the question, its skip hint, and the selected option. The CodeBuddy extractor previously scanned the whole line, so the hint's `跳过 / 不关联` text could win over an affirmative Team or Agent selection. The session was then persisted as bypassed, which disabled memory injection and L0 write-back.

This change:

- extracts the selected option after WorkBuddy's spaced `→` delimiter before applying bypass detection;
- preserves explicit skip selections and the `更多 →` pagination option;
- registers the selected Agent when a Team has no Task, because `task_id` is an optional memory isolation dimension;
- adds regression coverage for asset, Team, Agent and Task answers, explicit bypass, and pagination.

## Related Issue | 关联 Issue

None.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

Validation:

- `cd MemoryProxy && npm test` — 6 tests passed.
- End-to-end WorkBuddy form flow — selected Agent initialized, memory context available, and L0 messages read back successfully.
- `npm run typecheck` still reports errors already present on the base branch, including `SessionInitResult.resetFlow`, SessionInfo casts, the `skipAssetConfirm` config type, and the optional `@context-proxy/cost-guard` module. No new error category originates from this patch.

## Additional Notes | 其他说明

The parser only strips a delimiter when `→` is surrounded by the WorkBuddy display spacing and has a non-empty answer after it. This keeps the trailing arrow in the `更多 →` option intact.

### DIFF
--- a/MemoryProxy/src/session/codebuddy/__tests__/workbuddy-extractor.test.ts
+++ b/MemoryProxy/src/session/codebuddy/__tests__/workbuddy-extractor.test.ts
@@ -1,0 +1,60 @@
+import { describe, expect, it } from "vitest";
+import type { TeamOption } from "../../types.js";
+import {
+  BYPASS_MARKER,
+  extractAgentOnly,
+  extractAssetConfirm,
+  extractTaskOnly,
+  extractTeamFromOptionText,
+} from "../extractor.js";
+
+const team: TeamOption = {
+  team_id: "team-12345678",
+  team_name: "home",
+  agents: [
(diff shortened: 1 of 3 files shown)


Maxwell-Code07:

PR #1333, opened 2026-09-10
fix(memory-core): stop compact truncation loops spinning on oversized tool results

### LINKED ISSUE #832: [Bug] Infinite loop in tool-result truncation during compact

LovePlayCode:
## Description | 描述
Compact used to slice a large `tool_result` to 2000 characters and then append a truncation notice. The combined string stayed above the 2000-character skip threshold, so `while (tokensToFree > 0)` re-selected the same message forever and pinned CPU.

The notice is now counted inside the character budget, already-truncated messages are skipped, and a regression test covers the issue #832 fixture.

## Related Issue | 关联 Issue
Fixes #832

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

```
cd MemoryCore
npx vitest run src/offload_server/compact/compressor-truncate.test.ts
# Test Files  1 passed (1)
# Tests  4 passed (4)
```

## Additional Notes | 其他说明
`aggressiveCompress` / `emergencyCompress` used the same slice-then-append pattern; they now share `truncateToolResultText` so the length gate keeps working there too.

### DIFF
--- a/MemoryCore/src/offload_server/compact/compressor-truncate.test.ts
+++ b/MemoryCore/src/offload_server/compact/compressor-truncate.test.ts
@@ -1,0 +1,77 @@
+import { describe, expect, it } from "vitest";
+import {
+  TOOL_RESULT_TRUNCATE_CHARS,
+  truncateTailToolResults,
+  truncateToolResultText,
+} from "./compressor.js";
+import type { Message } from "./helpers.js";
+
+function toolResult(content: string): Message {
+  return { role: "tool", content };
+}
+
+function countingEstimator(maxCalls: number) {
+  let calls = 0;
+  return (text: string) => {
+    if (++calls > maxCalls) {
+      throw new Error(`estimateTokens called ${calls} times — truncation is not converging`);
+    }
+    return Math.max(1, Math.ceil(text.length / 4));
+  };
+}
+
+describe("truncateToolResultText", () => {
+  it("keeps short content unchanged", () => {
+    expect(truncateToolResultText("hello", 2000)).toBe("hello");
+  });
+
+  it("fits notice-inclusive output within truncateChars", () => {
+    const content = "x".repeat(TOOL_RESULT_TRUNCATE_CHARS + 500);
+    const truncated = truncateToolResultText(content, TOOL_RESULT_TRUNCATE_CHARS);
+    expect(truncated.length).toBeLessThanOrEqual(TOOL_RESULT_TRUNCATE_CHARS);
+    expect(truncated).toContain("content truncated");
+  });
+});
+
+describe("truncateTailToolResults", () => {
+  it("terminates when a single tool_result cannot free the requested budget", () => {
+    const truncateChars = TOOL_RESULT_TRUNCATE_CHARS;
+    const messages = [toolResult("x".repeat(truncateChars + 500))];
+    const tokenArray = [Math.ceil((truncateChars + 500) / 4)];
+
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1335, opened 2026-09-11
fix(sdk-ts): validate v3 response envelopes before accessing fields

Bryce-border:
[PR-description.md](https://github.com/user-attachments/files/32092866/PR-description.md)
# fix(sdk-ts): validate v3 response envelopes before accessing fields

When a v3 endpoint or intermediary returns JSON `null`, the TypeScript SDK
throws `TypeError: Cannot read properties of null (reading 'code')` instead
of `TDAMError`. This loses the HTTP status and request ID, including when
the response is an HTTP error. Other non-object JSON responses produce
misleading errors such as `[200] HTTP 200`.

Validate that the parsed envelope is a non-null, non-array object before
accessing its fields. Invalid envelopes now raise `TDAMError` with a clear
message, the request ID from response headers, and either the failing HTTP
status or `-1` for successful HTTP responses with an invalid envelope.

The change is limited to the shared v3 GET/POST transport. Valid success
envelopes, trace propagation, and business error details retain their
existing behavior.

## Validation

From `sdk/memory-core/typescript`:

```sh
npm test -- --reporter=verbose
npm run build
```

- Added 18 tests using mocked fetch responses; no live service is required.
- Without the fix: 12 tests fail and 6 pass.
- With the fix: all 18 tests pass.
- TypeScript build passes.
- `git diff --check` passes.

Verified on Windows with Node.js 20.17.0 and npm 10.8.2. This SDK declares
Node.js >=18 support. Full-stack Docker integration and the repository's
GitHub Actions workflow were not run.

### DIFF
--- a/sdk/memory-core/typescript/src/v3/http.ts
+++ b/sdk/memory-core/typescript/src/v3/http.ts
@@ -94,6 +94,14 @@ export class V3HttpTransport {
         );
       }
 
+      if (!envelope || typeof envelope !== "object" || Array.isArray(envelope)) {
+        throw new TDAMError(
+          response.ok ? -1 : response.status,
+          "API response must be a JSON object",
+          headerRequestId,
+        );
+      }
+
       const businessCode = typeof envelope.code === "number" ? envelope.code : undefined;
       if (!response.ok || businessCode !== 0) {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1344, opened 2026-09-11
fix(store-pool): forward sendDimensions to the sqlite embedding service

### LINKED ISSUE #1343: [Bug] store-pool.ts drops sendDimensions when building the embedding service — all embeddings fail silently and retrieval degrades to keyword-only
`store-pool.ts` builds the embedding service without passing `sendDimensions`, so the option silently defaults to `true` on the v3 multi-tenant path. Against any embedding endpoint that rejects an explicit `dimensions` parameter, **every** embed call then fails — and because all the call sites treat embedding failure as non-fatal, the deployment silently degrades to keyword-only retrieval with no error surfaced anywhere.

**Version:** `agentmemory/memory-core:latest` (image running 2026-09-11), `storeBackend: sqlite`, v3 API.

## The omission

`MemoryCore/src/core/store/store-pool.ts`, `createSqliteStore()`:

```ts
embeddingService = createEmbeddingService({
  provider: embCfg.provider,
  baseUrl: embCfg.baseUrl,
  apiKey: embCfg.apiKey,
  model: embCfg.model,
  dimensions: embCfg.dimensions,
  maxInputChars: embCfg.maxInputChars,
}, this.logger as StoreLogger);
```

No `sendDimensions`. `OpenAIEmbeddingService` then falls back to sending it:

```ts
// src/core/store/embedding.ts
this.sendDimensions = config.sendDimensions ?? true;
...
if (this.sendDimensions) {
  body.dimensions = this.dims;
}
```

The sibling construction site does pass it — `src/core/store/factory.ts`:

```ts
embeddingService = createEmbeddingService({
  ...
  sendDimensions: config.embedding.sendDimensions,
```

So the option works on the factory path and is dropped on the store-pool path. `sendDimensions` is a first-class config field (`src/config.ts:122`, parsed at `src/config.ts:610`, and again in `src/gateway/config.ts:655`), so from the operator's side it looks configured and supported.

## Symptom

Gateway config contains `embedding.sendDimensions: false` (verified identical on disk and inside the container). Every embed still fails:

```
Embedding API error: HTTP 400 Bad Request — {"error":{"message":
"Model \"Qwen-embedd\" does not support matryoshka representation,
 changing output dimensions will lead to poor results.","type":"BadRequestError"}}
```

Confirmed against the endpoint directly — the parameter is the only difference:

| Request | Result |
|---|---|
| `{"model":..., "input":"測試"}` | `200`, returns a 4096-dim vector |
| `{"model":..., "input":"測試", "dimensions":4096}` | `400`, message above |

## Why it is silent, which is the real problem

Every consumer treats the failure as non-fatal, so the service stays "healthy" while storing nothing:

```
[l1-writer] [vec-dual-write] Embedding FAILED for id=..., will write metadata only: ...
[sqlite] [L1-upsert] START id=... (no embedding — metadata-only write)
[sqlite] [L1-upsert] Skipping vec write (no embedding) id=...
[sqlite] [L0-upsert] Skipping vec write (no embedding) id=...
[l1-dedup] embedBatch failed (non-fatal, FTS may still run): ...
[tdai_memory_search] [hybrid-vec] Embedding search failed (non-fatal): ...
```

Meanwhile the health surface still reports everything as available:

```json
"stores": { "vectorStore": true, "embeddingService": true }
```

Those flags mean "configured", not "working", so neither `/health` nor a capabilities probe reveals the outage.

## Measured retrieval impact

Planted a distinctive fact, let the L0→L1 pipeline distil it, then queried two ways:

- **Lexical overlap present** — atom returned rank 1, score 0.950.
- **Semantic paraphrase with no shared terms** (deliberately different vocabulary and register for the same concept) — the atom is **absent from the top 30**; all 30 hits are unrelated.

That is BM25 doing all the work. The RRF hybrid has one arm permanently dead, and the only way to find out is to read container logs at DEBUG.

## Suggested fix

One line, matching what `factory.ts` already does:

```ts
  dimensions: embCfg.dimensions,
+ sendDimensions: embCfg.sendDimensions,
  maxInputChars: embCfg.maxInputChars,
```

Worth considering alongside it:

1. **Audit the two construction sites for other dropped fields.** `sendDimensions` is the one that bites here, but the same call is also missing `timeoutMs`/`proxyUrl`-class options that `factory.ts` forwards; any of them could diverge the same way.
2. **Surface embedding health.** A deployment whose vector arm is 100% failing should not report `embeddingService: true`. Even a counter of consecutive embed failures on `/health` would turn this from "discovered by accident weeks later" into something monitorable.
3. **Note that `dimensions` cannot be omitted as a workaround** — `OpenAIEmbeddingService` throws when it is missing or `<= 0`, so operators hitting this have no config-only escape and must patch the image.

## Backfill consequence

Anyone who has been running in this state has atoms that were written metadata-only. Fixing the code repairs new writes but leaves existing rows unvectorised, so a re-embedding path (or at least a documented procedure) would help affected deployments recover semantic search over their history.

Environment: macOS 15 (arm64), Docker, official prebuilt `agentmemory/memory-core:latest`, `storeBackend: sqlite` with sqlite-vec, remote OpenAI-compatible embedding endpoint (4096-dim, no matryoshka support).

L4XB:
## Description | 描述

`StorePool.createSqliteStore()` in `MemoryCore/src/core/store/store-pool.ts` built the embedding service without `sendDimensions`, so `OpenAIEmbeddingService` fell back to `true` on the store-pool (v3 multi-tenant) path even when the config disabled it. Against an endpoint that rejects an explicit `dimensions` parameter every embed call failed, and because embedding failures are non-fatal the deployment silently degraded to keyword-only retrieval. `factory.ts` already forwards the option; the pool now passes `sendDimensions: embCfg.sendDimensions` as well.

`MemoryCore/src/core/store/store-pool.test.ts` (new) builds a `StorePool` in sqlite mode with the embedding service, BM25 encoder, SQLite store and Kafka producer mocked, calls `getStore()` and asserts that `createEmbeddingService` receives `sendDimensions: false` from the config (and `true` when configured so). The first case fails on the current branch because the key is missing.

## Related Issue | 关联 Issue

Fix #1343

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过 (`vitest run src/core/store/store-pool.test.ts`: 2 passed)
- [x] No existing features affected | 无影响现有功能 (one added key on the sqlite path; tcvdb and mongodb paths untouched)

## Additional Notes | 其他说明

The repository had no vitest files under `MemoryCore/src` yet; the new test follows the `include` pattern in `vitest.config.ts`.

### DIFF
--- a/MemoryCore/src/core/store/store-pool.test.ts
+++ b/MemoryCore/src/core/store/store-pool.test.ts
@@ -1,0 +1,79 @@
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import type { MemoryTdaiConfig } from "../../config.js";
+
+const mocks = vi.hoisted(() => ({
+  createEmbeddingService: vi.fn(() => ({})),
+}));
+
+vi.mock("./embedding.js", () => ({
+  createEmbeddingService: mocks.createEmbeddingService,
+  NoopEmbeddingService: class {},
+}));
+vi.mock("./bm25-local.js", () => ({ createBM25Encoder: () => undefined }));
+vi.mock("./sqlite/memory-store.js", () => ({
+  VectorStore: class {
+    async close(): Promise<void> {}
+  },
+}));
+vi.mock("../report/kafka-metric-producer.js", () => ({
+  metricProducer: { initialize: async () => undefined },
+}));
+
+import { StorePool } from "./store-pool.js";
+
+const logger = { info() {}, warn() {}, error() {}, debug() {} };
+
+function memoryCfg(sendDimensions: boolean): MemoryTdaiConfig {
+  return {
+    bm25: { enabled: false, language: "en" },
+    embedding: {
+      enabled: true,
+      provider: "openai",
+      baseUrl: "https://embeddings.example/v1",
+      apiKey: "test-key",
+      model: "text-embedding-3-small",
+      dimensions: 1536,
+      sendDimensions,
+      maxInputChars: 8000,
+    },
+  } as unknown as MemoryTdaiConfig;
+}
+
+describe("StorePool sqlite embedding service", () => {
+  const dirs: string[] = [];
+
+  afterEach(() => {
+    mocks.createEmbeddingService.mockClear();
+    for (const dir of dirs.splice(0)) rmSync(dir, { recursive: true, force: true });
+  });
+
+  it("forwards sendDimensions from the embedding config (#1343)", async () => {
+    const dataDir = mkdtempSync(path.join(tmpdir(), "store-pool-"));
+    dirs.push(dataDir);
+    const pool = new StorePool({ mode: "sqlite", memoryCfg: memoryCfg(false), dataDir, logger });
+
+    await pool.getStore("tenant-a", null);
+
+    expect(mocks.createEmbeddingService).toHaveBeenCalledTimes(1);
+    expect(mocks.createEmbeddingService.mock.calls[0]?.[0]).toMatchObject({
+      provider: "openai",
+      model: "text-embedding-3-small",
+      dimensions: 1536,
+      sendDimensions: false,
+    });
+  });
+
+  it("keeps sendDimensions enabled when the config says so", async () => {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1348, opened 2026-09-11
fix(memory-core): preserve source fields in conversation search

Bryce-border:
# fix(memory-core): preserve source fields in conversation search

FTS conversation search drops `session_id`, `user_id`, and `agent_id` while
mapping L0 store results to `ConversationSearchResultItem`. The store returns
these fields and the result interface requires them, but callers lose the
source metadata when searching historical conversations.

This also affects client-side hybrid search: RRF visits FTS hits first and
retains that copy when the vector path returns the same record. Even a complete
vector hit therefore loses its source fields after merging.

Copy the three fields in the FTS result mapping, matching the existing vector
and native-hybrid mappings. No public API, ranking, or filtering changes are
required.

## Validation

- Added five tests covering FTS, vector, client-side hybrid, native hybrid,
  and fallback to FTS after an embedding failure.
- Before the fix: three failures (FTS, hybrid, fallback) and two passes.
- After the fix: all five tests pass, including RRF deduplication and scores.
- Plugin bundle builds successfully with `tsdown`.
- Skill queue isolation guard and `git diff --check` pass.

Commands from `MemoryCore` with Node.js 24.19.0:

```sh
npm test -- --reporter=verbose
npm run build:plugin
BASE_REF=origin/feat/server_team bash scripts/ci/check-skill-queue-isolation.sh
git diff --check
```

The tests mock store and embedding responses and exercise the real conversation
search implementation. The checked-out module's test configuration discovers
only this new test file. Live database integration and migration-script builds
were not run.

### DIFF
--- a/MemoryCore/src/core/tools/conversation-search.test.ts
+++ b/MemoryCore/src/core/tools/conversation-search.test.ts
@@ -1,0 +1,98 @@
+import { describe, expect, it, vi } from "vitest";
+import type { EmbeddingService } from "../store/embedding.js";
+import type { IMemoryStore, L0SearchResult } from "../store/types.js";
+import { executeConversationSearch } from "./conversation-search.js";
+
+function conversation(id: string): L0SearchResult {
+  return {
+    record_id: id,
+    session_key: `key-${id}`,
+    session_id: `session-${id}`,
+    team_id: "team-1",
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1349, opened 2026-09-11
fix(proxy): send a configurable bearer on the auth verify call

### LINKED ISSUE #1341: [Bug] MemoryProxy cannot authenticate against a gateway-protected MemoryCore: verifyUserKey() sends no Authorization header and auth config has no field for one
`MemoryProxy` cannot authenticate against a MemoryCore that has gateway auth enabled: `verifyUserKey()` calls `/v3/meta/auth/verify` without an `Authorization` header, and the `auth` config section has no field to supply one. Every request through the proxy then fails with `authentication_error`.

**Version:** `906b582` (branch `feat/server_team`, 2026-09-11)

## The client never sends a bearer

`MemoryProxy/src/auth.ts`, `verifyUserKey()`:

```ts
const fetchOpts: RequestInit = {
  method: "POST",
  headers: {
    "content-type": "application/json",
    "x-tdai-service-id": serviceId,
  },
  body: JSON.stringify({ user_key: userKey }),
};
...
const url = config.url.replace(/\/+$/, "") + "/v3/meta/auth/verify";
```

Only `content-type` and `x-tdai-service-id`. There is also no config field for a token — the `auth` section in `config.example.yaml` is just:

```yaml
auth:
  enabled: true
  url: "http://kernel.example.com:8420"
  timeoutMs: 5000
```

and the `AuthConfig` type in `src/types.ts` has no token/apiKey/bearer member. Note that `tdai.apiKey` and `skill.serviceToken` **do** exist for the other two kernel clients, so this looks like an omission in one client rather than a deliberate design.

## Reproduction

1. Run MemoryCore with gateway auth enabled (a non-empty `server.apiKey` / `TDAI_GATEWAY_API_KEY`).
2. Point MemoryProxy at it with `auth.enabled: true` and `auth.url` set to that gateway.
3. Send any request through the documented Claude Code path:

```
POST http://127.0.0.1:<port>/claude-code/default/v1/messages
Authorization: Bearer sk-mem-...      # a valid business user key
```

Result:

```json
{"type":"error","error":{"type":"authentication_error",
 "message":"Authentication failed: auth service returned HTTP 401"}}
```

Calling the same endpoint directly isolates it to the missing header:

| Request to `/v3/meta/auth/verify` | Result |
|---|---|
| `x-tdai-service-id` + `x-tdai-user-key` header, no bearer (≈ what the proxy sends) | `401` |
| bearer + `x-tdai-user-key` **header** | `400 user_key: Invalid input: expected string, received undefined` |
| bearer + `user_key` in **body** | `200 {"code":0,"data":{"valid":true,...}}` |

The proxy already gets the body shape right (row 3's body); it is only missing the bearer.

## Impact

Any deployment that turns gateway auth on cannot use MemoryProxy's own auth layer. That is not a corner case — MemoryCore's `/v3` surface has no user-level authorization of its own, so binding it to loopback **and** setting a gateway key is the natural hardening step, and it is the one that breaks the proxy.

There is already a trace of this in the repo: `deploy`'s MemoryCore start script leaves the gateway key empty by default. That default makes the proxy work, but it also means the documented happy path ships with the kernel's only transport-level auth disabled.

The two workarounds are both bad:

- **Turn off the gateway key** — makes the proxy work by removing the kernel's transport auth.
- **Set `auth.enabled: false`** — what I ended up doing. It is survivable because the client's `sk-mem` key is still forwarded as `x-tdai-user-key` to the kernel (`anthropicHandler.ts:686`), so identity still flows and MemoryCore remains the real gatekeeper; the proxy was only doing a duplicate check. But a deployment should not have to disable an auth feature to use an auth-protected backend.

## Suggested fix

Add an optional token to `AuthConfig` and send it, mirroring what `tdai.apiKey` / `skill.serviceToken` already do for the sibling clients:

```ts
headers: {
  "content-type": "application/json",
  "x-tdai-service-id": serviceId,
  ...(config.apiKey ? { authorization: `Bearer ${config.apiKey}` } : {}),
},
```

with a matching `auth.apiKey` in `config.example.yaml`. Behaviour is unchanged when it is empty, so this is backwards compatible.

Environment: macOS 15 (arm64), Node 22.22.3 (MemoryProxy enforces v22.x), MemoryCore v3 on loopback with `storeBackend: sqlite`.

L4XB:
## Description | 描述

`MemoryProxy/src/auth.ts` `verifyUserKey()` called `POST /v3/meta/auth/verify` with only `content-type` and `x-tdai-service-id`, and `AuthConfig` had no field for a token. Against a MemoryCore whose gateway is protected (`server.apiKey` / `TDAI_GATEWAY_API_KEY`) every verify call answered 401, so every proxied request failed with `authentication_error`, exactly as measured in #1341 (bearer + `user_key` in the body is what the endpoint accepts; the proxy already had the body right and was only missing the header).

This adds `auth.apiKey`:

- `AuthConfig.apiKey` (`src/types.ts`, plus the raw yaml type) with `""` as the default in `src/config.ts`, so the existing `${VAR}` expansion can reference a secret.
- `verifyUserKey()` sends `authorization: Bearer <apiKey>` when it is set and is byte-for-byte unchanged when it is empty. `auth.init` logs whether a bearer is configured (never the value).
- `config.example.yaml` documents the new key next to `timeoutMs`, mirroring `tdai.apiKey` / `skill.serviceToken` for the sibling kernel clients.

Not covered here: the follow-up in the issue that `auth.enabled: false` silently disables the memory pipeline. That is a separate problem in how the empty `userId` is handled downstream and deserves its own change.

## Related Issue | 关联 Issue

Fix #1341

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

New `MemoryProxy/src/__tests__/auth.test.ts` (the first file under the vitest `src/__tests__` include) stubs `fetch` and asserts: the bearer and body shape sent to `/v3/meta/auth/verify`, the exact header set when `apiKey` is empty, and that a 401 from the gateway is still a rejection.

```
cd MemoryProxy && npx vitest run src/__tests__/auth.test.ts
Test Files  1 passed (1)
     Tests  3 passed (3)
npx tsc --noEmit   # 60 pre-existing errors on feat/server_team, unchanged by this PR
```

### DIFF
--- a/MemoryProxy/config.example.yaml
+++ b/MemoryProxy/config.example.yaml
@@ -376,6 +376,8 @@ auth:
   enabled: true            # 是否启用鉴权
   url: "http://kernel.example.com:8420"   # 内核鉴权服务 base URL（自动拼接 /v3/meta/auth/verify）
   timeoutMs: 5000          # 鉴权请求超时（毫秒）
+  apiKey: ""               # 可选：内核 gateway 开启 API Key（server.apiKey / TDAI_GATEWAY_API_KEY）时
+                           #   作为 Authorization: Bearer 发送；留空则不带该头（与 tdai.apiKey 含义相同）
 
 
 # ── Admin / Ops 端点鉴权（服务级 shared secret） ─────────────────────────────
--- a/MemoryProxy/src/__tests__/auth.test.ts
+++ b/MemoryProxy/src/__tests__/auth.test.ts
@@ -1,0 +1,72 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("../report/log.js", () => ({
+  log: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() },
+}));
+
+import { initAuth, verifyUserKey } from "../auth.js";
+
+const VALID = { code: 0, data: { valid: true, user: { user_id: "usr-1" } } };
+
+function jsonResponse(status: number, body: unknown): Response {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: { "content-type": "application/json" },
+  });
+}
+
+describe("verifyUserKey against a gateway-protected MemoryCore (#1341)", () => {
+  const fetchMock = vi.fn<typeof fetch>();
+
+  beforeEach(() => {
+    fetchMock.mockReset();
+    vi.stubGlobal("fetch", fetchMock);
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+    initAuth({ enabled: false, url: "", timeoutMs: 0, apiKey: "" });
+  });
+
+  it("sends the configured bearer on the verify call", async () => {
+    fetchMock.mockResolvedValue(jsonResponse(200, VALID));
+    initAuth({ enabled: true, url: "http://kernel:8420/", timeoutMs: 0, apiKey: "gw-secret" });
+
+    const result = await verifyUserKey("sk-mem-1", "default");
+
+    expect(result).toEqual({ userId: "usr-1", rejected: false });
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+    const [url, init] = fetchMock.mock.calls[0];
+    expect(url).toBe("http://kernel:8420/v3/meta/auth/verify");
+    expect(init?.headers).toEqual({
+      "content-type": "application/json",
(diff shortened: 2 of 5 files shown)


Maxwell-Code07:

PR #1354, opened 2026-09-12
feat(core): append-only memory events for L1 writes + session diff endpoint

1173591564:
## Description

Adds an append-only `memory_events` stream on the automatic L1 extraction write path, plus a session-level diff query endpoint that aggregates those events into a human-readable change set.

### Motivation

`writeMemory` currently lands dedup decisions (store / update / merge / skip) directly: the new record is appended to JSONL + upserted to the vector store, and superseded rows are deleted from retrieval. After the fact there is no record of *what changed during a session* — which memories were created, which existing ones were replaced, and what their previous content was. The existing `memory_audit` table covers explicit management APIs (atomic/update, atomic/delete, scenario/write, ...) and intentionally stores no content or session dimension, so it cannot answer this question.

This PR fills the gap with a second, complementary event stream:

| | `memory_audit` (existing) | `memory_events` (new) |
|---|---|---|
| Trigger | Explicit mutation APIs | `writeMemory` dedup landing (automatic extraction) |
| Content | None (by design) | Content snapshots — before/after |
| Session dimension | None | `session_key` / `session_id` + `origin_session_*` |
| Relationship | None | `supersedes` / `superseded_by` links |

### Event model

- `store` → one `created` event
- `update` / `merge` → one `superseded` event per target (old-content snapshot) + one `updated`/`merged` event carrying `supersedes: [...]`
- `skip` → no event (nothing was persisted)

`superseded` events record the **acting** session in `session_id`/`session_key` and the replaced record's original owner in `origin_session_id`/`origin_session_key`, supporting both "what did this session change" and "which of my memories got replaced".

### Query API

`POST /v2/memory/diff` and `POST /v3/memory/diff` (v3 enforces the strict team+agent+user isolation triple like the other data-plane routes).

Request: `{ session_id, session_key?, op?, since?, until?, limit?, offset? }`

Response aggregates raw events into changes:

```json
{ "changes": [
  { "op": "updated", "record_id": "m_b", "content": "salary 6000", "version": 1,
    "event_ts": "...", "replaced": [
      { "record_id": "m_a", "content": "salary 5000", "origin_session_id": "ses-x", ... }
    ] }
], "total": 1 }
```

Orphaned `superseded` rows (whose replacing event never landed, or landed on another page) surface as standalone changes so nothing is silently dropped.

### Notes

- Pure addition — write semantics (JSONL append + vector upsert/delete) unchanged; event writes are best-effort and never block the write path.
- `appendMemoryEvent` / `queryMemoryEvents` are optional `IMemoryStore` methods; backends that don't implement them get a clean `501`.
- The SQLite `queryL1Records` ignores `recordIds` (pre-existing behavior — it also feeds the `maxVersion` calculation, which this PR deliberately leaves untouched). The writer narrows the returned rows to `target_ids` itself before emitting `superseded` events.

## Related Issue

N/A — design groundwork for session-level memory auditing / diffs.

## Change Type

- [x] New feature (non-breaking change which adds functionality)

## Self-test Checklist

- [x] I have tested this change locally
- [x] `npm run build` (tsdown) passes
- [x] Added unit tests: `memory-events.test.ts` (SQLite round-trip/filters/pagination), `l1-writer-events.test.ts` (event emission per dedup action), `memory-diff.test.ts` (route aggregation + isolation)
- [x] No changes to existing write/retrieval semantics; event failures are non-fatal

## Additional Notes

Verified locally end-to-end (tsx runner, node:sqlite): created/updated/merged/superseded event shapes, skip producing no event, origin-session preservation, op filtering, cross-tenant isolation, v3 422 on missing triple, 400 on missing session_id, and the `/v2` twin route.


Maxwell-Code07:

PR #1355, opened 2026-09-12
fix(memory-core): publish L3 persona after successful generation

Artemedi:
## Summary

- run L3 file tools in an isolated draft workspace and publish `persona.md` only after the complete runner succeeds
- reject generation, profile/index reads, and missing changed-scene evidence instead of advancing the L3 checkpoint
- make filesystem publication atomic and harden sandbox path containment against same-prefix sibling traversal

## Verification

- `npm test` — 2 files, 9 tests passed
- deterministic local OpenAI-compatible AI-SDK smoke — first response wrote the draft, three continuation attempts returned HTTP 429, generation rejected, and the live persona remained byte-for-byte unchanged
- `git diff --check` — passed

## Known baseline issue

The aggregate `npm run build` reaches and passes the plugin, migration, export, and read-local-memory builds, then fails because the checked-out branch's `package.json` references the absent `scripts/seed-v2/tsconfig.json`. This PR does not touch that script.

### DIFF
--- a/MemoryCore/src/adapters/standalone/llm-runner.test.ts
+++ b/MemoryCore/src/adapters/standalone/llm-runner.test.ts
@@ -1,0 +1,15 @@
+import path from "node:path";
+import { describe, expect, it } from "vitest";
+import { resolveSandboxedPath } from "./llm-runner.js";
+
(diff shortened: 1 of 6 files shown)


Maxwell-Code07:

PR #1356, opened 2026-09-12
feat(panel): memory review — session diff viewer + revert endpoint

1173591564:
## Description

**Stacked on #1354** — this branch is based on `feat/l1-memory-events`; the diff below includes that PR's changes. Please review only the last commit (`feat(panel): memory review — session diff viewer + revert endpoint`).

Adds a review workflow on top of the `memory_events` stream: a Panel page that renders a session's memory changes like a pull request diff, plus a revert endpoint that undoes an applied change.

### MemoryCore

- `memory_events` gains `snapshot_json` on `superseded` rows — the full previous `L1RecordRow` serialized, required to faithfully restore a superseded record.
- New op `reverted`: reverting a change is itself a memory mutation, so it appends to the same append-only stream (audit trail stays complete).
- `POST /v2|v3/memory/diff/revert`:
  - `created` → `deleteL1(record_id)`
  - `updated` / `merged` → `deleteL1(record_id)` + `upsertL1` rebuild from each superseded `snapshot_json` (re-embeds via the request's EmbeddingService when available; metadata+FTS otherwise)
  - Idempotent: a second revert on the same record → `409`; unknown record → `404`; the revert event is attributed to the **original** session so the change's diff view shows its own rejection.
- `POST /memory/diff` marks each change `reverted: true` when a `reverted` event exists for its `record_id`.

### MemoryPanel

- Backend: `POST /api/v1/memory/diff` + `POST /api/v1/memory/diff/revert` — thin transparent proxies to the kernel via `kernelHttp.postEnvelope` (same pattern as the skill proxy routes).
- Frontend: new `MemoryReviewPage` (`/memory/review`) — team/agent/user/session picker, change cards showing op tag + content + before/after replaced snapshots (with origin session), Revert button per active change, `reverted` badge on reverted ones.
- Menu entry `memory_review` under the assets group; zh-CN + en-US strings.

### Revert semantics (post-hoc, not a gate)

Changes take effect immediately as today; review is a revert/undo operation, not a pending-approval gate. A future iteration could add a pending mode, but that changes "when memory becomes recallable" and belongs in a separate design.

## Related Issue

N/A — builds on #1354's event stream.

## Change Type

- [x] New feature (non-breaking change which adds functionality)

## Self-test Checklist

- [x] Kernel: revert flow verified end-to-end locally (tsx + node:sqlite): snapshot restore, `reverted` marker in diff, 409 idempotency, 404 unknown, cross-tenant revert blocked
- [x] Panel backend: `tsc --noEmit` clean
- [x] Panel web: `tsc --noEmit` clean
- [x] `npm run build` (tsdown) passes for MemoryCore

## Additional Notes

Pre-existing base behaviour found while testing: `deleteL1Batch`'s isolation filter carries `session_id`, so a cross-session `update` decision does not actually delete the old record (it's skipped by `rowMatchesIsolation`). The superseded event is still written, so the diff is accurate — but the old row stays recallable. Worth a separate bug-fix PR; intentionally not fixed here.


Maxwell-Code07:

PR #1216, opened 2026-08-30
fix(memory-core): update skill usage_count/last_used_at on read (#1189)

tung515:
Fixes #1189

## Description | 描述

`handleGet()` in `skill-handlers.ts` only reads the skill content — it never calls `MetadataService.touchAssetUsage()`. This is inconsistent with `handleCreate()`, which registers the asset via `ensureSkillAsset()`. As a result, `meta_assets.usage_count` and `last_used_at` stay at 0/null no matter how many times a skill is read.

**Fix:** call `metaSvc.touchAssetUsage(row.skill_id)` right after a successful read, wrapped in its own try/catch so a failure here never blocks the skill response — matching the "read-time self-heal path must not block the response" contract already documented on `ensureSkillAsset()`'s failure-strategy comment in `metadata-service.ts`.

## Related Issue | 关联 Issue

Fixes #1189

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [ ] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

## Additional Notes | 其他说明

**Tests added** (`MemoryCore/src/gateway/__tests__/skill-handlers.test.ts`):
- asserts `touchAssetUsage` is called once with the correct `skill_id` after a successful read;
- asserts a failure inside `touchAssetUsage` does not affect the response (still returns the skill with `code: 0`).

**Manual end-to-end verification:** created a skill via `/v3/skill/create`, called `/v3/skill/get` 5 times, confirmed `meta_assets.usage_count` went `0 -> 5` and `last_used_at` was populated in the SQLite metadata store.

PR #1117, opened 2026-08-21
fix(skill): expose extraction action outcomes

Gout999:
## Summary

Addresses #1009 with a privacy-preserving observability slice for conversation-add
skill extraction.

The worker now reports aggregate candidate outcomes (`created`, `updated`,
`non_create`, and action counts) in its logs and task completion metadata. The helper
does not emit skill content or skill names.

This PR does not change the extractor's decision policy or add a new task-result API;
it makes the existing outcomes inspectable first.

## Verification

- `npm test -- --run src/core/skill/conversation-add/candidate-outcome.test.ts` (2 tests passed)
- `git diff --check`

Signed-off-by: Gout999 <gout999@users.noreply.github.com>

PR #1358, opened 2026-09-12
fix(memory-core): expose extraction configuration in skill archive responses

### LINKED ISSUE #972: [Bug] skill_extract 工具被无条件注入，但调用恒返回 40003「触发路径已下线」
## 环境
- **部署模式**：server team mode（`deploy/global-images` 三件套）
- **版本**：`@tencentdb-agent-memory/memory-tencentdb-v2` @ commit `4dca55c`
- **OS**：macOS 14

## 问题描述
`skill_extract` 被无条件放入 read-only 工具集注入给 LLM，但代理层对 `extract` 子路径无条件返回 `40003 触发路径已下线`。LLM 看到一个可用工具，按其描述调用后必然失败——浪费一轮工具往返，并造成工具声明与实际行为不符。

## 复现步骤
1. 按 `deploy/global-images` 部署三件套并接入 coding agent。
2. 观察注入的 readTools 列表，可见 `skill_extract`。
3. 让 LLM 调用 `skill_extract`（或直接打 `.../extract`）。
4. 返回 `code=40003`、HTTP 400「触发路径已下线」。

## 期望行为
extract 路径已下线时，不应再将 `skill_extract` 作为可用工具注入（否则诱导必然失败的调用）。

## 证据（commit `4dca55c`）
- 注入：`MemoryProxy/src/injection/injectors/skill-tools-injector.ts:101-105`（头注释 `:16-17` 也列了它）
- 拒绝：`MemoryProxy/src/skill/skill-bridge.ts:560-577`（`sub==='extract'` → `40003`）
- `ALLOWED_SUBPATHS` 仍含 `extract`：`skill-bridge.ts:143`

## 建议修复
从 readTools 移除 `skill_extract`，保留注释说明恢复条件（需先恢复 core 侧 extract 处理）。

---
*基于源码静态阅读，行号请按最新 commit 核实。*

VittorioCai:
## Description | 描述

`POST /v3/skill/extract`, `POST /v3/skill/conversation/add` and
`POST /v3/skill/conversation/force-archive` answer a success with `ok`/`status` and a
`task_id`. That answer is accurate: `SkillTriggerService.archive()` writes the archive,
appends the `SkillTaskEntry` and calls `enqueueAgent`, all three inside the tasks mutex.
It has never meant that a Skill was produced — and with `skill.extraction.enabled=false`
the accepted task cannot be executed in that configuration, while no field in the
response says so.

The caller therefore cannot distinguish "accepted, extraction under way" from "accepted,
extraction cannot run here": both are `code: 0` with a `task_id`. A client that waits for
a skill to appear waits forever, and a client polling `/v3/skill/list` cannot tell an
empty result caused by the switch from one caused by a session with nothing worth
keeping.

One of those callers is the product's own agent-facing tool. MemoryProxy injects
`skill_extract` into the read-only tool set of every session
(`MemoryProxy/src/injection/injectors/skill-tools-injector.ts`), described to the model as
"立即归档当前对话触发一次 skill 抽取（异步任务，由后台 agent 分析对话内容生成 skill）".
When the model calls it, the bridge forwards it to Core's
`/v3/skill/conversation/force-archive` (`skill-bridge.ts`, the `sub === "extract"` branch)
and returns Core's envelope to the model verbatim — the response body is passed through
unchanged except for team-search filtering. So with extraction off, the model is told it
triggered extraction, is handed a `task_id`, and has nothing in the answer to tell it
otherwise. With this field it does.

This adds one field, `extraction_enabled`, to the success responses of those three entry
points. It reports the configuration switch and nothing more:

- **not** worker readiness and **not** task completion — `true` only means the switch is on;
- `null`, never `false`, when the resolved skill config cannot be reached, so
  "cannot tell" stays distinguishable from "it is off".

The archive contract is unchanged: the slice is still archived, the task is still
registered and enqueued, and error responses keep their existing shape. Rejecting the
call instead would be worse — the archive and the task entry are already written by the
time the handler returns, so a failure response would invite a retry that duplicates
them.

## Related Issue | 关联 Issue

Related: #972 — `skill_extract` is injected unconditionally while the call cannot do what
its description promises. The 40003 "trigger path retired" half of that report no longer
applies (the bridge now forwards `extract` to `conversation/force-archive`), but the other
half does: the tool is still advertised to every session, and when extraction is off the
call now *succeeds* and still produces nothing. **This PR does not fix #972** — it does not
change what is injected into the tool set, which is what that issue asks for. It makes the
failure visible in the response instead of silent.

Also related but not overlapping: #1117 (see Additional Notes). No issue was filed for this
specific gap — it was found while running the product end to end and hitting it.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [x] Documentation update | 文档更新 — API doc and both SDKs declare the new field

## Self-test Checklist | 自测清单

  `npx vitest run src/gateway/skill-archive-extraction-flag.test.ts` — 6 tests, all passing:

  1. extraction off — the slice is still archived, `archive` is still called once,
     `ok`/`task_id`/`archive_key` unchanged, and `extraction_enabled: false`;
  2. extraction on — response keys are exactly the previous four plus the new one, and
     nothing in it speaks about the outcome of extraction;
  3. config unreachable — `null` in both cases: the optional dep absent, and the dep
     present but returning `undefined` (skill not yet constructed). Never a fabricated
     `false`;
  4. `conversation/add` — reported both below the threshold (`status: "ok"`) and at it
     (`status: "archived"`, `archived` preserved);
  5. `force-archive` — reported on both the `empty` and the `archived` answer;
  6. a failed archive keeps its original error (`50001`, original message) and the new
     field is **not** injected into the error envelope.

  Also reproduced live against a running Core — evidence under Additional Notes.

  The field is additive on success envelopes only; no request schema, no status code and
  no existing field changes. `bash scripts/ci/check-skill-queue-isolation.sh` with
  `BASE_REF=origin/feat/server_team` → PASS. `tsc` on the touched file reports the same
  18 pre-existing diagnostics before and after (none added). `npm pack --dry-run` does
  not include the new test file (`!src/**/*.test.ts` in `files`), so the published
  package is unchanged in shape. MemoryProxy and MemoryPanel pass the response through
  and ignore unknown fields, so they keep working untouched.

### Reproduction

Live, on standalone + sqlite, with `skill.extraction.enabled: false`:

```bash
curl -sS -H "authorization: Bearer $KEY" -H "x-tdai-user-key: $KEY" \
  -H 'content-type: application/json' -H 'x-tdai-service-id: default' \
  -X POST http://localhost:8420/v3/skill/extract -d '{
    "user_id":"usr-…","team_id":"team-…","agent_id":"agt-repro-extraction-flag",
    "session_id":"repro-1",
    "messages":[{"role":"user","content":"…"},{"role":"assistant","content":"…"}]}'
```

① the response — no mention of the switch:

```json
{"code":0,"message":"ok","request_id":"req-6f8365f3512d416d",
 "data":{"ok":true,"task_id":"skill-extract-task-fe84b9ed",
         "archived_at_ms":1789235870833,"archive_key":"skill_buffer/…/data-….jsonl"}}
```

② the task really is registered and enqueued —
`…/agt-repro-extraction-flag/_tasks.json` after the call:

```json
{"tasks":[{"task_id":"skill-extract-task-fe84b9ed","enqueued_at_ms":1789235870833,
           "archive_key":"skill_buffer/…/data-1789235870833.jsonl", "…":"…"}]}
```

③ the worker pool starts and then cannot execute it — `docker logs`:

```
INFO  [skill-worker-pool] start pool_id=skill-pool-default-7 concurrency=60 …
ERROR [skill-worker-pool] skill-pool-default-7#0 consumeAgent error:
      [skill-worker-pool] standalone SkillExtractor unavailable (instance=default)
ERROR [skill-worker-pool] skill-pool-default-7#1 consumeAgent error: … (repeats)
```

An empty candidate pool is deliberately **not** offered as evidence: a real extraction
may also produce no candidates, so on its own it proves nothing.

### Scope — stated explicitly

- **Not** a claim that the task is never queued. It is queued; ① – ③ above show the task
  entry and the worker attempts.
- **Not** a claim that every storage mode honours the switch. The extractor is built per
  storage mode — service mode builds one per instance
  (`buildSkillExtractorForInstance`), standalone uses the process singleton
  (`TdaiCore.getSkillExtractor`) — and this reproduction covers the standalone path only.
  The field reports the process-level resolved configuration, which is what
  `getResolvedSkillConfig()` returns.
- **Not** a change to the retry policy for tasks already in the queue. The repeated
  `consumeAgent error` above is left exactly as it is.
- **Not** a readiness probe. With `extraction.enabled=true` but no LLMRunner the resolver
  records a degradation and keeps `enabled: true`; this field still reports `true`, and
  the doc says so. Surfacing degradations would be a different, larger change.

### Relationship to #1117

[#1117](https://github.com/TencentCloud/TencentDB-Agent-Memory/pull/1117)
(`fix(skill): expose extraction action outcomes`) reports what extraction *did* —
aggregate candidate outcomes in the worker's logs and task-completion metadata — and
touches `candidate-outcome.ts` and `extract-worker.ts`. That is the worker side, after
extraction runs. This PR is the synchronous HTTP response, before anything runs, and
touches `skill-handlers.ts`. No file and no field overlaps. The two are complementary:
#1117 tells you what came of a task that executed, this tells you whether one can
execute at all.

A search of this repository's full pull-request history (not only recent PRs) for
`extraction_enabled`, `skill.extraction.enabled`, the three route paths, `force-archive`,
and `skill-handlers` / `extract-worker` / `trigger-service` found no other PR touching
these response fields. #1216 also edits `skill-handlers.ts`, but only `handleGet` — no
hunk overlap.

### Size

26 lines in the implementation (`skill-handlers.ts`: one helper plus four call sites),
125 lines of test, and 61 lines of API-doc and SDK-type declaration —
`v3-api-memorycore-doc.md`, the three TypeScript response interfaces, and the three
Python client docstrings — because a response field nobody has documented is not
finished.

### CI

`.github/workflows/pr-ci.yml` triggers on `pull_request: branches: [main]`, so it does
not run for a PR targeting `feat/server_team`. The checks above were run locally instead;
happy to re-run anything you would like to see.

### DIFF
--- a/MemoryCore/src/gateway/skill-archive-extraction-flag.test.ts
+++ b/MemoryCore/src/gateway/skill-archive-extraction-flag.test.ts
@@ -1,0 +1,125 @@
+/**
+ * The three archive entry points answer a success with `ok`/`status` and a `task_id`.
+ * That answer is accurate — `SkillTriggerService.archive()` writes the archive, appends
+ * the task and enqueues it — but it says nothing about whether extraction is configured
+ * to run at all. With `skill.extraction.enabled=false` the task is accepted and then
+ * cannot be executed, and until now no field in the response said so.
+ *
+ * These tests pin the field that says it, and pin what it must NOT say:
+ *   - it reports the configuration switch, never worker readiness or task completion;
+ *   - it is `null` — not `false` — when the resolved config cannot be reached;
+ *   - it does not change the archive contract, and it is not injected into errors.
+ */
+import { describe, test, expect, vi } from "vitest";
+import { handleExtract, handleConversationAdd, handleForceArchive } from "./skill-handlers.js";
+import type { SkillRouterDeps } from "./skill-handlers.js";
+import type { V2AuthContext } from "./v2-schemas.js";
+
+const auth: V2AuthContext = { apiKey: "k", serviceId: "inst-1" };
+const RID = "req-test";
+
+const logger = { debug: () => {}, info: () => {}, warn: () => {}, error: () => {} } as unknown as SkillRouterDeps["logger"];
+
+/** Only the two fields these handlers read off the resolved config. */
+const cfgWith = (enabled: boolean) => () => ({
+  extraction: { enabled, chunkMaxBytes: 1_000_000, headKeepBytes: 400_000, tailKeepBytes: 400_000 },
+  compress: { toolContentThresholdBytes: 4096, headBytes: 1024, tailBytes: 1024 },
+}) as unknown as ReturnType<NonNullable<SkillRouterDeps["getResolvedSkillConfig"]>>;
+
+const archiveOk = () => vi.fn(async () => ({ taskId: "task-1", archivedAtMs: 1_700_000_000_000, archiveKey: "k/1" }));
+
+function deps(over: Partial<SkillRouterDeps> & { archive?: ReturnType<typeof archiveOk>; handle?: unknown; buffer?: unknown } = {}): SkillRouterDeps {
+  const archive = over.archive ?? archiveOk();
+  return {
+    getSkillCore: () => undefined,
+    logger,
+    getResolvedSkillConfig: over.getResolvedSkillConfig,
+    resolveConversationAdd: async () => ({
+      trigger: { archive },
+      handler: { handle: over.handle ?? (async () => ({ status: "ok" })) },
+      buffer: over.buffer ?? {
+        readCurrent: async () => ({ messages: [{ role: "user", content: "hi" }] }),
+        readMeta: async () => ({}),
+        writeCurrent: async () => {},
+        writeMeta: async () => {},
+      },
+    }) as never,
+    ...over,
+  } as SkillRouterDeps;
+}
+
+const extractBody = {
+  user_id: "u1", team_id: "t1", agent_id: "a1",
+  messages: [{ role: "user", content: "how do I fix the exit line" }],
+};
+const addBody = { ...extractBody, session_id: "s1" };
+const forceBody = { space_id: "inst-1", user_id: "u1", team_id: "t1", agent_id: "a1", session_id: "s1" };
+
+describe("extraction switch in the archive responses", () => {
+  test("extraction off: the slice is still archived and the task still registered — and the response says the switch is off", async () => {
+    const archive = archiveOk();
+    const res = await handleExtract(extractBody, auth, RID, deps({ archive, getResolvedSkillConfig: cfgWith(false) }));
+    const data = res.data as Record<string, unknown>;
+    // the existing contract is untouched
+    expect(res.code).toBe(0);
+    expect(data.ok).toBe(true);
+    expect(data.task_id).toBe("task-1");
+    expect(data.archive_key).toBe("k/1");
+    // the task really was registered — this is not a "rejected before archiving" change
+    expect(archive).toHaveBeenCalledTimes(1);
+    // …and the caller can now see why nothing will come of it
+    expect(data.extraction_enabled).toBe(false);
+  });
+
+  test("extraction on: behaviour unchanged, and `true` is not a claim that extraction finished", async () => {
+    const res = await handleExtract(extractBody, auth, RID, deps({ getResolvedSkillConfig: cfgWith(true) }));
+    const data = res.data as Record<string, unknown>;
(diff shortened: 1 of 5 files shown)


Maxwell-Code07:

PR #1281, opened 2026-09-07
feat(proxy): add dynamic skill queue strategies

lyra-planet:
### Problem

The existing `session_init` path selects Skill only when a session starts and injects the listing into the system prompt:

```text
system + Skill0 + queue1 + queue2 + ... + queueN
```

After initialization, user queues do not trigger another listing query. A Skill created during the task remains invisible to that task, and a conversation that moves to another business flow keeps using its original Skill set.

Dynamic injection also cannot rely on client history. MemoryProxy modifies the request sent upstream, but the client stores and replays only its original conversation. Proxy-owned Skill blocks must therefore be reconstructed on every request.

### What Changed

This PR keeps `session_init` as the default and adds three queue-aware strategies:

| Strategy | Retrieval | Current queue | Historical blocks |
|---|---|---|---|
| `every_queue` | Configurable recent user queues (default 3) | Full BM25 listing | Restored byte-for-byte |
| `latest_only` | Configurable recent user queues (default 3) | Full BM25 listing | Removed from old queues |
| `adaptive_queue` | Configurable recent user queues (default 3) | New/forgotten Skill delta | Restored byte-for-byte |

```text
every_queue:    queue1 + Skill1 + queue2 + Skill2 + ... + queueN + SkillN
latest_only:    queue1 + queue2 + ... + queueN + SkillN
adaptive_queue: queue1 + Delta1 + queue2 + [none] + ... + queueN + DeltaN
```

### Request Flow

```mermaid
flowchart LR
    C[WorkBuddy or Codex client] -->|full Responses input| H[MemoryProxy handler]
    H --> Q[Configurable recent real user queues]
    Q -->|normalized, max 6000 chars| I[SkillInjector]
    I -->|team + agent + query| S[MemoryCore BM25 listing]
    S -->|TOP20 summaries| I
    I --> D{Strategy}
    D -->|every_queue| E[Restore snapshots + append full listing]
    D -->|latest_only| L[Strip old blocks + append latest listing]
    D -->|adaptive_queue| A[Filter by lastInjectedTurn + append delta]
    E --> U[Rebuilt upstream input]
    L --> U
    A --> U
    U --> M[Upstream LLM]
    M -->|skill_view| B[Proxy Skill bridge]
    B --> S
    M -->|business calls| T[Agent tools]
```

### Retrieval and Injection

- `extractRecentUserQueues()` reads the real Responses `input[]`, not the synthetic pipeline message.
[...]

PR #1359, opened 2026-09-12
fix(proxy): honor read-only mode in skill listing instructions

VittorioCai:
## Description | 描述

`skillRuntime.allowLlmWrite` is **false by default** — `DEFAULT_CONFIG` in
`MemoryProxy/src/config.ts`, and `config.example.yaml` documents it as such. Under that
default `skill-bridge` answers every write subpath with `40302` / HTTP 403, and
`SkillToolsInjector` correctly leaves `skill_patch` / `skill_create` out of the tool list
it renders for the model.

The `<available_skills>` header did not get the memo. `skill-injector.ts` rendered these
two sentences unconditionally:

```
If a skill has issues, fix it with the `skill_patch` skill-bridge tool.
After difficult/iterative tasks, offer to save the approach as a new skill
(`skill_create`). If a skill you loaded was missing steps, had wrong commands, or
needed pitfalls you discovered, update it before finishing.
```

So on a default deployment the model is told to use a tool it was never given, and the
call comes back 403. That costs a tool round-trip, and leaves the model holding an
instruction it cannot carry out. `allowLlmWrite` is not consulted anywhere in that file.

This splits the write sentences out of the header and renders them only when the switch
is on. With writes off the model is told what it can still do — report the problem in its
reply — and **no write tool is named**, so nothing in the read-only text invites a call
that would be rejected. Everything else in the block is untouched: the mandatory-load
directive, the curl reminder, the listing itself, and the footer.

The flag already existed one line away: `injection/index.ts` computed it for
`SkillToolsInjector`; it now passes the same value to `SkillInjector` as well.

**MemoryCore's copy of this header is deliberately not changed.** The file comment asks
that the two copies be kept in sync, so to be explicit about the divergence: the switch
is proxy-side and Core knows nothing about it, and Core's `SKILL_LISTING_HEADER` is
exported but never referenced by Core itself — `/v3/skill/listing` returns only the
`<available_skills>` lines — so on this path the header the model sees is the proxy's.
The comment now records this.

## Related Issue | 关联 Issue

None filed for this. Found while running the product end to end with the default
configuration.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

  New `MemoryProxy/src/injection/injectors/__tests__/skill-injector-readonly.test.ts` —
  9 tests, written before the fix (4 of them failing for the reason above), all passing
  after:

  - read-only renders no `skill_patch` / `skill_create` / "update it before finishing" /
    "fix it with";
  - read-only says what to do instead (report it), and says the session is read-only;
  - read-only keeps the rest of the block — the mandatory-load directive, `skill_view`,
    the listing verbatim, the footer;
  - write-enabled keeps both original sentences word for word;
  - the two renderings share the same opening and the same tail, so only that clause
    swaps;
  - omitting the flag falls back to read-only, matching the product default;
  - and three that drive the real injector through `prewarm()` with a stubbed core
    client, so the value is shown to travel config → injector → rendered block, off, on,
    and unset.

  ```
  cd MemoryProxy && npx vitest run src/injection/injectors/__tests__/skill-injector-readonly.test.ts
  Test Files  1 passed (1)
       Tests  9 passed (9)
  ```

  With `allowLlmWrite: true` the header is byte-for-byte what it was. The only other
  change is one argument at the registration site. `npx tsc --noEmit` reports the same
  **60** pre-existing errors on `feat/server_team` before and after. `BASE_REF=origin/feat/server_team
  bash MemoryCore/scripts/ci/check-skill-queue-isolation.sh` → PASS.

## Additional Notes | 其他说明

Scope, stated plainly:

- **Not** a change to what the bridge accepts or rejects. `WRITE_SUBPATHS` and the 40302
  branch are untouched; this only stops the prompt from asking for something that branch
  will refuse.
- **Not** a change to `SkillToolsInjector`, which was already correct.
- **Not** a claim about hosts that render the block themselves. The proxy path is what
  this fixes.
- An agent whose listing is empty never received these sentences to begin with — the
  injector returns no block when the listing is `(none)` — so this changes nothing for
  those sessions.

### The replacement wording is yours, not mine

One part of this is a product decision I had to make to open the PR, and it should be
yours: the sentence the read-only branch renders. What the model is told is prompt
surface, so treat that text as a placeholder — happy to take whatever phrasing you
prefer, or to drop the clause entirely and simply omit the write instructions with no
replacement. The gating is the part this PR is actually about.

The one thing I would keep for a non-taste reason: whatever replaces it, better not to
name a write tool. Something like "you cannot use `skill_patch` here" puts the tool name
back in front of the model, which is what invites the 403 in the first place.

### Overlap with #1281

[#1281](https://github.com/TencentCloud/TencentDB-Agent-Memory/pull/1281)
(`feat(proxy): add dynamic skill queue strategies`) touches the same two files, and one
of its hunks rewrites the exact line this PR rewrites — the `new SkillInjector({...})`
call in `injection/index.ts`. So whichever lands second will conflict there. The
resolution is to keep both fields:

```ts
new SkillInjector({
  coreSkill: config.coreSkill,
  queueStrategy: config.injection.skillQueueStrategy,   // #1281
  allowLlmWrite,                                        // this PR
}),
```

The two are otherwise independent: #1281 changes *where and when* the listing is
injected (`point` / `anchor` / `cacheStrategy` via a queue strategy) and its diff contains
no occurrence of `allowLlmWrite`, `skill_patch`, `SKILL_LISTING_HEADER`, or the write
sentences — it does not address this bug. Both PRs also add tests under
`injectors/__tests__/`, with different filenames, so there is no collision there.

Size: +43/−8 in the two source files (most of it comments explaining the switch), plus
103 lines of test.

### DIFF
--- a/MemoryProxy/src/injection/index.ts
+++ b/MemoryProxy/src/injection/index.ts
@@ -300,14 +300,14 @@ function buildPipelineBundle(config: ProxyConfig): PipelineBundle {
     // RAG-driven `<cloud_skills>` block. Calls /v3/skill/search at prewarm time.
     // When coreSkill is unconfigured (no serviceToken), the searchSkills call
     // will fail and the injector silently degrades to no <cloud_skills> block.
+    const allowLlmWrite = config.skillRuntime?.allowLlmWrite ?? false;
     registry.register(
-      new SkillInjector({ coreSkill: config.coreSkill }),
+      new SkillInjector({ coreSkill: config.coreSkill, allowLlmWrite }),
     );
 
     // Always inject the curl-recipe `<skill_tools>` block alongside the
     // dynamic `<cloud_skills>` block. Even when there are no skills to
     // recommend, the LLM still needs to know how to create / search them.
-    const allowLlmWrite = config.skillRuntime?.allowLlmWrite ?? false;
     registry.register(new SkillToolsInjector({ proxyBaseUrl: proxyBaseUrl!, allowLlmWrite }));
   }
 
--- a/MemoryProxy/src/injection/injectors/__tests__/skill-injector-readonly.test.ts
+++ b/MemoryProxy/src/injection/injectors/__tests__/skill-injector-readonly.test.ts
@@ -1,0 +1,103 @@
+/**
+ * `skillRuntime.allowLlmWrite` is false by default (`config.ts` DEFAULT_CONFIG,
+ * and `config.example.yaml` says so). Under it the bridge rejects every write
+ * subpath with 40302/403, and `SkillToolsInjector` correctly omits skill_patch /
+ * skill_create from the tool list it shows the model.
+ *
+ * The `<available_skills>` block is the one place that did not get the memo: its
+ * header told the model to patch a skill that has issues and to update a skill
+ * before finishing, no matter what the switch said. A model that complied called
+ * a tool it had never been given and got a 403 back.
+ *
+ * These tests pin both directions: read-only says report, write-enabled keeps the
+ * existing instructions verbatim.
+ */
+import { describe, test, expect } from "vitest";
+import { wrapAvailableSkillsBlock, SkillInjector } from "../skill-injector.js";
+
+const LISTING = "<available_skills>\n  - skl-abc: demo skill\n</available_skills>";
+
+describe("available-skills block honours the write switch", () => {
+  test("read-only: no instruction to edit a skill", () => {
+    const out = wrapAvailableSkillsBlock(LISTING, false);
+    expect(out).not.toMatch(/skill_patch/);
+    expect(out).not.toMatch(/skill_create/);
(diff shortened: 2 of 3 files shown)


VittorioCai:
One thing I should have flagged in the description: the read-only replacement sentence is wording I chose, not behaviour. Happy to take whatever phrasing you prefer, or to drop the clause entirely and simply omit the write instructions — the gating is the part that matters, the text is yours.

The only bit I would keep for a non-taste reason: whatever replaces it, better not to name a write tool. Saying something like "you cannot use skill_patch here" puts the tool name back in front of the model, which is what invites the 403 in the first place.

Maxwell-Code07:

PR #1305, opened 2026-09-09
feat(proxy): add structured skill access telemetry

ccdf1137:
# feat(proxy): add structured skill access telemetry

## Description | 描述

Skill Bridge currently records HTTP attempts without the resolved Skill ID and
version. A request version may be overridden by a session pin, and name lookups
resolve the ID in Core. Add response-derived `skill_id` / `skill_version` to
existing tool-call telemetry for successful get/get-by-name content responses.
Invalid JSON, nonzero business codes, missing versions and metadata-only reads
leave these fields empty/NULL. Responses remain unchanged when telemetry fails.

A pure extractor isolates parsing and evidence semantics. ClickHouse DDL and
idempotent migrations supply compatible defaults for older rows/writers.
This records proxy-observed loaded evidence only, not Skill adoption or success.

## Related Issue | 关联 Issue

Closes #1304.

Related to #1189 and #1216 (Core aggregate usage counters); this PR does not
implement those counters.

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [x] New feature | 新功能
- [x] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过（30 unit/handler tests, mocked dependencies）
- [ ] No existing features affected | 无影响现有功能（targeted compatibility verified; full integration pending）

## Additional Notes | 其他说明

- Base: feat/server_team at c387ea4534d08f3204d50a137ef55206d7c50301.
- `npm ci --no-audit --no-fund` succeeded without disabling scripts.
- Baseline `npm test` found no files. Added tests match existing Vitest include;
  full target-module run now passes 30 tests in 2 files.
- `npm run typecheck`: 60 pre-existing diagnostics, output unchanged before/after.
- No Proxy build script exists. Tests also pass under temporary Node 22.23.2:
  2 files / 30 tests. The system default remains Node 24.14.1.
- ClickHouse 25.8.33.6 integration: migrated a legacy table twice, confirmed
  pre/post-migration legacy writes default to empty/NULL, and queried a new row
  with skill_id/version/status/latency. No real Core/Redis/COS or Agent E2E claimed.
- Apply the two documented ALTERs before rolling out new writers. Startup
  migration is best effort; permission failures/races can still lose telemetry.
- No Skill body or credentials added to logs; existing request-body logging unchanged.
- No turnSeq inference, Core changes, usage counters or maturity changes.
[...]

PR #1360, opened 2026-09-12
fix(proxy): record completed skill download calls in telemetry

VittorioCai:
## Description | 描述

`bridge-telemetry.ts` states the contract as one `bridge_call` per completed upstream
request, and the main path keeps both halves of it: one emit once a response is in hand
(4xx and 5xx included), one emit when upstream never answers.

The `files/download` branch only has the second. Its own comment records why that one was
added:

```
埋点补齐: 与主路径 :822 对称, upstream 未响应也算一次调用。
之前这个 catch 分支静默 return, 导致 curl 视角"打了 N 次" CH 少一条。
```

The same hole was closed for the failure path and left open for the success path. So a
download that works produces no row at all, and the branch is silent exactly when it
succeeds.

Measured against the handler with a stubbed sink:

| path | `bridge_call` rows |
|---|---:|
| `files/read`, success | 1 |
| `files/download`, success | **0** |
| `files/download`, network error | 1 |

This emits once the response is in hand, mirroring the main path field for field, placed
before the three returns that follow it — status passthrough, unparseable envelope,
decoded bytes — so each of those counts as exactly one completed call and no path can
produce two. Response bodies, status codes and headers are untouched.

## Related Issue | 关联 Issue

None filed. Found while reading the bridge's telemetry paths against the contract in
`bridge-telemetry.ts`.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

  New `MemoryProxy/src/skill/__tests__/skill-bridge-download-telemetry.test.ts` — 6 tests,
  written before the fix. Three of them passed immediately (they pin behaviour that was
  already correct), three failed for the reason above and pass now:

  - success → exactly one row, `executedEndpoint: "files/download"`, `upstreamStatus: 200`;
  - upstream 4xx → exactly one row, real status recorded;
  - upstream responded but the envelope carries no usable content → still exactly one,
    neither zero nor two;
  - network error → the row that already existed is still there, `upstreamStatus: 0`;
  - the response itself is unchanged — the success path still returns the raw bytes and
    the skill's own content-type;
  - and a control: a `files/read` success on the main path emits exactly one row, which is
    the first line of the table above — measured, not inferred from reading that path.

  ```
  cd MemoryProxy && npx vitest run src/skill/__tests__/skill-bridge-download-telemetry.test.ts
  Test Files  1 passed (1)
       Tests  6 passed (6)
  ```

  The tests drive the real `createSkillBridgeHandler` with an injected fetcher and a
  seeded session, and capture emissions through a mocked telemetry module — no network,
  no ClickHouse.

  Emission only; no response body, status code or header changes, and no new failure mode
  — `emitBridgeToolCallTelemetry` already swallows sink errors so telemetry never blocks
  the request. `npx tsc --noEmit` reports the same **60** pre-existing errors on
  `feat/server_team` before and after. `BASE_REF=origin/feat/server_team bash
  MemoryCore/scripts/ci/check-skill-queue-isolation.sh` → PASS.

### Who reads these rows

Not a field nobody consumes: `MemoryCore/src/gateway/analytics/analytics-sql.ts` builds
its `bridge_calls` and `prev_bridge_calls` CTEs from `tool_call_logs` where
`kind = 'bridge_call'`, and uses them for call counts and per-session call rates. A
missing row on the path that succeeds biases exactly the metric that is supposed to show
the bridge being used. This reaches persisted statistics only where the ClickHouse
collection is configured; where it is not, the fix simply makes the two paths consistent.

### Scope

- **Not** a change to what the download branch returns, or when it returns it.
- **Not** a change to the schema of a `bridge_call` row — the same fields the main path
  already sends, with `executedEndpoint: "files/download"`.
- **Not** a change to the reject-telemetry path, which was already emitting on every
  early return.

### One thing I did not touch

The sibling comment above the catch-branch emit says `与主路径 :822 对称`, and that line
number no longer points at anything related — the main path's emits sit much further down.
My own comment deliberately refers to the main path by behaviour instead of by line, so it
cannot rot the same way. I left the existing `:822` alone to keep this diff to the bug;
say the word and I will fix it here or in a one-line follow-up.

### Overlap with #1305

[#1305](https://github.com/TencentCloud/TencentDB-Agent-Memory/pull/1305) also edits
`skill-bridge.ts`, but in a different region — an import near the top and one line inside
the *main* path's emit, adding `...extractResolvedSkillAccess(sub, resp.status, respText)`.
There is no hunk overlap with this change, which is ~300 lines earlier in the download
branch, and the two do not conflict. Worth noting for whoever reviews both: if #1305 lands,
the same spread would presumably belong on this new emit too, and I am happy to add it in
a follow-up or fold it in here — say which you prefer.

Size: +19 in `skill-bridge.ts` (the emit plus the comment saying why it sits where it
does), plus 116 lines of test.

### DIFF
--- a/MemoryProxy/src/skill/__tests__/skill-bridge-download-telemetry.test.ts
+++ b/MemoryProxy/src/skill/__tests__/skill-bridge-download-telemetry.test.ts
@@ -1,0 +1,113 @@
+/**
+ * `bridge-telemetry.ts` 的契约是每次上游请求完成都发一条 `bridge_call`,主路径照做:
+ * 拿到响应发一条(含 4xx/5xx),upstream 不响应也发一条。files/download 分支只补了后者
+ * ——那处 catch 的注释还写着"之前这个 catch 分支静默 return, 导致 curl 视角『打了 N 次』
+ * CH 少一条",也就是说同一类洞已经为失败路径补过,成功路径被落下了。
+ *
+ * 下游是真有人读的:Core Analytics 用这些行算调用次数与会话调用率。成功下载缺行会把
+ * 这些指标算低。
+ *
+ * 这些用例钉住"每条路径恰好一条",并且不改响应本身。
+ */
+import { describe, test, expect, vi, beforeEach } from "vitest";
+
+const emitted: Array<Record<string, unknown>> = [];
+vi.mock("../../memory/bridge-telemetry.js", () => ({
+  emitBridgeToolCallTelemetry: (row: Record<string, unknown>) => { emitted.push(row); },
+  emitBridgeRejectTelemetry: (row: Record<string, unknown>) => { emitted.push({ ...row, _reject: true }); },
+  agentSourceFromSessionKey: (k: string) => String(k).split(":")[0] || "claude-code",
+}));
+
+const { createSkillBridgeHandler } = await import("../skill-bridge.js");
+const { getSessionStore } = await import("../../session/store.js");
+
+const SESSION = "codebuddy:dl-test";
+const config = {
+  coreSkill: { endpoint: "http://core.invalid", serviceToken: "t", timeoutMs: 1000, serviceId: "sp-1" },
+  skillRuntime: { allowLlmWrite: false },
+} as never;
+
+const ctx = (sub = "files/download") => {
+  const req = new Request(`http://proxy/skill-bridge/v3/skill/${sub}`, {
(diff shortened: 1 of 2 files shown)


Maxwell-Code07:

PR #1365, opened 2026-09-13
fix(store): pass sendDimensions through in store-pool's embedding service

### LINKED ISSUE #1343: [Bug] store-pool.ts drops sendDimensions when building the embedding service — all embeddings fail silently and retrieval degrades to keyword-only

Zeming-Yuan:
`createSqliteStore()` builds the EmbeddingService without `sendDimensions`, so the option falls back to `true` and every embed request carries an explicit `dimensions` field.

Against an endpoint that rejects that parameter — SiliconFlow's `Pro/BAAI/bge-m3` answers `HTTP 400 {"code":20015,"message":"The parameter is invalid. Please check again."}` — every embed call fails. Because every call site treats embedding failure as non-fatal, the deployment degrades silently:

- `l1-writer` `[vec-dual-write]` writes metadata only, so `l1_vec` stays empty
- recall's vector half returns 0 candidates indefinitely (FTS keyword search still works, which masks it)
- `l1-dedup` falls back to "store all as new", so duplicates accumulate

Nothing surfaces beyond WARN lines. On the deployment where this was diagnosed it went unnoticed for 26 days, ending with 395 L1 memories stored without vectors and recall reporting `vec=0hits` on every query.

`core/store/factory.ts` passes the field, which is why the L0 capture path keeps working on the very same configuration — the two constructors simply disagree.

## Change

One line, mirroring `factory.ts`:

```diff
         dimensions: embCfg.dimensions,
+        sendDimensions: embCfg.sendDimensions,
         maxInputChars: embCfg.maxInputChars,
```

## Verification

Against `https://api.siliconflow.cn/v1/embeddings` with `model: Pro/BAAI/bge-m3`:

| request body | result |
|---|---|
| `{"input":["..."],"model":"Pro/BAAI/bge-m3"}` | `200` |
| `{"input":["..."],"model":"Pro/BAAI/bge-m3","dimensions":1024}` | `400`, `code 20015` |

With this change the service is constructed with `sendDimensions: false` as configured, and `l1_vec` receives embeddings again — verified end-to-end on a standalone gateway (`vec=0hits` before, `vec=15hits` after). Rows written while the bug was live keep their metadata as-is; re-embedding them from `l1_records` restores the vector half without touching metadata.

Fixes #1343

---

Note: `CONTRIBUTING.md` asks for PRs against `develop_server_team` or `master`, but neither branch exists in this repository (only `feat/server_team`, `feat/server`, `main`). I've targeted the default branch — happy to retarget if that's wrong.


Maxwell-Code07:

Zeming-Yuan:
Flagging for whoever reviews this: #1344 targets the same omission in the same file, and both patches make the same one-line change (`sendDimensions: embCfg.sendDimensions` in `createSqliteStore()`).

I double-checked there is no third call site — `createEmbeddingService` is only invoked from `factory.ts` (already forwards the flag) and `store-pool.ts` (this fix) — so either patch closes #1343 completely.

The test added in 52ba203 fails on the pre-fix code (2 failed) and passes with the fix (2 passed). No preference on which PR gets merged; just flagging the overlap so it doesn't cost extra review time.

PR #1366, opened 2026-09-13
fix(gateway): let flush tasks fall through to executeL1

Zeming-Yuan:
`executeFlush()` overrides the worker's fallback with `core.handleSessionEnd()`, which only re-enqueues another `flush` task. No code path in the flush flow ever reaches `runL1WithStore`, so `POST /session/end` never triggers extraction.

The worker already implements the intended behaviour, and `v2-router.ts` states it outright:

```ts
// MemoryCore/src/services/pipeline-worker.ts
case "flush": return this.executor.executeFlush?.(task, signal) ?? this.executor.executeL1(task, signal);

// MemoryCore/src/gateway/v2-router.ts
// "flush" tasks behave like L1 work (see executor.executeFlush fallback);
```

The override is what disabled it.

## Why it's worse than a wasted call

`flushSession()` cancels the idle timer *before* enqueuing the flush:

```ts
// MemoryCore/src/utils/stateful-pipeline-manager.ts
await this.stateBackend.removeTimer(effectiveInstanceId, `${sessionKey}:L1_idle`);
await this.stateBackend.enqueueTask({ type: "flush", ... });
```

So a no-op flush turns "idle-time catch-up extraction" into "idle-time extraction cancellation": the idle safety net is gone and extraction falls back to waiting for the next `everyNConversations` threshold. Host plugins that call `/session/end` on idle make this the common path rather than an edge case.

## Change

Remove the override so the fallback applies. Nothing else needs to change:

- `cascadeSchedule()` already treats `"flush"` like `"L1"` — resets `conversation_count` and advances the L2 timer;
- `executeL1`'s dedup guard only fires for `triggeredBy === "timer_scanner"`, so a flush still runs when the session count is low.

## Verification

Standalone gateway, one `/capture` followed by `POST /session/end`:

| | result |
|---|---|
| before | `Task completed: flush`, no `[l1-extractor] Extracting` line, no `markL1ExtractionComplete` |
| after | `markL1ExtractionComplete session=... extracted=1`, and the record is stored with a vector in `l1_vec` |

Part of #549 (Defect 2)

---

Note: `CONTRIBUTING.md` asks for PRs against `develop_server_team` or `master`, but neither branch exists in this repository (only `feat/server_team`, `feat/server`, `main`). I've targeted the default branch — happy to retarget if that's wrong.


Maxwell-Code07:

PR #1371, opened 2026-09-14
feat(memory-core): add direction A memory evaluation

Shujin-Li:
Add the Direction A low-cost causal memory-effect evaluation implementation, selected zero-provider tests, public-safe frozen results, portable analysis materials, and the FINAL2 report.

Key points:
- FULL/REMOVE continuous fixed-4 causal reference
- NORMAL-based low-cost evaluation path
- Mem2 source model + Evo residual adaptation
- 9/9 public selected test files passed, 99/99 tests, zero-provider
- portable R1 reconstruction passed
- Mem2 shows empirical positive evidence but is not formally certified
- Fresh cross-environment superiority was not demonstrated
- evaluation logic remains isolated from the normal production path

The final Markdown/PDF report and three figures are included under the Direction A documentation directory.


Maxwell-Code07:

PR #1390, opened 2026-09-17
fix(knowledge): raise wiki-summary maxOutputTokens 1024 → 4096

nicwn:
## Problem

Wiki ingest summaries came back empty (`summary generation done (len=0, empty=true)`) even when the LLM call itself succeeded with a valid API key.

## Root cause

Reasoning models can spend more than 1024 tokens on hidden reasoning before emitting any visible text. With `maxOutputTokens: 1024`, the summary call finished with `finishReason: length` and empty text — so every wiki ingest ended with an empty summary despite correct auth.

## Fix

Raise the cap to 4096. The result is still sliced to 256 chars downstream (`text.slice(0, 256)`), so the stored summary size and behavior are unchanged; only the budget for reasoning + text is corrected.

Found while debugging summary failures with a rotated key: auth was fixed first, but summaries still returned empty until this cap was raised.


Maxwell-Code07:

Issue #1396, opened 2026-09-17
[Feature] Per-layer (L1/L2/L3) model overrides and multi-provider LLM config for standalone mode


Madin-H23:
### Is this feature related to a problem? | 该功能需求是否与某个问题相关？

We run a standalone gateway (single user, `deployMode: standalone`) and hit two operational needs that the current single-channel LLM config cannot express:

**1. One model config cannot serve all four pipeline stages well.** The stages have opposite requirements:

- **L1 extraction** must return strict JSON inside a small budget. Reasoning/thinking models tend to spend the output budget on reasoning: with `maxTokens: 4096` we repeatedly observed truncated responses (`finishReason=length` → `No JSON array found`), and per #1395 those batches were then silently skipped.
- **L2 scene synthesis / L3 persona** are judgement + synthesis tasks that benefit from a **larger** budget and from thinking **enabled**.

Today `llm.model` / `llm.maxTokens` / `llm.timeoutMs` are global, so tuning one stage degrades another. There is currently no way — not even per-layer `maxTokens` — to express this in the config.

**2. Only one provider can be configured at a time.** `llm.baseUrl` / `llm.apiKey` / `llm.model` describe exactly one endpoint, and the `provider` key is already taken by the transport mode (`openai` | `proxy`). Switching between, say, a monthly-plan gateway and a pay-as-you-go fallback means editing the yaml and restarting; there is also no way to declare *which env var* holds a given key, so keys get pasted into the yaml file.

### Describe the solution you'd like | 描述你期望的解决方案

Two additions, independent and backwards compatible (both default to the current behaviour):

**(a) Per-layer overrides — `llm.layers.{l1,l2,l3}`**
```yaml
llm:
  model: deepseek-v4-flash        # global default (unchanged)
  maxTokens: 4096
  layers:
    l1:                           # strict JSON, no thinking, small budget
      model: deepseek-v4-flash
      maxTokens: 4096
      disableThinking: true
    l2:                           # synthesis: bigger budget + thinking
      maxTokens: 16384
      disableThinking: false
    l3:
      model: deepseek-v4-pro
      maxTokens: 16384
```
Semantics we use locally and would propose: any omitted field falls back to the global value; an unknown layer key under `layers` fails fast at startup (typos must not be silently ignored); `disableThinking` has to be applied where the request body is built (in our environment the AI SDK compatible path did not serialize it, so we inject it at the fetch layer).

[...]

Issue #1395, opened 2026-09-17
[Bug] L1 extraction failures advance the L0→L1 cursor and silently drop conversations


Madin-H23:
### Describe the bug | 问题描述

**L1 extraction failures still advance the L0→L1 cursor, so the affected conversations are skipped forever** — no retry, no warning, and the operator cannot see it.

When the L1 LLM call fails (429/quota, network error, timeout), or when the response cannot be parsed as a JSON array (e.g. output truncated by `max_tokens`, `finishReason=length`), the pipeline still calls `markL1ExtractionComplete()` with the processed slice's max `recorded_at`. Since the cursor is the only filter for the next L0 read, those rows are never re-read.

Two paths in `MemoryCore` (master):

**(1) LLM error is swallowed — `src/core/record/l1-extractor.ts:212-216`**
```ts
} catch (err) {
  logger?.error(`${TAG} LLM extraction failed: ${err instanceof Error ? err.message : String(err)}`);
  return { success: false, extractedCount: 0, storedCount: 0, records: [], sceneNames: [] };
}
```
The L1 runner does not consume `success` for L1, and `src/utils/pipeline-factory.ts:658` then advances the cursor unconditionally:
```ts
await checkpoint.markL1ExtractionComplete(sessionKey, totalStored, maxRecordedAtMs || undefined, lastSceneName);
```

**(2) An unparseable response is treated as a valid empty extraction**
The parse path returns `{ scenes: [], emptyReason: "no_json" | "not_array" }`. `emptyReason` currently has **no consumer** anywhere in `MemoryCore/src`, so a truncated/garbled response is indistinguishable from a genuine "nothing to remember" — and the cursor still advances.

### To Reproduce | 复现步骤
1. Standalone gateway (`deployMode: standalone`), a single session, any OpenAI-compatible provider.
2. Force L1 failures: (a) use a provider that is over quota (429), or (b) point L1 at a reasoning model with a small `maxTokens` so responses get truncated mid-JSON.
3. Capture a few conversation rounds, then wait for the L1 idle trigger (default 600s).
4. In the gateway logs, observe the failure immediately followed by a cursor advance:
```
ERROR [l1-extractor] LLM extraction failed: Failed after 3 attempts. Last error: quota exhausted, will reset later
INFO  [checkpoint] markL1ExtractionComplete session=<session>: extracted=0, cursor=1789633946176
WARN  [l1-extractor] No JSON array found in extraction response
[...]

PR #1401, opened 2026-09-17
feat(memory-core): per-layer LLM overrides (llm.layers.{l1,l2,l3})

Madin-H23:
## Description | 描述

Adds optional per-layer LLM overrides: `llm.layers.{l1,l2,l3}` with `model` / `maxTokens` / `timeoutMs`.

The stages have opposite needs: L1 extraction must return strict JSON inside a small budget (a reasoning model can burn the whole budget on reasoning → truncated response → nothing extracted, see #1395), while L2/L3 are judgement + synthesis tasks that benefit from a larger budget. With only global `llm.*`, tuning one stage degrades another.

```yaml
llm:
  model: deepseek-v4-flash      # global default, unchanged
  maxTokens: 4096
  layers:
    l1: { model: strict-json-model, maxTokens: 4096 }
    l2: { maxTokens: 16384, timeoutMs: 300000 }
    l3: { model: deepseek-v4-pro }
```

Semantics: omitted fields fall back to the global value; an absent `layers` section or layer behaves exactly as before; an unknown layer name (`l4`, `L1`, …) fails fast at startup instead of being silently ignored. `TDAI_LLM_*` env overrides intentionally do not apply to layer fields.

## Related Issue | 关联 Issue

Refs #1396 — this is capability (a). The multi-provider part (b) will follow as its own PR.

## Change Type | 修改类型

- [x] New feature | 新功能
- [ ] Bug fix | Bug 修复

## Self-test Checklist | 自测清单

Evidence:

- `npx vitest run` → **7/7** (new `src/config.layers.test.ts`): parse incl. fail-fast (`l4`, `L1`, non-object), merge semantics (only present fields change; absent section/layer returns the base config unchanged; base object is not mutated).
- **Isolated gateway run** (standalone, separate port + data dir, stub upstream): with `layers.l1.model: layer-l1-model` the *real* L1 run logs `model=layer-l1-model` and extraction completes — the override reaches the actual runner, not just the parser. Without the two wiring points below, the override was silently ignored; this check is what caught them.
- Absent `layers` → behaviour byte-identical to before (verified by the same run without the section).

## Additional Notes | 其他补充

Two load-bearing wiring points, both subtle in the existing code and both fixed here:

1. `shouldOverrideRunnerFactory()` must also return true for a layer that carries an override — otherwise standalone + `provider=openai` keeps reusing the host factory and the override never applies.
2. The gateway's top-level `llm` → `memory.llm` splice rebuilds the object **field by field**, so `layers` has to be carried over explicitly.

（中文摘要）新增可选的分层模型覆盖 `llm.layers.{l1,l2,l3}`（model / maxTokens / timeoutMs）：缺省字段回落全局、未配置时行为与现状完全一致、未知层名启动即 fail-fast。动机=四个阶段诉求相反（L1 要小预算严格 JSON，L2/L3 要大预算综合），只有全局配置时调一层必伤另一层（触发 #1395 的截断丢批次）。本地 vitest 7/7；并用隔离网关实跑验证层内模型真的到达 runner（`model=layer-l1-model`），过程中因此抓到两处会静默吞掉覆盖的接线点（override 判定 + 逐字段 splice）。多厂商（providers[]）将作为独立 PR 跟进。


Maxwell-Code07:

Issue #1146, opened 2026-08-24
[Bug] MemoryProxy Docker → upstream local joignable : fetch failed / 502 sur /v1/chat/completions


stef5962:
# [Bug] MemoryProxy Docker → local upstream reachable from container: `fetch failed` / 502 on `/v1/chat/completions`

**OpenClaw Version | 版本:** N/A (MemoryProxy standalone Docker)
**Plugin Version | 插件版本:** `agentmemory/memory-proxy:latest` (branch `feat/server_team`)
**Operating System | 操作系统:** Windows 11 / Docker Desktop 4.87.0
**System Specification | 系统配置:** Intel Core i7-12700 / 32GB RAM / RTX 3080 Ti 16GB VRAM

---

## Summary

MemoryProxy running inside a Docker container can reach the upstream for the **connectivity check**, but consistently fails when forwarding real `/v1/chat/completions` requests with:

```
502 Bad Gateway
pipeline.error stage=FORWARD error="fetch failed" error.name="TypeError"
```

Because the proxy logger does **not** show `err.cause`, the real underlying error (`ECONNREFUSED`, `ENOTFOUND`, `UND_ERR_CONNECT_TIMEOUT`, etc.) remains unknown.

**No public issue currently reproduces this exact case**: local upstream reachable from the container, connectivity check OK, but forwarding fails on `fetch()`.

---

## To Reproduce | 复现步骤

1. Start Unsloth Studio on the host listening on `127.0.0.1:8888`.
2. Start a `node-proxy` container on the same Docker bridge network as MemoryProxy, forwarding to the host upstream.
3. Start MemoryProxy with:
   - `upstream.url` set to the `node-proxy` address reachable from the MemoryProxy container (`http://172.18.0.3:8889/v1` or `http://node-proxy:8889/v1`)
   - all other features disabled (`tdai.enabled: false`, `injection.enabled: false`, `auth.enabled: false`, etc.)
4. Send a real OpenAI-compatible request:
   ```bash
   curl -X POST http://localhost:8096/v1/chat/completions \
     -H "Authorization: Bearer sk-unsloth-..." \
     -H "Content-Type: application/json" \
     -d "{\"model\":\"qwen3.8/Qwen3.8-27B-UD-Q4_K_S\",\"messages\":[{\"role\":\"user\",\"content\":\"Bonjour\"}]}"
   ```
5. Observe the response and MemoryProxy logs.

[...]

PR #1404, opened 2026-09-17
fix(proxy): expose safe network error cause

iyuan7:
## Problem

Node.js fetch network failures commonly surface as `TypeError: fetch failed`. The useful network diagnostics are stored in `Error.cause`, but MemoryProxy structured logs only exposed the outer error message and name.

## Root cause

Forward handlers pass the original error into `Pipeline.error()`, which delegates to the shared `log.error()` facade. `report/log.ts::emit()` copied only `error.message` and `error.name`, dropping structured `Error.cause` metadata before the configured backend received it. The local file logger also received only the caller data. WorkBuddy's forward catch additionally reduced the error to an informational string before entering structured logging.

## Fix

- Extract one level of allowlisted cause fields in the shared logging facade: name, code, errno, syscall, address, and port.
- Limit allowlisted cause strings to 200 characters.
- Intentionally omit the free-form `Error.cause.message` field.
- Send the same safe error attributes to the local file logger and configured backend.
- Route WorkBuddy forward failures through the existing `Pipeline.error()` path without changing forwarding behavior.

## Tests

Added focused regression coverage for:

- errors without a cause
- normal Error causes, without emitting their free-form messages
- Node-style network causes such as ECONNREFUSED
- free-form cause messages containing credential-like text are not emitted
- primitive causes
- arbitrary object causes
- throwing getters on allowlisted cause fields
- rejection of non-allowlisted credential fields

The RED test on the unmodified implementation failed four cause-related cases because only the outer message and name were present.

## Security / privacy

Free-form `Error.cause.message` is intentionally not logged. The logger does not serialize the complete cause object and does not recurse into nested causes. It reads only the fixed structured network metadata allowlist: name, code, errno, syscall, address, and port. Headers, Authorization values, cookies, API keys, request/response bodies, environment variables, arbitrary nested properties, and free-form cause messages are not inspected or emitted. Regression tests verify that credential-like cause messages and non-allowlisted authorization/apiKey values never reach the emitted attributes.

## Validation

- `npm test -- src/report/__tests__/log.test.ts` — passed, 8 tests
- `npm test` — passed, 1 file / 8 tests
- `npm run typecheck` — 60 existing repository diagnostics remain; the same 60 failures were reproduced on the unmodified base SHA `0b7bc097f963d3d53b088cd56f79b4bc868e0023`, with no diagnostics in the revised logger/test or WorkBuddy forward change area
- Build — not run because MemoryProxy/package.json defines no build script
- `git diff --check` — passed

## Scope

This PR improves diagnostics only. It does not attempt to fix the upstream/Docker networking failure itself. It does not change fetch configuration, retry behavior, timeouts, connectivity checks, or response handling.

Refs #1146


Maxwell-Code07:

PR #1406, opened 2026-09-17
feat(memory-core): opt-in disableThinking control for the standalone LLM path

### LINKED ISSUE #1403: [Feature] No way to control reasoning/thinking in the standalone LLM path (truncated responses become empty extractions)
### Is this feature related to a problem? | 该功能需求是否与某个问题相关？

Yes — a reasoning model can silently destroy an extraction batch.

In our standalone deployment the L1 layer runs with `maxTokens: 4096`. When the active model emits reasoning before the answer, the reasoning consumes the output budget, the visible content is cut off mid-JSON (`finishReason=length`, e.g. a 121-char fragment) and the extraction yields nothing. On one day this produced **8 such batches** for L1; combined with a provider quota outage they caused the data loss described in #1395.

There is currently **no switch for this anywhere in the standalone path**:

- `StandaloneLLMConfig` (`src/adapters/standalone/llm-runner.ts`) has no thinking/reasoning field;
- `llm-runner.ts` never touches the request body for it.

We also verified that routing it through the AI SDK is not sufficient on the compatible path: a fetch-layer probe showed the outgoing body had `thinking: undefined` when the field was passed via `providerOptions.openai.thinking`, i.e. it never reached the provider — so the value has to be injected where the request body is built.

### Describe the solution you'd like | 描述你期望的解决方案

A per-provider `disableThinking` (and per-layer, see #1396) honoured by the standalone runner, implemented at the request-body/fetch layer instead of relying on providerOptions:

```yaml
llm:
  providers:
    - name: some-gateway
      extra:
        disableThinking: true
```

We already run exactly that as a local patch (L1 = thinking off + small budget; L2/L3 = thinking on + large budget) and can send it as a PR.

- Only raising `maxTokens`: works around truncation but wastes budget and does not fix the strict-JSON case;
- Telling users to select non-reasoning models: loses the quality benefit on L2/L3 — which is exactly why the per-layer split in #1396 matters;
- Documenting it as a known limitation: leaves a silent failure mode in place.

### Additional context | 补充说明

Environment: standalone · Node 22 · OpenAI-compatible providers (multiple vendors, incl. monthly-plan gateways).

Relation to the other reports: #1395 makes a failed batch *deferred* instead of lost, and #1396 lets each layer pick its own model/budget — this report is the missing knob so L1 can reliably ask for strict JSON while L2/L3 keep reasoning enabled.

（中文摘要）standalone 链路**完全没有思考/推理开关**：L1 用 4096 预算时，推理模型会把预算烧在思考上、可见内容被截断成半截 JSON（`finishReason=length`），当日实测 8 个批次因此零产出（叠加限额即 #1395 的丢数据）。且经 fetch 层探针实证，走 AI SDK 的 `providerOptions.openai.thinking` 在兼容路径上**根本没有发到上游**（body 里 thinking=undefined）——必须在构造请求体处注入。建议加 per-provider（并配合 #1396 做到 per-layer）的 `disableThinking`；本机已按此实现并可用，方向认可即可提 PR。

Madin-H23:
## Summary

Reasoning-enabled models can silently spend the whole output budget on thinking: `content` comes back empty or truncated mid-JSON with `finishReason=length`, and L1 extraction then parses nothing — while the operator has no configuration knob to turn thinking off. We hit this in a production deployment with a GLM-series model through the standalone path: the same extraction batch went from `extracted=0` to normal output once thinking was disabled at the HTTP layer.

Two implementation notes that make the obvious approaches not work:

- `thinking` is not an OpenAI-standard parameter; each vendor has its own dialect. This PR injects `{"type": "disabled"}` — the convention used by Zhipu/GLM-style OpenAI-compatible endpoints, hence **opt-in**.
- The AI SDK's openai-compatible provider does not serialize `providerOptions.openai.thinking` into the request body (verified with a fetch probe), so the injection is done by wrapping `fetch` when the flag is on.

## What this does

- `llm.disableThinking?: boolean` on the standalone LLM config, on both config surfaces (plugin `llm.*` and gateway `llm.*`);
- three-state resolution: explicit config value wins; otherwise env `TDAI_DISABLE_THINKING === "true"` (strict match, so `"1"`/`"yes"` don't accidentally enable it); default off;
- when unset, no fetch wrapper is installed and requests are byte-identical to before;
- carried through the gateway `memory.llm` splice and `resolveRuntimeLlm` so all `StandaloneLLMRunner` construction sites (pipeline, per-call L1/L2/L3, seed) see it.

Usage:

```yaml
llm:
  baseUrl: https://open.bigmodel.cn/api/paas/v4
  model: glm-5.3-flash
  disableThinking: true   # or env TDAI_DISABLE_THINKING=true
```

## Testing

- New `src/adapters/standalone/llm-disable-thinking.test.ts` (10 tests): precedence matrix (explicit `false` overriding env, strict env matching), body rewrite (replaces an existing thinking block / injects when absent / non-JSON passthrough), fetch wrapper (string bodies rewritten, non-string bodies and unrelated headers untouched).
- `vitest run`: 10/10 passing.

Closes #1403

## 中文摘要

推理型模型会把输出预算烧在思考上，content 为空或 JSON 截断（finishReason=length），L1 蒸馏解析不到内容，而运维侧没有配置开关可关。实测 GLM 系模型经 standalone 路径在 HTTP 层关闭思考后同一批提取恢复正常。实现上有两个坑：`thinking` 非 OpenAI 标准参数（本 PR 注入 `{"type":"disabled"}`，智谱系兼容端点约定，故为 opt-in）；AI SDK 的 openai-compatible provider 不会把 providerOptions.openai.thinking 序列化进请求体（fetch 探针实测），因此改为在 fetch 层包装注入。新增 `llm.disableThinking` 三态判定（显式配置 > env TDAI_DISABLE_THINKING 严格匹配 > 默认关），未配置时请求与改动前逐字节一致；两个配置面 + gateway splice + resolveRuntimeLlm 全链透传。新增 10 个单测全过。


Maxwell-Code07:

Issue #1382, opened 2026-09-16
[Bug] With embedding disabled, recall degrades silently for CJK: short Chinese queries return 0 hits ("No usable FTS tokens from query")


harriffer:
> 中文摘要：关掉 embedding 后召回不是报错，而是静默退化——短中文查询（如“口令”）返回 0 命中，日志报 `No usable FTS tokens from query`；而 ASCII marker 能命中，所以冒烟测试看起来是绿的。建议二选一并写进文档：CJK 场景必须开 embedding，或者修 CJK 分词让短查询产生可用 token。

---

**Environment**
- OS: Windows 11, standalone mode, commit `8f2dc83`
- Integration: Hermes Agent + plugin `memory_tencentdb` (v1)
- Config: `embedding.enabled: false` (no embedding endpoint was available), `recall.strategy: hybrid`, `bm25.enabled: true`, `bm25.language: zh`

**Summary**
With embedding disabled, recall does not degrade loudly — it degrades silently, and for short Chinese queries it returns nothing at all.

**Observed**
- `recall_top_score` on returned memories is ~3e-6, i.e. there is effectively no ranking signal.
- Multi-character/latin queries work (a marker like `MARKER-ZQ-7198dfe0` is found).
- Short Chinese queries return zero hits and the log says `No usable FTS tokens from query`. Reproduced with: `口令`, `闪电阻尼器`, `专属测试口令` — all 0 hits.
- Net effect: a smoke test with an ASCII marker looks green, while day-to-day Chinese usage returns nothing.

**Why this matters**
For any CJK-language user, "recall" appears to work and quietly does not. This is worse than an explicit failure because the agent stops searching instead of escalating.

**Suggested fix — pick one and make it explicit**
1. Treat embedding as required for CJK: refuse to start (or log a prominent warning) when `embedding.enabled: false` and `bm25.language` is a CJK language.
2. Fix CJK tokenisation on the FTS path so short CJK queries produce usable tokens (jieba appears to be present but yields no usable tokens for short queries).
3. Document the limitation clearly in the README for CJK deployments.

Additionally, exposing `recall_top_score` in the recall response/log would let integrators notice this class of silent degradation without reading raw logs.

Maxwell-Code07:
Thank you so much for your attention and contribution! We will arrange an internal review for this PR shortly, and all feedback will be shared right here in the discussion.

Madin-H23:
/claim — taking this one.

[...]

PR #1408, opened 2026-09-17
fix(memory-core): make CJK recall work without jieba via character-bigram fallback

Madin-H23:
## Summary

`Fixes #1382` — with one correction to the report's diagnosis, verified locally.

On a machine where `@node-rs/jieba` loads, `buildFtsQuery("口令")` returns `"口令"` (non-null) — the `No usable FTS tokens from query` branch cannot fire for the report's queries. The failure the report observed lives one layer down: **when jieba cannot be loaded at all** (bundled deployments regularly ship without the matching native binary), the old fallback

- stored the **raw sentence** on the write side (`tokenizeForFts`), so FTS5's unicode61 tokenizer kept each contiguous CJK run as ONE giant token;
- matched whole runs on the query side.

Result: no CJK query term can ever hit, while ASCII words still match — exactly the report's symptoms (ASCII marker found, `口令` / `闪电阻尼器` / `专属测试口令` all 0 hits, smoke test green, everyday usage silent-zero).

## What this does

- new `segmentCjkFallback` — dictionary-less CJK segmentation used by **both** sides when jieba is unavailable:
  - index mode (write): CJK runs contribute unigrams + bigrams; latin/digit runs stay whole; script boundaries split mixed runs; single-char runs yield their unigram;
  - `queryMode` (query side): only bigrams/whole words — unigram query terms would OR-match unrelated documents, so precision is kept;
  - write/query alignment holds on all three backends (SQLite FTS5, TCVDB sparse, MongoDB `$search`) since they share `store/tokenize.ts`;
- a one-time `console.warn` when jieba fails to load, so degraded CJK tokenization is visible instead of silent (the report's "make it explicit" ask).

## Testing

17 new tests in `src/core/store/tokenize.test.ts`:

- pure segmentation: index/query modes, script boundaries (`部署API服务`), punctuation runs, empty input;
- degraded mode with jieba force-disabled: `buildFtsQuery("口令")` stays usable, the write side no longer stores the raw sentence, indexed/queried tokens stay aligned;
- jieba path via a deterministic fake dictionary;
- FTS5 end-to-end on `node:sqlite` (unicode61 default tokenizer, mirroring `memory-store.ts`): a short CJK query hits a document that returned **0 hits** before this change with jieba force-disabled — the exact #1382 scenario;
- ASCII marker queries keep working in degraded mode (the green smoke test still passes).

## 中文摘要

复现并修正 #1382 的根因判断：jieba 可加载时查询侧本可产出 token；真正的故障是 **jieba 原生模块缺失**（打包部署常见）时回退路径把整句原文写入 FTS，unicode61 将连续中文当一个 token，任何 CJK 查询都无法命中而 ASCII 仍可命中——与报告全部症状吻合。修复：新增 `segmentCjkFallback` 无词典回退（写侧 unigram+bigram、查询侧仅 bigram/整词保精度、脚本边界切分），读写两侧对齐、SQLite/TCVDB/MongoDB 三后端通用；jieba 加载失败时告警一次使降级可见。新增 17 个测试，含 node:sqlite FTS5 端到端回归（禁用 jieba 后短中文查询由 0 命中变为命中）。


Maxwell-Code07:

PR #1410, opened 2026-09-17
fix(proxy): stop Pi profile splitting sections inside nested XML blocks

### LINKED ISSUE #1411: [Bug] Pi profile splits sections inside nested XML blocks — memory block injected twice
`PiProfile.splitByPiLabels` opens a new top-level section on any `Label:` line **regardless of XML nesting**. Pi's real system prompt nests `<project_instructions path="...">` inside `<project_context>`, and the files those blocks carry (`AGENTS.md` and friends) routinely contain label-shaped lines of their own.

Because `PI_SLOT_MAP` routes both `skills` and `rules` to the `Guidelines` key, and `pipeline.ts` `applyInjection` lands the injected block on **every** segment matching that key, the memory block is injected twice — and the second copy is misplaced inside the project instructions. The enclosing `<project_context>` segment is also truncated at the nested label, which moves the `task_context` anchor.

**Version:** `0b7bc09` (branch `feat/server_team`, 2026-09-17). Introduced by #1126.

## Root cause

`MemoryProxy/src/injection/agents/pi/profile.ts`:

```ts
const LABEL_RE = /^([A-Z][A-Za-z ]*):\s*$/;
const XML_OPEN_RE = /^<([a-z_]+)>\s*$/;

for (const line of lines) {
  const labelMatch = LABEL_RE.exec(line);   // no nesting check
  const xmlMatch = XML_OPEN_RE.exec(line);  // no nesting check
  if (labelMatch || xmlMatch) {
    flush();
    currentKey = labelMatch ? labelMatch[1].trim() : xmlMatch![1];
    currentKind = "markdown_section";
  }
  buffer.push(line);
}
```

The real Pi prompt shape (see the fixture that shipped with #1126):

```
<project_context>

Project-specific instructions and guidelines:

<project_instructions path="/example/AGENTS.md">
# Example project instructions
- Follow existing patterns in the codebase.
</project_instructions>

</project_context>
```

Note `<project_instructions path="...">` carries an attribute, so it does not match `XML_OPEN_RE` — it is treated as plain body text, while the `Guidelines:`-style lines inside it **do** match `LABEL_RE` and split the section.

The pipeline injects on every match (`MemoryProxy/src/injection/pipeline.ts`):

```ts
if (key && segments.some((s) => s.key === key)) {
  const newSegments = profile.applyAnchor(segments, { key, relation: hook.anchor.relation }, text);
  sysMsg.blocks = [{ type: "text", content: profile.rebuild(newSegments) }];
  return;
}
```

`SkillInjector` (`{ slot: "skills", relation: "before" }`) and `TdaiProfileMemoryInjector` (`{ slot: "memory", relation: "inside_append" }`) both run through this path.

## To Reproduce

Drive the real `InjectionPipeline` with the Pi profile registered, using a realistic Pi system prompt whose `<project_instructions>` body contains a `Guidelines:` line:

```bash
cd MemoryProxy
npx tsx repro-pipeline.ts          # script attached below / in the linked PR
```

## Evidence

Measured on `0b7bc09` (unfixed) vs. the fix:

```
### UNFIXED (base 0b7bc09) ###
=== REAL PIPELINE RESULT ===
system prompt bytes : 671
injection count     : 2 (DUPLICATED)
marker positions    : 203, 448
nested block starts : 388
marker inside nested: true

### FIXED ###
=== REAL PIPELINE RESULT ===
system prompt bytes : 646
injection count     : 1 (correct)
marker positions    : 203
nested block starts : 388
marker inside nested: false
```

`marker inside nested: true` on the unfixed build is the misplaced second copy. Segment-level view:

```
before: segments = [null, "Available tools", "Guidelines", "project_context", "Guidelines"]
after:  segments = [null, "Available tools", "Guidelines", "project_context"]
```

The `project_context` section is also truncated at the nested label on the unfixed build (it ends before `</project_instructions>`), which is what shifts the `task_context` anchor.

## Expected behavior

A label-shaped line inside an open XML block is body text, not a section opener. The block must be injected exactly once, into the top-level section, and `parse → rebuild` must stay lossless (the profile's stated safety guarantee).

## Impact

Any Pi session whose project instructions contain a label-shaped line — e.g. `Guidelines:`, `Never:`, `Prefer:`, `Ownership rules:` — gets the skills/rules memory block twice, once in the wrong place. This is the common case, not an edge case: PI-Desktop's own `AGENTS.md` contains 10 such lines.

## Note on test coverage

The fixture that would have caught this (`pi/__tests__/fixtures/pi-system-prompt.txt`) was deleted in release commit `220af62` (v2.0.2-beta.1), so this path currently has no regression net. A fix PR with regression tests follows.

muzimu217:
🏆 **队伍 | Team：TenBase**

## Description | 描述

Fixes #1411.

`PiProfile.splitByPiLabels` opened a new top-level section on any `Label:` line regardless of XML nesting. Pi's real system prompt nests `<project_instructions path="...">` inside `<project_context>`, and the files those blocks carry (`AGENTS.md` and friends) routinely contain label-shaped lines of their own.

Two consequences, both reproducible on `feat/server_team`:

- **Duplicate injection.** `PI_SLOT_MAP` routes both `skills` and `rules` to the `Guidelines` key, and `pipeline.ts` `applyInjection` lands the block on *every* segment matching that key. A nested `Guidelines:` line therefore produced a second `Guidelines` section, and the memory block was injected twice — the second copy misplaced inside the project instructions.
- **Truncated block.** The enclosing `<project_context>` segment was cut at the nested label, moving the `task_context` anchor.

Fix: track XML nesting depth and skip label/opener matching at depth > 0. Depth is precomputed and suppression applies only when the tags balance, so a malformed prompt (stray or unclosed tag) keeps the previous line-by-line behaviour rather than swallowing every later section. `parse → rebuild` stays lossless.

## Related Issue | 关联 Issue

Fixes #1411

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

### 1. End-to-end through the real `InjectionPipeline`

`MemoryProxy/src/injection/agents/pi/__tests__/pipeline-injection.test.ts` drives the actual runtime path — `HookRegistryImpl` + `OpenAIAdapter` + `InjectionPipeline`, with the Pi profile registered as `agentProfiles["pi"]` exactly as the proxy wires it — against a realistic Pi system prompt whose `<project_instructions>` body contains a `Guidelines:` line. The hook is a test double mirroring `SkillInjector`'s anchor (`{ slot: "skills", relation: "before" }`).

Measured on the base commit vs. this PR, same input:

```
### UNFIXED (base 0b7bc09) ###
=== REAL PIPELINE RESULT ===
system prompt bytes : 671
injection count     : 2 (DUPLICATED)
marker positions    : 203, 448
nested block starts : 388
marker inside nested: true

### FIXED (this PR) ###
=== REAL PIPELINE RESULT ===
system prompt bytes : 646
injection count     : 1 (correct)
marker positions    : 203
nested block starts : 388
marker inside nested: false
```

`marker inside nested: true` on the unfixed build is the misplaced second copy. Segment-level view:

```
before: segments = [null, "Available tools", "Guidelines", "project_context", "Guidelines"]
after:  segments = [null, "Available tools", "Guidelines", "project_context"]
```

### 2. Regression tests

```
cd MemoryProxy && npx vitest run src/injection/agents/pi/__tests__/

without this change:  Tests  9 failed | 4 passed (13)
with this change:     Tests  13 passed (13)
```

The 9 failing cases cover both the parser (`splitByPiLabels` section keys, lossless rebuild, enclosing-block integrity) and the runtime path (injection count, placement relative to the nested block).

### 3. Baseline / blast radius

- `npx tsc --noEmit`: 60 errors, identical to the pre-existing baseline (the same count is cited in #1339); 0 of them in `injection/agents/pi`.
- `parse → rebuild` losslessness regression-checked against the real Pi prompt shape — `detect()`, segment count, and `rebuild(parse(x)) === x` all unchanged.
- The change touches only `PiProfile`; no public interface, slot map, or other profile is modified.

## Additional Notes | 其他说明

`MemoryProxy` has no CI job on `feat/server_team` — `.github/workflows/pr-ci.yml` only triggers on PRs targeting `main` and runs with `working-directory: MemoryCore`. The verification above was therefore run locally, as the repo's own module READMEs instruct. If you'd like the proxy suite wired into CI, I'm glad to send that as a separate PR.


Maxwell-Code07:

Issue #1296, opened 2026-09-08
recall.scoreThreshold applied across incommensurable score scales (BM25/cosine 0-1 vs RRF ~0.03)


kevinyuan:
## Summary
`recall.scoreThreshold` (default `0.3`) is applied as a single absolute cutoff, but the scores it is compared against live on **different, incommensurable scales** depending on the search strategy.

## Details
- **keyword** (FTS/BM25 via `bm25RankToScore`): 0–1 scale → `0.3` is a plausible cutoff.
- **embedding** (cosine, `1 - distance`): 0–1 scale → `0.3` is plausible.
- **hybrid** (RRF, `k=60`): scores are ~`0.016`–`0.05` (RRF of two lists). An absolute `0.3` cutoff here would filter out essentially everything.

So the same `scoreThreshold` constant means very different things per strategy. In practice the hybrid path avoids the problem only because it does not apply the threshold at all (see related issue about `searchHybrid` ignoring `_threshold`). The keyword path additionally has a "document set is small → return below-threshold results anyway" fallback, which masks the scale mismatch.

## Suggestion
Pick one of:
1. Apply per-source thresholds *before* fusion (cosine ≥ x, BM25-score ≥ y) instead of one absolute value on the fused score;
2. Drop the absolute threshold for fused/RRF results and rely on top-N;
3. At minimum, document the per-strategy meaning of `scoreThreshold` so operators do not set a value that is a no-op or over-filters depending on strategy.

Maxwell-Code07:
We truly appreciate your valuable feedback! Our team will carry out an internal review and assessment promptly, and we will share our thoughts with you here soon.

Madin-H23:
/claim — taking this one.

**Proposed direction (2 + 3 combined, no behavior break):** the current split — `scoreThreshold` applies to the single-source keyword/embedding paths (0–1 scales) while the client-side RRF fusion deliberately ignores it — is defensible but **implicit**. The same yaml value silently changing meaning per strategy is the real bug. Plan:

1. Make the hybrid path's decision explicit in code: document (and assert in tests) that fused RRF scores are ranking-only and `scoreThreshold` is intentionally not applied after fusion;
2. Document the per-strategy semantics of `recall.scoreThreshold` in the README/config comments (0–1 on cosine/BM25-mapped scores; no-op on fused RRF results, which rely on top-N);
[...]

PR #1412, opened 2026-09-17
fix(memory-core): document and surface the per-strategy semantics of recall.scoreThreshold

Madin-H23:
## Summary

`Fixes #1296` — with **no behavior change to any strategy** (option 2 + 3 from the report).

The report is right that the same `recall.scoreThreshold` yaml value silently means different things per strategy. What the code does today is actually coherent — the problem is that it is **implicit**:

- **keyword**: FTS5 BM25 mapped to 0–1 via `bm25RankToScore` → threshold applies;
- **embedding**: cosine 1-distance, 0–1 → threshold applies;
- **hybrid**: the two legs run WITHOUT thresholds and are fused by RRF (k=60). Fused scores are `~1/(60+rank)` ≈ 0.016–0.05 — a **ranking signal, not a relevance probability** — so the threshold is deliberately not applied post-fusion, and hybrid relies on top-N (`maxResults`). Applying a 0–1 threshold there would empty hybrid recall entirely; applying per-leg thresholds pre-fusion would truncate the ranked lists the fusion needs.

## What this does

1. `RecallConfig.scoreThreshold` doc comment now states the per-strategy semantics and why fused RRF results must not be thresholded;
2. `searchMemories` / `searchHybrid` doc comments state the same, and the stale "keyword = Jaccard" description (the implementation is FTS5 BM25) is fixed;
3. observability: the hybrid (client RRF) and native-hybrid debug lines now report the top fused score and that the threshold was not applied — so operators can see when the setting is a no-op for their strategy (the visibility ask also raised by #1382);
4. `searchMemories` is exported `@internal` for tests.

## Testing

New `auto-recall-threshold.test.ts` (5 cases, fake store + fake embedder, no LLM):

- hybrid (client RRF): fused results with all scores < 0.3 are returned untouched — pins the no-threshold behavior;
- native hybrid: server-side RRF rankings returned unfiltered;
- the hybrid debug line reports `topRrfScore=` and `threshold not applied`;
- keyword and embedding paths still filter by the 0–1 threshold (regression guard against over-correcting).

## 中文摘要

按 #1296 选项 2+3 组合实施、不改任何策略的行为：keyword（BM25 映射 0–1）与 embedding（cosine 0–1）照常应用阈值；hybrid 的 RRF 融合分（~1/(60+rank)）是排序信号而非相关概率，"刻意不应用阈值"从隐式变为显式——配置注释、函数文档、调试日志（输出 topRrfScore 与 threshold not applied）三处钉死，并修复 searchMemories 过时的 Jaccard 描述。新增 5 个回归测试钉住该分工。


Maxwell-Code07:

PR #1413, opened 2026-09-17
fix(memory-proxy): make Pi profile splitting aware of XML nesting

Madin-H23:
## Summary

`Fixes #1411`. The root cause analysis in the issue is exact — this PR implements the fix.

`splitByPiLabels` opened a new top-level section on any `Label:` line regardless of XML nesting. Pi's real system prompt nests `<project_instructions path="...">` inside `<project_context>`, and the files those blocks carry (AGENTS.md and friends) routinely contain label-shaped lines. Two failure modes followed:

- a nested label line split the enclosing section, truncating `<project_context>` and moving the `task_context` anchor;
- since `PI_SLOT_MAP` routes both `skills` and `rules` to `"Guidelines"`, a nested `Guidelines:` line produced a second segment with that key, and `applyAnchor` injects into **every** matching segment — the memory block landed twice, the second copy misplaced inside the project instructions.

## What this does

- track XML nesting depth while scanning: the split decision uses the depth **before** the current line, so label-shaped lines and nested tags at depth > 0 stay in the enclosing section's body;
- the depth tracker recognizes attribute-bearing opens (`<project_instructions path="...">` — the old `XML_OPEN_RE` only matched bare tags) and close tags, clamped at 0;
- top-level behavior is byte-identical for prompts without nesting: bare opens still open sections, a top-level attribute-bearing open tag still stays plain body (unchanged), non-nested label splitting is untouched;
- `pipeline.ts` needs no change — the double-injection disappears once segmentation stops producing duplicate anchor keys.

## Testing

First test suite for MemoryProxy (`vitest` was configured but had zero tests) — `src/injection/agents/pi/__tests__/profile.test.ts`, 7 cases:

- the #1411 fixture: exactly one `Guidelines` section, `<project_context>` intact through the nested block (no truncation), the memory block injected exactly once and absent from the project instructions, lossless rebuild;
- non-nested prompts split exactly as before;
- a top-level attribute-bearing open tag stays plain body (unchanged) while still guarding the labels it carries;
- stray close tags never drive depth negative.

Verified the 4 regression cases fail on the pre-fix code.

## 中文摘要

按 #1411 的根因分析实施修复：`splitByPiLabels` 扫描时增加 XML 嵌套深度跟踪——深度>0 处的 `Label:` 行与嵌套标签一律留在所属节体内，不再错误开节；深度跟踪同时识别带属性的开标签（旧 `XML_OPEN_RE` 只认裸标签）与闭合标签（钳制不小于 0）。无嵌套的提示词分段逐字节不变（顶层带属性开标签仍是正文的行为也保持）；`pipeline.ts` 无需改动——分段不再产生重复锚点键后，双重注入自然消失。为 MemoryProxy 补上首个测试套件（7 例，含 #1411 fixture 的"恰好注入一次、project_context 完整、无损重建"回归，已验证修复前 4 例失败）。


Madin-H23:
Closing in favor of #1410 per the comparison in #1411 — it predates the issue, drives the real InjectionPipeline end-to-end, and handles malformed prompts more safely. Thanks @muzimu217 for the precise write-up.

PR #869, opened 2026-08-08
fix(proxy): parse tdai.anonymous config block so anonymous fallback works

mambo-wang:
## Problem

`buildConfig()` in `MemoryProxy/src/config.ts` only extracted `endpoint` / `apiKey` / `serviceId` from the `tdai` YAML block, silently dropping the `anonymous` section. As a result `config.tdai.anonymous` was always `undefined` at runtime, so the anonymous fallback in `deriveTdaiIdentity()` could never activate — anonymous clients (e.g. CodeBuddy without auth/session) never persisted L0 memories even when `tdai.anonymous.enabled: true` was configured.

## Fix

- `config.ts`: parse `tdai.anonymous` (`enabled` / `teamId` / `agentId` / `userId`) with sane defaults (`default` / `codebuddy` / `anonymous`)
- `types.ts`: add `anonymous` to `RawYamlConfig` and `TdaiConfig`
- `tdai/identity.ts`: accept the anonymous config + `agentSource`, fill only the identity fields that are actually missing (explicit opt-in via `enabled: true`)
- `handler.ts` / `anthropicHandler.ts`: always derive identity (with anonymous fallback) instead of skipping derivation when prompt injection is skipped — L0 writing and injection are now independent

## Description | 描述
修复 MemoryProxy 在解析 `tdai` YAML 配置块时**静默丢弃 `anonymous` 子块**的问题，使无 auth / 无 session 的匿名客户端（如 CodeBuddy）也能通过默认身份兜底正常写入 L0 记忆。

根因：`buildConfig()` 中只从 `yaml.tdai` 提取了 `endpoint` / `apiKey` / `serviceId`，`anonymous` 段从未被解析进 `ProxyConfig`。因此 `deriveTdaiIdentity()` 永远读不到 `config.tdai.anonymous`，匿名兜底无法启用；当 `team_id` / `user_id` / `agent_id` 任一缺失时直接返回 `null`，匿名客户端的 L0 记忆既无法写入也无法被检索。

改动内容：
- `config.ts`：解析 `yaml.tdai.anonymous`（enabled / teamId / agentId / userId，带默认值），注入 `ProxyConfig.tdai.anonymous`
- `types.ts`：为 `TdaiConfig` / `RawYamlConfig` 补充 `anonymous` 字段类型定义
[...]

PR #1418, opened 2026-09-18
fix(proxy): restore memory identity when auth is disabled

iyuan7:
## Problem

When `auth.enabled=false`, MemoryProxy correctly forwards requests without rejecting the caller, but `verifyUserKey()` returns no proxy-verified `userId`. The handlers then used that empty identity for session initialization and the memory pipeline, so a request carrying a valid MemoryCore caller key could silently bypass session initialization, injection, and L0 recording.

## Root cause

Proxy-auth verification identity and MemoryCore memory identity were treated as the same value. With proxy auth disabled there is intentionally no verified proxy `userId`, even though the caller key can still be resolved by MemoryCore. The empty value reached the session/cache gates before the authoritative caller identity was available.

## Fix

- Add a small shared `resolveMemoryUserId()` helper in the MetadataClient layer.
- Keep a verified proxy `userId` when auth is enabled.
- When proxy auth is disabled and a caller key is present, resolve the real `user_id` through the existing authenticated `POST /v3/meta/user/get` API using `x-tdai-user-key`.
- Use the resolved `memoryUserId` consistently for SessionStore binding/recovery, session initialization, prewarm, injection, TDAI identity, mem commands, and L0 recording.
- If no caller key exists or resolution fails, continue forwarding but leave per-user memory/session features unavailable; no anonymous/shared identity is created.
- Keep `systemUsers` matching based on proxy-verified identity only.
- Remove raw caller-key prefixes and response-body fragments from metadata failure logs.

## Regression tests

Added `MemoryProxy/src/__tests__/auth-disabled-memory-bypass.test.ts` covering:

- auth-enabled verified identity without an extra lookup;
- auth-disabled valid caller-key resolution and complete session/memory flow;
- authenticated `/v3/meta/user/get` request shape and headers;
- no identity reuse across different caller keys;
- no-key and invalid-key forwarding with memory bypass and no L0 write;
- auth-enabled invalid-key rejection;
- system-user matching not using the separately resolved memory identity;
- credential-like caller keys and response bodies absent from logs.

The regression was RED on the original flow (`userId=""` -> session-init bypass -> no identity/L0) and is GREEN with this change.

## Compatibility / security

Auth-enabled valid and invalid-key behavior is preserved. Raw caller keys are never used as `userId` or emitted in logs. The change does not alter MemoryCore ACL checks, tenant isolation, or system-user semantics, and it does not introduce an anonymous memory bucket.

## Validation

- `npm test -- src/__tests__/auth-disabled-memory-bypass.test.ts` — 1 file, 9 tests passed.
- `npm test` — 1 file, 9 tests passed.
- `npm run typecheck` — 60 diagnostics, identical to the upstream baseline; no new diagnostics introduced.
- Build not run: `MemoryProxy/package.json` defines no build script.

`npm install` in the temporary upstream baseline worktree reported 4 existing dependency vulnerabilities (2 moderate, 2 high); no dependency changes are included here.

## Scope

This PR restores the existing per-user memory/session path when proxy auth is disabled and a valid MemoryCore caller key is available. It does not modify the bearer verification fix in #1349, add the anonymous fallback proposed by #869, change gateway auth, or alter MemoryCore endpoints/ACL behavior.

## Related

Refs #1341


Maxwell-Code07:

dangzitou (commented):
## Review

未发现 blocker。auth-disabled 场景现在可以用 caller key 通过 metadata 服务恢复 memory identity，同时没有 caller key 时不会伪造 per-user identity，也不会把一次失败解析变成请求失败。

### Evidence

- `cd MemoryProxy && npx vitest run src/__tests__/auth-disabled-memory-bypass.test.ts`：9/9 passed。
- `cd MemoryProxy && npm test`：1 个测试文件、9/9 passed。
- 用例覆盖 auth enabled 已有 verified user 时不额外查询、auth disabled 按 caller key 解析、无 key 返回 null、解析失败继续转发，以及不同 key 不能复用缓存 identity。

### Follow-up

当前证据是 helper 级测试，尚未覆盖 Anthropic/OpenAI/Codex/WorkBuddy 的真实 handler 请求。建议至少补一个 auth-disabled + session-init 的 handler 回归，验证 metadata 成功时能注入并写入 L0，解析失败时仍保持原始请求转发且不产生错误 identity。

PR #352, opened 2026-07-01
fix(offload): preserve namespaced model ids

RerankerGuo:
## Description | 描述

Fix `offload.model` parsing for provider/model refs whose model id itself
contains slashes, for example:

```text
siliconflow/deepseek-ai/DeepSeek-V4-Flash
```

The previous `split("/", 2)` path truncated that model id to `deepseek-ai`,
which can make the local offload LLM request fail with a bad model id. The same
split also prevented exact `contextWindow` lookup for slash-containing model
ids.

This PR keeps the change intentionally small:
- adds an offload-local `parseOffloadModelRef()` helper that splits only at the
  first `/`,
- uses it when initializing the local offload LLM client,
- uses it for exact provider model `contextWindow` lookup,
- adds focused tests for namespaced model ids and malformed / alias refs.

## Related Issue | 关联 Issue

Fix #24

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能
- [ ] Documentation update | 文档更新
- [ ] Code optimization | 代码优化

## Self-test Checklist | 自测清单
- [x] Verified locally | 本地验证通过
- [x] No existing features affected | 无影响现有功能

## Verification | 验证

```bash
PATH=$HOME/.nvm/versions/node/v24.15.0/bin:$PATH npm_config_cache=/tmp/npm-cache-tam-24 npx vitest run src/offload/model-ref.test.ts
PATH=$HOME/.nvm/versions/node/v24.15.0/bin:$PATH npm_config_cache=/tmp/npm-cache-tam-24 npm test
PATH=$HOME/.nvm/versions/node/v24.15.0/bin:$PATH npm_config_cache=/tmp/npm-cache-tam-24 npm run build
git diff --check
```

Results:
- `src/offload/model-ref.test.ts`: 3 tests passed.
- `npm test`: 5 files / 70 tests passed.
- `npm run build`: plugin and scripts build passed.

## Notes | 说明

There is another open implementation (#240). This PR takes a narrower route by
keeping the parser local to the offload module instead of moving the
clean-context runner parser, while still covering both the local LLM model id
and context-window lookup paths.

PR #33, opened 2026-05-17
Fix offload model parsing for namespaced model ids

kylezhang:
## 背景

`config.offload.model` 现在用 `split("/", 2)` 拆 `provider/model`。遇到这类模型 ID 时：

```text
siliconflow/deepseek-ai/DeepSeek-V4-Flash
```

实际传给模型服务的 model 会变成 `deepseek-ai`，后面的 `DeepSeek-V4-Flash` 被截掉，最终导致接口返回 badRequest。

## 修改

- 新增共享的 `parseModelRef`，只按第一个 `/` 拆 provider，后面的内容保留为完整 model id。
- `offload.model` 本地 LLM 初始化改用同一套解析逻辑。
- `contextWindow` 查找也改用同一套解析，避免带命名空间的模型 ID 匹配失败。
- 保留 `clean-context-runner` 原有导出，避免影响已有引用。
- 补充 parser 测试，并更新 offload model 配置说明。

## 验证

```bash
npx vitest run src/utils/model-ref.test.ts
npm run build:plugin
git diff --check
```

Fixes #24

Issue #24, opened 2026-05-16
[Feature] 希望offload modelId兼容 类似 siliconflow/deepseek-ai/DeepSeek-V4-Flash 这样的id


havcino:
### Is this feature related to a problem? | 该功能需求是否与某个问题相关？请描述

config.offload.model: "siliconflow/deepseek-ai/DeepSeek-V4-Flash" badRequest

### Describe the solution you'd like | 描述你期望的解决方案

offload/index.ts line 354 split ["siliconflow", "deepseek-ai/DeepSeek-V4-Flash"]

Maxwell-Code07:
Hi @havcino，感谢反馈！
确实是个待优化的点，我们会尽快修复这个解析逻辑。感谢提出！🙏

leoge007:
你们自己设计的时候能否考虑一下实际执行场景下的LLM成本，哪有那么多人都是opus4.7+gpt5.5双修不限量。

YOMXXX:
Triage update: kept #352 as the canonical fix for this issue.

Rationale: #352 is the narrowest mergeable implementation I found. It preserves slash-containing model IDs after the provider separator for both local LLM initialization and context-window lookup, and adds focused regression coverage.

Verified locally:
- COREPACK_ENABLE_AUTO_PIN=0 pnpm vitest run src/offload/model-ref.test.ts
- COREPACK_ENABLE_AUTO_PIN=0 pnpm test
- COREPACK_ENABLE_AUTO_PIN=0 pnpm build
- git diff --check

Duplicate cleanup:
- Closed #240 in favor of #352.
- Marked #33 as superseded by #352.

PR #1419, opened 2026-09-18
fix(offload): preserve namespaced model ids

AgileInnov8tor:
## Description | 描述

Offload local-mode LLM calls truncated namespaced model ids at the second slash, so `siliconflow/deepseek-ai/DeepSeek-V4-Flash` was treated as model `deepseek-ai` instead of `deepseek-ai/DeepSeek-V4-Flash`. Parsing now splits only on the first slash.

Rebased onto current `main` (not a reopen of #33). Merge kept `main`'s `collect` mode, auth-profile API-key fallback, and `disableThinking`.

## Related Issue | 关联 Issue

Related: #24
Related: #33, #352

## Change Type | 修改类型
- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

`parseModelRef` cases (simple id, namespaced id, trim, rejects) passed via bun import of `src/utils/model-ref.ts`. Full `vitest` suite was not run (worktree dependency install failed).

## Additional Notes | 其他说明

Head is `AgileInnov8tor:fix/offload-namespaced-model-ids` because origin push to TencentCloud is disabled. Original work is kylezhang's `3191dae`.

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -139,7 +139,7 @@
         "description": "Context Offload 设置 — 多层上下文压缩系统（独立开关，默认关闭）",
         "properties": {
           "enabled": { "type": "boolean", "default": false, "description": "是否启用 Context Offload（默认关闭，不影响 Memory 功能）" },
-          "model": { "type": "string", "description": "Offload 使用的 LLM 模型（格式: provider/model），未填写时使用 openclaw 默认模型" },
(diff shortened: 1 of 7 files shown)


AgileInnov8tor:
Closing: opened in error. Product source should not have been changed.

PR #1420, opened 2026-09-18
fix(metadata): bind permission queries to authenticated caller

ccqq01:
## Summary

Bind caller-sensitive metadata queries to the authenticated `x-tdai-user-key` principal.

## Vulnerability

`/v3/meta/asset/list-accessible` and `/v3/meta/acl/check` accepted `user_id`/`user_key` from the request body and passed those values to the metadata service. Any user with a valid key could therefore query another user's accessible assets or probe ACL decisions for that user, including private/restricted assets.

## Fix

Both handlers now require an authenticated user and overwrite the body subject with `ctx.userId` before calling the service. The request shape remains backward compatible; spoofed identity fields are ignored.

## Validation

- `git diff --check`
- Reviewed against the existing V3 authentication flow and caller-scoped service methods.

Related: #848 (the existing PR explicitly leaves `acl/check` and `asset/list-accessible` subject resolution as follow-up work).


ccqq01:
Closing this PR pending author approval before submission.

PR #1423, opened 2026-09-18
fix(pi): disable developer role for strict upstreams

### LINKED ISSUE #1422: Pi provider sends developer role to strict OpenAI-compatible upstreams

N/A — Pi 0.84.1

### Plugin Version | 插件版本

feat/server_team @ 0b7bc09; pi-plugin 0.1.0

### Operating System | 操作系统

Windows 11

_No response_

### Describe the bug | 问题描述

The first-class Pi provider registers a reasoning OpenAI Chat Completions
model without setting `compat.supportsDeveloperRole`.

Because the configured base URL is the local MemoryProxy URL rather than the
actual upstream URL, Pi does not detect strict upstreams such as DeepSeek.
It sends the system prompt with `role: "developer"`.

Strict OpenAI-compatible upstreams reject that role with HTTP 400, so Pi gets
no model response and the downstream tool, memory capture, and skill-mining
flow cannot proceed.

### To Reproduce | 复现步骤

1. Start MemoryProxy with a strict OpenAI-compatible upstream, such as DeepSeek.
2. Configure Pi 0.84.1 to load `MemoryCore/pi-plugin` and use the `tdai` provider.
3. Send a normal prompt, for example `reply OK`.
4. Inspect the forwarded OpenAI Chat Completions request or upstream response.

The request contains a `developer` system-prompt message and the strict
upstream returns HTTP 400.

### Expected behavior | 预期行为

The Pi TDAI provider should explicitly set
`compat.supportsDeveloperRole: false`, causing Pi to send the system prompt
as `role: "system"` for proxy-routed requests.

_No response_

### Additional context | 补充信息

This is independent of #1421 skill sync. It was found during real Pi/Proxy
validation and belongs to the existing Pi provider integration.

kbob3687-hub:
## Summary

- Explicitly disable the OpenAI Chat Completions `developer` role for the Pi TDAI provider.
- The local Memory Proxy URL hides the eventual upstream from Pi's compatibility detection; strict OpenAI-compatible upstreams such as DeepSeek reject `developer` messages.
- Add a regression test for the provider model compatibility declaration.

## Validation

- `npm test`
- `npx tsc --noEmit`
- `npm pack --dry-run`
- `git diff --check`

Closes #1422


Maxwell-Code07:

PR #1427, opened 2026-09-18
fix(proxy): resolve Pi sessions from bridge L1

### LINKED ISSUE #1424: [Bug] Pi sessions are missing from Skill and Memory Bridge L1 prefix fallback

N/A — Pi 0.84.1

### Plugin Version | 插件版本

feat/server_team @ 0b7bc09; MemoryProxy

### Operating System | 操作系统

Windows 11 (source-level, OS-independent)

N/A

### Describe the bug | 问题描述

The Skill Bridge and Memory Bridge L1 session resolvers do not include the `pi:` agent prefix.

The main OpenAI-compatible handler stores Pi sessions in SessionStore under the composite key `pi:<sessionId>`. However, both bridge implementations only probe the bare session ID, `codebuddy:<sessionId>`, and `claude-code:<sessionId>`.

Therefore, an initialized Pi session already present in the same process is always missed by the bridge L1 fast path. Under normal conditions this is masked by the persistent L2 binding fallback. If the binding repository is unavailable, returns no binding, or encounters a transient read failure, both bridges reject the request with HTTP 401 even though the initialized Pi session is still available in L1 memory.

Affected locations:

- `MemoryProxy/src/skill/skill-bridge.ts` (`loadSessionIdsL1`)
- `MemoryProxy/src/memory/memory-bridge.ts` (`loadSessionIdsL1`)

### To Reproduce | 复现步骤

1. Start MemoryProxy with the first-class Pi adapter and session initialization enabled.
2. Complete Pi session registration for a session such as `session-123`.
3. Confirm that SessionStore contains the initialized session under `pi:session-123`.
4. Call either the Skill Bridge or Memory Bridge with `x-conversation-id: session-123`.
5. Make the L2 binding unavailable or have `getBinding()` return null, while leaving the same-process L1 session intact.
6. Observe that `loadSessionIdsL1()` probes only the bare, `codebuddy:`, and `claude-code:` keys. It does not probe `pi:session-123`.
7. The bridge returns HTTP 401 `session not initialized` despite the valid Pi session being present in L1.

### Expected behavior | 预期行为

Both bridges should resolve an initialized Pi session from the L1 key `pi:<sessionId>` without requiring the L2 binding fallback.

The implementation should also avoid introducing another independently maintained agent-prefix list if possible. Candidate prefixes should be shared with or derived from the registered `AgentKind` values, with regression coverage for both the Skill Bridge and Memory Bridge.

Typical behavior when L1 misses and L2 cannot recover the binding:

```text
[skill-bridge] session=session-123 L1 miss → L2 binding lookup
HTTP 401
{"code":40101,"message":"[skill-bridge] session not initialized; cannot derive identity"}
```

The Memory Bridge has the equivalent failure:

```json
{"code":40101,"message":"[memory-bridge] session not initialized; cannot derive identity"}
```

### Additional context | 补充信息

This was found while validating the first-class Pi integration from #1126.

It is related in pattern to #1246: both problems are caused by hardcoded agent name lists drifting from the `AgentKind` registry. However, #1246 concerns URL whitelist normalization, while this issue concerns the Skill and Memory Bridge L1 session lookup.

The healthy L2 binding path currently masks the problem, which is why normal Pi skill-sync and memory-search flows can still succeed. The bug becomes visible when L2 recovery is unavailable even though the same-process L1 session remains valid.

kbob3687-hub:
## Summary

- Resolve bare bridge session IDs against every first-class agent source, including `pi:<sessionId>`.
- Move the L1 candidate-key lookup into one shared helper used by both the Skill Bridge and Memory Bridge.
- Derive the runtime candidate list from the `AgentKind` registry so a newly registered agent cannot be omitted from either bridge.
- Add regression coverage for Pi L1 resolution with no L2 binding fallback.

## Validation

- `npm test`
- Targeted TypeScript check for the new registry and shared resolver
- `git diff --check`

Closes #1424

Related: #1246 covers a separate URL whitelist prefix registry; this PR only fixes bridge L1 session lookup.


dangzitou (commented):
## Review

未发现 blocker。这个改动把 bare session id 映射为各 agent source 的候选 key，同时保留已经带 source 前缀的 key，适合 Pi 等 bridge 只传裸 session id 的接入场景。

### Evidence

- `cd MemoryProxy && npx vitest run src/session/__tests__/session-key-candidates.test.ts`：3/3 passed。
- `cd MemoryProxy && npm test`：1 个测试文件、3/3 passed。
- 测试覆盖 bare id 展开为 `claude-code:`、`pi:` 等候选、复合 id 不再重复展开，以及候选顺序；Memory Bridge 和 Skill Bridge 都调用了同一个 helper。

### Follow-up

目前测试集中在 helper，没有覆盖真实 Hono Memory Bridge/Skill Bridge 请求。建议补一个 Pi bare session id 的 HTTP 回归，确认 L1 命中时返回的 identity 正确且不会错误落到 L2；这样可以验证用户实际接入路径，而不是只验证候选列表生成。

kbob3687-hub:
@dangzitou Thanks for the review and the concrete test suggestion. Added the HTTP regression in commit `2beb72f`.

The new test mounts the actual Hono Memory Bridge handler and sends `POST /memory-bridge/v3/atomic/query` with a bare Pi `x-conversation-id`. It seeds only `pi:<sessionId>` in L1 and injects an observable L2 binding mock, then asserts:

- the request succeeds through the real bridge route;
- the upstream body contains the Pi session's `user_id`, `team_id`, `agent_id`, and `task_id`;
- the L2 `getBinding` fallback is never called.

Validation: `cd MemoryProxy && npm test` -> 2 files, 4/4 tests passed.

Maxwell-Code07:

PR #1430, opened 2026-09-18
feat(memory-core): add an opt-in time window for LLM pipeline tasks

Zeming-Yuan:
## Description | 描述

Adds an opt-in time window for LLM-backed pipeline tasks. While the window is closed, `L1`/`L2`/`L3`/`flush` tasks are deferred to the next opening instead of being run.

Deferring matters more than it looks. `moveToDeadLetter()` cleans up by removing the session's `L1_idle` and `L2_schedule` timers, and nothing re-arms them — so a task that fails hard enough to be dead-lettered takes that session's extraction with it permanently. A deferred task leaves the L1 cursor untouched, so the L0 backlog is still there at the next opening: the work lands late instead of never.

The gate is **off by default**, so existing deployments behave exactly as before.

| Variable | Default | Meaning |
| --- | --- | --- |
| `TDAI_LLM_WINDOW` | `off` | `on` enables the gate |
| `TDAI_LLM_WINDOW_HOURS` | `9-18` | Weekday hours kept closed; weekends always open |

The gate sits ahead of the `permitPool` acquire — a deferred task does no work, so it should not hold a pipeline permit while it waits.

## Related Issue | 关联 Issue

None. This came out of a standalone deployment where the extraction LLM competes with interactive work during business hours; happy to open an issue first if that is preferred.

## Change Type | 修改类型
- [ ] Bug fix | Bug 修复
- [x] New feature | 新功能

## Self-test Checklist | 自测清单

## Additional Notes | 其他说明

20 test cases across `src/utils/llm-window.test.ts` and `src/services/pipeline-worker.test.ts` — timer-member mapping per task type (`flush` → `L1_idle`, `L3` → `L3_deferred`), team/agent-scoped members, the ack path, the unacked path when re-arming fails, and the untouched pass-through when the window is open.

`MemoryCore/src` has no tsconfig (the module builds through tsdown with `dts: false`), so the assertions run through vitest. Both changed source files were additionally checked with `tsc --strict --noEmit` against a scratch config and report no errors.


Maxwell-Code07:

PR #1431, opened 2026-09-18
fix(proxy): version injection cache by hook config

iyuan7:
## Problem

Persistent session-init hook cache entries are keyed by session identity and logical hook ID, but not by the configuration that generated the cached blocks.

After changing `injection.externalGatewayUrl` and restarting MemoryProxy, an existing session can therefore continue receiving a cached `<skill_tools>` or `<tdai_memory_tools>` block containing the old gateway URL. Shared cache backends can also let differently configured proxy nodes reuse the same generated block.

## Root cause

`hook.id` is stable while the generated content changes with `proxyBaseUrl` and, for `SkillToolsInjector`, `allowLlmWrite` and `SKILL_VIEW_MODE`. `ContextBlock.metadata.cacheKey` cannot invalidate this lookup because it only exists after a hook has executed; `session_init` cache lookup happens before execution.

## Fix

Add an optional configuration fingerprint to content-producing hooks and a shared `hookCacheStorageKey()` helper. The helper keeps the logical hook ID unchanged for registry, observers, and logs, while appending a deterministic short SHA-256 variant only to the opaque persistent storage key.

Both `prewarmAll()` and request-time `InjectionPipeline` reads/self-heal writes use the same helper. `SkillToolsInjector` versions by normalized gateway URL, write permission, and skill view mode; `TdaiMemoryToolsInjector` versions by normalized gateway URL.

The URL/config values are hashed and never written into Redis hash fields, KV path segments, or SQLite hook IDs.

## Regression tests

The focused test was RED on the base commit:

```text
old config prewarms old URL
new pipeline with the same session and hook ID reads the old storage entry
old URL is injected and new URL is absent
```

It is GREEN with this change and covers:

- external gateway URL change for Skill tools
- same-configuration cache hit
- `allowLlmWrite` change
- `SKILL_VIEW_MODE` change
- external gateway URL change for TDAI memory tools
- opaque deterministic fingerprints and legacy hooks without a variant

## Compatibility

- Hooks without `cacheVariant` continue using their existing logical hook ID.
- Existing stale entries are not deleted or migrated; the new configuration no longer addresses them. They are cleaned through the existing cache/session lifecycle.
- No SQLite, Redis, or KV schema/layout migration is required.
- Session, tenant, and user isolation keys are unchanged.

## Scope

This PR fixes the stale persistent injection hook-cache aspect of #1235 only. It does not change fallback network-interface selection, add runtime config management, modify cache TTLs, or change MemoryCore behavior.

## Validation

- `npm test -- src/injection/__tests__/injection-cache-config-version.test.ts` — 1 file, 7/7 tests passed
- `npm test` — 1 file, 7/7 tests passed
- `npm run typecheck` — 60 diagnostics on this branch; the same 60 pre-existing diagnostics were reproduced on the unmodified upstream base, with no errors in changed files or the new test
- `git diff --check` — passed
- `MemoryProxy/package.json` defines no build script; build not run

## Related

Refs #1235


Maxwell-Code07:

PR #1435, opened 2026-09-19
fix(gateway): surface store degraded state in GET /health

### LINKED ISSUE #1432: [Bug] /health reports status=ok while the store is degraded — captures, L1 writes and search silently no-op
### Environment

N/A — source-level analysis of MemoryCore @ `feat/server_team` (0b7bc09). Line numbers refer to that baseline; function names are stable anchors.

### Summary

The sqlite `VectorStore` enters `degraded = true` when the sqlite-vec extension fails to load or when schema initialization fails (corrupt DB file, disk full):

- `MemoryCore/src/core/store/sqlite/memory-store.ts:383` / `:399` — the two init-time catch blocks set `degraded` and log once.

From that moment **every operation silently no-ops**:

- `upsertL1()` → `[L1-upsert] SKIPPED (degraded mode)`, returns `false`
- L0 upsert → `[L0-upsert] SKIPPED (degraded mode)` — **new captures are dropped, not queued**
- `L1-search` → skipped (line ~1329), `deleteL1()` → `false` (line ~1446), `countL1()` → `0` (line ~1583)

The only signal is the init-time error log. `GET /health` — which the code itself says operators and orchestrators (k8s liveness, docker health-check) rely on — never reflects it:

```ts
// MemoryCore/src/gateway/server.ts:1462-1477 (handleHealth)
status: this.core.getVectorStore() ? ok : degraded,
stores: { vectorStore: !!this.core.getVectorStore(), ... }
```

`handleHealth` only checks that a store object *exists*; it never calls `store.isDegraded()`. `isDegraded()` is currently consumed only by `pipeline-factory.ts:321` (JSONL read fallback) and contract tests — nowhere in the health/observability path.

### Impact

A corrupt DB file or a failed sqlite-vec load (e.g. after a dependency update) produces: memories silently stop being generated, captures are silently lost, recall returns empty — while `/health` stays green and the panel shows a healthy instance. This is the same user-visible family as #1386 (green health, dead feature), but at the persistence layer and with data loss on the capture path.

### Expected behavior

`/health` should surface per-store health, e.g. `stores.vectorStore: ok | degraded` with the init `reason` string, and the overall `status` should become `degraded` when the only persistence store is in no-op mode. The skill stores expose the same `isDegraded()` and share the same blind spot.

### Suggested direction

- `handleHealth`: consult `store.isDegraded()` (and skill stores) when building the response; include the init `reason`.
- Optionally log a recurring (rate-limited) degraded-mode warning on the no-op paths, since today the one-shot init log is easy to miss in rotated logs.

Related: #1155 (proposes self-healing keyed on `degraded === true` — consistent with surfacing it), #1386.

Madin-H23:
## Related issue

Fixes #1432

## Summary

The sqlite store enters degraded mode (every operation silently no-ops — L0 upserts are dropped, L1 upserts skipped, searches and deletes no-op) when the sqlite-vec extension fails to load or schema init fails (`sqlite/memory-store.ts`, the two init-time catch blocks). The only signal was the init-time error log, while `GET /health` kept reporting `status=ok` because `handleHealth` only checked that a store object exists — it never consulted `store.isDegraded()`. Operators and orchestrators relying on the health endpoint (k8s liveness, docker healthcheck) saw a healthy instance while captures were being silently lost.

Changes:

- new `gateway/health.ts`: `resolveHealthStatus(vectorStore)` — `"degraded"` when there is no store **or** the store reports `isDegraded()`;
- `handleHealth` uses it for the overall `status` and adds `stores.vectorStoreDegraded` next to the existing presence flags (additive field, existing consumers unaffected);
- `HealthResponse.stores` type extended accordingly; the single construction site is `handleHealth`.

The skill stores expose the same `isDegraded()` and could be folded into the same response as a follow-up; kept out to stay minimal.

## Verification

- `pytest`-equivalent: `npx vitest run src/gateway/health.test.ts` → **3 passed**
  - healthy store → `ok`; degraded store → `degraded`; no store → `degraded` (previous behavior preserved)
  - the degraded-store case fails against the previous `getVectorStore() ? "ok" : "degraded"` logic — this is the regression the issue describes
- `npx vitest run src/gateway` → all green (no other gateway tests affected)

## Additional context

None.


Maxwell-Code07:

PR #1426, opened 2026-09-18
fix(gateway): never strand L0 backlog on a stale conversation_count + standalone manual L1 drain endpoint

Madin-H23:
## Summary

Companion to the cursor semantics in #1395 / #1397 — same subsystem, different failure mode: **the cursor correctly stays behind, but the deferred backlog can still strand forever.**

**The defect (observed in a production standalone deployment, 2026-09-17):** `executeL1` skipped timer-fired L1 tasks whenever `conversation_count === 0`. That counter is not a reliable "nothing pending" signal — a completing L1 resets it to zero *unconditionally* while a fresh capture increments concurrently. If the reset lands after the increment, the counter reads 0 while real L0 rows sit past the cursor:

```
22:42:54  capture A → warmup L1 starts (slow provider call)
22:42:55  capture B → count 0→1, idle timer armed (600s)
22:42:58  L1(A) completes → count reset to 0   ← swallows B's increment
22:52:55  idle timer fires → "L1 skipped: already processed (count=0)"
          → B's two rows stranded: timer consumed, no new capture re-arms it
          → UI shows "2 轮未蒸馏" indefinitely, L1 receives nothing
```

This is precisely the state that #1397's defer-on-failure semantics depend on being recoverable from: deferred batches need a pickup path, and the timer path had a hole a counter race could open permanently.

## What this does

**Part 1 — never strand (executeL1):** the cheap counter fast-path stays, but a "0" only sets a flag; the skip is finalized *after* `resolveStore`, once the L0 source of truth confirms there are truly no rows past the session's cursor (new helper `countPendingL0Rows`, cursor via new `TdaiCore.getL1Cursor`). When the counter lied, the task proceeds and logs `counter race guard`.

**Part 2 — rescue + control:** new standalone-only `POST /v2/pipeline/l1/trigger` `{ session_key }` enqueues an immediate L1 via `StatefulPipelineManager.enqueueL1Drain` — operators can drain pending L0 now instead of waiting for everyN/idle thresholds, and can rescue batches stranded before the guard existed. Service mode 404 (mirrors `/v2/pipeline/status`); 503 when the pipeline manager is not wired. Drain tasks carry no `triggeredBy` marker, so they bypass the timer dedup path (same as the existing auto-drain enqueue). New deps thunk `getStatefulPipelineManager` because the manager is assigned after the router deps are built.

## Testing

[...]

PR #1436, opened 2026-09-19
fix(gateway): don't report store failures as successful deletes

### LINKED ISSUE #1434: [Bug] DELETE /v1/memories returns 200 on store failure — deleteL1 conflates not-found with error
### Environment

### Summary

`deleteL1()` returns `false` for two very different outcomes — "record not found" and "the store failed" — and the gateway's batch-delete endpoint turns that boolean into a `200` with a count, so store failures are reported as success:

```ts
// MemoryCore/src/core/store/sqlite/memory-store.ts
deleteL1(recordId, filter?): boolean {
  if (this.degraded) return false;          // failure mode #1: degraded no-op
  ...
  } catch (err) {
    this.logger?.warn(`${TAG} delete failed (non-fatal): ...`);
    return false;                            // failure mode #2: real error, swallowed
  }
```

```ts
// MemoryCore/src/gateway/v2-router.ts (~1328-1352)
for (const id of ids) {
  const ok = await store.deleteL1(id, deleteFilter);
  if (ok) { deletedCount++; deletedIds.push(id); }
}
...
return successEnvelope<AtomicDeleteData>({ deleted_count: deletedCount }, requestId);  // 200 regardless
```

A client cannot distinguish "these ids did not exist" from "the database is broken and nothing was deleted". While the store is degraded (see #1432) **every** delete returns `200 { deleted_count: 0 }` indefinitely; a transient error mid-batch silently under-deletes and the response still reads as success.

### Impact

- Reconciliation callers that rely on `deleted_count` silently diverge from reality.
- Operators debugging "why can't I delete these memories" get a 200 and no error anywhere but a warn log.
- In degraded mode the endpoint looks healthy while never deleting anything.

### Expected behavior

The store layer should let the router distinguish the two outcomes — e.g. a `throwOnError`-style opt-in (same pattern as `queryL0GroupedBySessionId` added for the L1 guard) or a structured return `{ deleted: boolean; error?: Error }`:

- store error → `5xx` error envelope;
- genuine miss → current `200` with the lower `deleted_count`.

The same conflation exists in `deleteL1Batch()` (`false` on failure vs `true` on success-with-zero-matches) and in the skill-store delete paths, so the fix is best shaped at the interface level rather than patched per endpoint.

### Related

#1432 (degraded mode is invisible to `/health`, which is what makes the eternal `200 { deleted_count: 0 }` state undetectable), #1426 (introduced the `throwOnError` opt-in pattern for read paths).

Madin-H23:
## Related issue

Fixes #1434

## Summary

`deleteL1()` returned `false` for two very different outcomes — "record not found" and "the store failed" (degraded mode, DB errors) — and the gateway batch-delete endpoint folded that boolean into `deleted_count` and answered `200`. A swallowed failure was therefore reported as success with a lower count, indistinguishable from a genuine miss; while the store was degraded the endpoint answered `200 { deleted_count: 0 }` indefinitely.

Changes, following the same opt-in pattern as `queryL0GroupedBySessionId` (see #1426):

- `IMemoryStore.deleteL1` gains `opts?: { throwOnError?: boolean }`;
- sqlite and tcvdb implementations rethrow instead of returning `false` when opted in (degraded mode and caught errors); every other caller keeps the non-fatal `false` contract;
- mongodb already propagates errors; the param is accepted for interface parity;
- the delete endpoint's per-id loop opts in and maps a store failure to a `502` error envelope, while genuine misses keep the current `200` + lower `deleted_count` semantics.

## Verification

- `npx vitest run src/core/store/sqlite/memory-store.l1-delete.test.ts` → **4 passed**
  - missing record → `false` (unchanged); existing record → `true`
  - failed store (closed handle) → `false` by default (non-fatal contract preserved), **rejects** with `throwOnError`
  - red-check: the rethrow case fails against the unpatched store (verified via `git stash`), so it covers the fix rather than passing incidentally
- `v2-router.ts` has the only `deleteL1` call site in the gateway; no other callers are affected.

## Additional context

The same not-found/error conflation exists in `deleteL1Batch()` and the skill-store delete paths; if maintainers prefer, those can follow the same opt-in pattern as a follow-up rather than widening this diff.


Maxwell-Code07:

PR #1437, opened 2026-09-19
fix(store): don't cache a StorePool entry whose init failed

### LINKED ISSUE #1433: [Bug] StorePool.getStore() caches a store whose init() failed and never retries it — a transient DB outage at first contact permanently degrades the instance
### Environment

### Summary

`StorePool.getStore()` (`MemoryCore/src/core/store/store-pool.ts`) registers the freshly created store in the pool **before** initializing it, swallows the init error, and hands the store to the caller regardless:

```ts
this.pool.set(instanceId, { pooledStore, configFingerprint: fingerprint, lastAccessedAt: now });  // ~line 222

try {
  await pooledStore.store.init();
} catch (e) {
  this.logger.warn(`${TAG} Store init failed for ${instanceId}: ${e}`);   // ~lines 236-239: swallowed
}
return pooledStore;   // broken store served to the caller
```

Every subsequent `getStore()` for the same instance + config fingerprint takes the cache-hit path (~lines 194-196) and **never re-runs `init()`** — there is no unhealthy flag, no bounded retry, no eviction.

### Impact

- **tcvdb / mongodb backends**: `init()` is the connection/table check. One transient network error or DB restart at first contact leaves a permanently broken store in the pool. All reads then hit the stores' blanket `catch → return []` (e.g. `tcvdb/memory-store.ts:1276-1279` and siblings), writes fail silently — a full silent memory outage for that instance while `/health` stays green (it compounds with the health blind spot; see #1432 about degraded state never being surfaced).
- **sqlite backend**: `init()` catches its own failure classes internally and sets `degraded = true` (`sqlite/memory-store.ts:383/399`), so the pool-level catch rarely fires — but the instance ends in the same permanent no-op mode all the same. The pool cannot distinguish "initialized" from "degraded/no-op".

### Expected behavior

A store whose `init()` failed should not be cached as if healthy. Options, in increasing invasiveness:

1. On init failure, `closeEntry()` + evict so the next `getStore()` re-creates and re-inits (natural retry at next touch);
2. Keep the entry but mark it unhealthy and retry `init()` with bounded backoff on subsequent `getStore()` calls;
3. At minimum, rethrow after logging so the first caller fails loudly instead of silently serving no-ops.

### Related

#1155 — the plugin-layer `_storeInitCache` has the same "failed init cached forever" flaw in `initStores()`; the self-healing pattern proposed there (validate cached entry health on access, discard and re-init) maps directly onto `StorePool` as well. That issue also motivates exposing `degraded`/closed state so the health check can be meaningful (#1432).

Madin-H23:
## Related issue

Fixes #1433

## Summary

`StorePool.getStore()` registered the pool entry **before** initializing the store and swallowed init errors, so one transient DB outage at first contact left a broken store cached forever: every later `getStore()` hit the cache (same config fingerprint) and never re-ran `init()`. On the tcvdb/mongodb backends that meant all reads hit the stores' blanket `catch → return []`, writes failed silently, and `/health` stayed green for the lifetime of the process.

Changes:

- on init failure the pool now **closes** the half-initialized store, **drops** the entry, and **rethrows** the original error — the first caller sees the real failure instead of silently serving no-ops;
- the next `getStore()` for that instance re-creates and re-inits the store, which is the natural retry (a recovered DB heals on next touch, a persistent outage keeps failing loudly);
- a successful init is still cached exactly as before; the sqlite backend's internal degraded-mode handling (its `init()` returns normally) is unchanged — that state's visibility is addressed separately in #1435 / issue #1432.

## Verification

- `npx vitest run src/core/store/store-pool.test.ts` → **3 passed**
  - init throws → `getStore()` rejects, `store.close()` called, pool entry dropped (`pool.size === 0`);
  - two consecutive `getStore()` calls after a failure → `init()` attempted **twice** (natural retry, previously 1);
  - init succeeds → store cached and returned (existing behavior unchanged, single init);
  - red-check: the first two cases fail against the unpatched pool (verified via `git stash`).
- `npx vitest run src/gateway` → all green.

## Additional context

None.


Maxwell-Code07:

PR #1439, opened 2026-09-19
fix(gateway): report degraded vector stores in health

### LINKED ISSUE #1432: [Bug] /health reports status=ok while the store is degraded — captures, L1 writes and search silently no-op

zxxwyt:
## Description | 描述

`GET /health` previously considered a vector store healthy whenever its object existed. SQLite deliberately retains a store object after sqlite-vec or schema initialization fails, but all store operations then become no-ops. This PR reports that state accurately.

`stores.vectorStore` now returns `ok`, `degraded`, or `unavailable`. When SQLite initialization caused the degradation, `stores.vectorStoreReason` includes its safe diagnostic reason. The top-level `status` becomes `degraded` unless the vector store is usable.

## Related Issue | 关联 Issue

Fixes #1432

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

Validation:

- `npm test`: 4/4 tests passed, covering unavailable, healthy, degraded-with-reason, and degraded-without-reason stores.
- `npm run build:plugin`: passed.
- `git diff --check`: passed.

## Additional Notes | 其他说明

The health response changes the existing `stores.vectorStore` value from a boolean presence check to an explicit status string. Callers that consume this field should handle the new values. This PR focuses on the gateway's vector store; it does not include the separate StorePool retry work in #1433 or delete-error semantics in #1434.


Maxwell-Code07:

PR #1444, opened 2026-09-19
fix(memory-core): unblock npm run build by dropping dangling build:se…

Nerdlet369:
## Description | 描述

`npm run build` fails on a clean checkout of `MemoryCore`:


Maxwell-Code07:

PR #1453, opened 2026-09-19
fix(panel): allow agent deletion when SkillCore is disabled

### LINKED ISSUE #1218: [Bug] Agent deletion is blocked when SkillCore is disabled

Nerdlet369:
## Description | 描述

`POST /api/v1/agent/delete-cascade` always calls `skill/list` before archiving an agent. When SkillCore is not enabled, MemoryCore returns the explicit capability-absence envelope (`code=404`, `message="Skill module not enabled"`), and the route forwarded it straight back — so agents could never be archived, even when they own no skills at all.

This PR treats **only that exact envelope** as an empty skill collection and continues to archive. The match is exact on both `code` and `message`: any other 404 or error keeps the existing fail-closed behavior, so unrelated upstream failures are never swallowed.

Pagination guard: the capability miss is only interpreted as "no skills" when no page has been returned yet (`offset === 0 && all.length === 0`). Without this guard, a first page of 100 skills with `total=101` followed by the capability miss would archive the agent without deleting those 100 skills.

Adds `MemoryPanel/tests/panel/agent-lifecycle.test.ts` (the module had no tests before) with 6 route-level regression cases.

## Related Issue | 关联 Issue

Fixes #1218

Note: #1219 addresses the same issue but has been a draft since Aug 30 with the pagination-boundary review comment unresolved; this PR includes that guard plus regression coverage. Happy to close mine in favor of #1219 if the author picks it back up.

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

- [x] No existing features affected | 不影响现有功能

`npm test` (6/6), `npm run typecheck` and `npm run build` all pass. The two fail-closed cases confirm existing behavior is unchanged.

## Additional Notes | 其他说明

Verified by mutation testing: removing the capability-match makes the "SkillCore disabled" test fail, and removing the pagination guard makes the "mid-pagination capability miss" test fail — the tests have teeth.


Maxwell-Code07:

PR #1421, opened 2026-09-18
feat(pi): sync mined skills into native Pi skills

kbob3687-hub:
## Summary

- Add `/tdai-memory-sync-skills` to pull mined skills through the current Pi session's Skill Bridge.
- Install validated skill packages into Pi's native `~/.pi/agent/skills/<skill-name>/` directory, including manifest resources such as scripts and templates.
- Preserve hand-written skills. A same-name directory without `tdai-remote.json` is skipped; a different remote skill using an existing managed name is also skipped.
- Make repeat syncs idempotent and reload Pi after one or more skills are installed.
- Harden native sync by rechecking local ownership before replacement, preserving binary resource bytes and executable flags, and fetching all paginated skills.
- Document the five capture roles and their priorities in the Pi adapter README.

## Capture Semantics

This PR reuses the existing five-role mining format:

- `system`: protocol context; Pi's large static prompt is excluded from mining.
- `user`: task intent.
- `tool_call`: ordered supporting evidence, retaining `tool_call_id`.
- `tool_result`: ordered supporting evidence paired with its tool call.
- `assistant`: final outcome.

The user's intent and final assistant response are the semantic anchors. Tool calls and results provide the verified procedure.

## Scope and Isolation

The command uses the current session's Skill Bridge. The client does not send caller-controlled Team or Agent IDs; the Proxy derives user, team, and agent scope from the established Pi session binding.

## Validation

- All applicable unit tests pass: Windows completed 16 tests, while the Unix executable-bit assertion was verified separately in Linux.
- Linux Node 22 container `npm test -- --run __tests__/skill-sync.test.ts` - all 15 tests passed, including executable permission preservation.
- `npx tsc --noEmit` - passed.
- `npm run e2e:skill-sync` - passed with real Pi RPC; verifies command registration, confirmation, paginated list requests, session-scoped Bridge headers, base64 resource transfer, byte-exact native installation, and marker creation.
- `npm pack --dry-run` and `git diff --check` - passed.
- Real Pi + Proxy + Memory Core verification:
  - Completed a tool-using Pi conversation through the real Proxy.
  - Ran `/tdai-memory-sync-skills`: `1 synced`, followed by Pi reload.
  - Verified the native skill directory contains `SKILL.md` and `tdai-remote.json`.
  - Re-ran sync: `1 already current` with no duplicate.
  - Verified a hand-written same-name skill is skipped unchanged.

Closes #1394

PR #1455, opened 2026-09-19
fix(gateway): distinguish store failure from not-found in DELETE /memories

### LINKED ISSUE #1434: [Bug] DELETE /v1/memories returns 200 on store failure — deleteL1 conflates not-found with error

KITE919:
## Summary

Fixes [#1434](https://github.com/TencentCloud/TencentDB-Agent-Memory/issues/1434): `DELETE /v1/memories` returns 200 with count even when store fails, conflating 'not found' with 'store failure'.

## Root Cause

`deleteL1()` in SQLite and TcVDB implementations returned `false` for both:
- Record not found (normal case)
- Store failure (error case)

The gateway batch-delete endpoint turned this boolean into `HTTP 200` with a count, so store failures were reported as success.

### `MemoryCore/src/core/store/sqlite/memory-store.ts`

Removed outer `catch` block from `deleteL1()`. Exceptions now propagate to caller (gateway) for proper error handling. Inner `catch` still handles transaction rollback.

### `MemoryCore/src/core/store/tcvdb/memory-store.ts`

Removed `catch` block from `deleteL1()`. Exceptions now propagate to caller.

### `MemoryCore/src/gateway/v2-router.ts`

Added `try/catch` around `store.deleteL1()` calls in the batch-delete loop. Store failures are now logged as errors but don't block processing of other IDs (non-fatal for batch operations).

## Verification

- MongoDB implementation (`mongodb/memory-store.ts`) already throws on failure — no change needed
- SQLite and TcVDB now match MongoDB behavior
- Gateway logs store failures but continues processing other IDs in batch
- Existing tests should still pass (normal 'not found' case still returns `false`)

## Impact

- `DELETE /v1/memories` now correctly reports HTTP 400/500 on store failure instead of HTTP 200
- Batch delete continues on individual failures (non-fatal)
- No impact on normal 'record not found' case (still returns `false`)

## Related

- Independent of #1421 skill sync
- Part of existing delete endpoint fix


Maxwell-Code07:

Issue #1264, opened 2026-09-05
[Bug] Standalone (non-container) deploy: missing LOG_LEVEL knob, LOG_PATH default disables log rotation, MemoryKnowledge logger still UTC after #129


sysiphysus:
## OpenClaw Version | OpenClaw 版本

Not applicable — this report concerns the standalone deployment of `MemoryCore` + `MemoryKnowledge` (not the OpenClaw plugin path).

## Plugin Version | 插件版本

Reproduced on tag `v2.0.0` (`0aff21a`). **Also confirmed still present on the current default-branch HEAD** (verified by reading `MemoryKnowledge/src/logger.ts` and `MemoryCore/src/gateway/server.ts` from HEAD via the API).

## Operating System | 操作系统

macOS (darwin), Apple silicon.

## System Specification | 系统配置

Standalone (non-container) deployment: two long-running Node services supervised by `launchd` (`KeepAlive=true`, `ThrottleInterval=30`), each started via `npx tsx`. stdout/stderr are redirected to on-disk log files by the service manager (`StandardOutPath` / `StandardErrorPath`). Single host, moderate interactive workload (a few dozen semantic searches per day plus periodic incremental ingestion).

---

## Describe the bug | 问题描述

Three related gaps in the observability layer when running **outside a container**. They are reported together because they interact (see *Expected behavior*): the logging layer appears to assume (a) a container filesystem layout and (b) that verbose DEBUG output is acceptable, with no way for an operator to opt out.

### 1. `MemoryKnowledge` logger emits UTC with no timezone marker, inconsistent with `MemoryCore`

`MemoryKnowledge/src/logger.ts:11` (HEAD):

```ts
function ts() {
  return new Date().toISOString().replace("T", " ").slice(0, 23);
}
```

→ produces `2026-09-05 08:16:21.677` — **UTC, with no timezone marker in the line**.

Meanwhile `MemoryCore/src/gateway/server.ts:140-142` (HEAD):

```ts
function nowLocalIso(): string {
  return dayjs().format("YYYY-MM-DDTHH:mm:ss.SSSZ");
}
```

→ produces `2026-09-05T16:16:21.677+08:00` — **local time with explicit offset**, and the adjacent comment (lines 132-138) explains the intent: the wall-clock should match what the operator sees in `tmux` / `tail -f`, while staying ISO 8601 round-trippable.

[...]

PR #1456, opened 2026-09-19
fix(core): honor LOG_LEVEL in gateway logger

lyxnbclass:
## Summary

- add leveled console logging for the standalone MemoryCore gateway
- honor LOG_LEVEL values debug, info, warn, and error while preserving debug as the default
- document the setting in the English and Chinese MemoryCore READMEs
- add regression coverage for default logging, threshold filtering, and case-insensitive values

## Scope

This addresses the MemoryCore LOG_LEVEL portion of #1264. It intentionally does not change LOG_PATH behavior or MemoryKnowledge timestamp formatting.

## Validation

- npm test
- npm run build:plugin


Maxwell-Code07:

PR #1458, opened 2026-09-20
feat(pi): interactive /tdai-memory-forget with redacted preview and confirm gate

### LINKED ISSUE #1393: [shepherding] `/tdai-memory-forget` — interactive redacted-preview memory deletion
### Is this feature related to a problem? | 该功能需求是否与某个问题相关？请描述

> Shepherding issue split out of #1070.
> Follow-up PR author: @kbob3687-hub. Targets `feat/server_team`.

## Background

The repository currently has no first-class "forget" command. Low-level
delete APIs exist (`sdk/memory-core/.../memory-prompt.delete`,
`MemoryCore/.../skill-core.delete`) but there's no UX layer that lets a
Pi user selectively delete memory entries with a redacted preview before
committing — which is exactly the pattern we want across clients, not
just Pi.

#1070 landed a working `/tdai-memory-forget` interactive flow. It's a
net-new UX primitive worth landing on its own so it can later be lifted
to CodeBuddy / Codex / OpenCode.

## In scope

- Interactive `/tdai-memory-forget` command in the Pi adapter surface
- Redacted preview (show what will be deleted, with sensitive fields
  masked) before commit
- Explicit confirm/cancel gate (destructive-op discipline — same pattern
  as `mem:create-task` / `mem:update-task`)
- Wire-in point: Pi adapter command surface; underlying deletes reuse
  existing SDK/core APIs — no new low-level delete code

## Out of scope

- Modifying `MemoryCore` delete semantics
- Bulk / policy-based forgetting (that's a separate design)
- Lifting the command to other clients — do that in a follow-up once
  the Pi shape has stabilized

## Acceptance criteria

- [ ] Preview shows redacted content, not raw memory
- [ ] Confirm required before any actual delete call
- [ ] Cancel path leaves memory untouched (unit test)
- [ ] Deletes are idempotent (double-confirm ≠ double-delete)
- [ ] No modifications to `handler.ts` or `session/codebuddy/init.ts`
- [ ] Docs include one screenshot / transcript of the preview flow

## Related

- Closes down the `/tdai-memory-forget` portion of #1070
- Reuses `memory-prompt.delete` / `skill-core.delete` — no new deletion
  primitives should appear

## Suggested branching

Branch name: `pi-adapter/tdai-memory-forget` against `feat/server_team`.

### Describe the solution you'd like | 描述你期望的解决方案

/

_No response_

### Additional context | 补充说明

_No response_

kbob3687-hub:
## Summary

- Add `/tdai-memory-forget <keyword>` to the Pi command surface: locate one deletable Memory Prompt or Skill, show a bounded redacted preview, and delete only after explicit confirmation.
- Redact before rendering: known secret patterns plus values bound to sensitive field names, truncated on a UTF-8 boundary to a 320-byte budget.
- Gate the destructive call behind `ui.confirm`; the preview path issues no delete request, and headless Pi fails closed before sending anything.
- Make deletion idempotent: one pending action per session, single-flight execution, and a repeated confirmation returns success without deleting again.
- Derive identity from the established Pi session rather than caller-supplied IDs.
- Reuse the existing `/v3/memory-prompt/delete` and `/v3/skill/delete` APIs; add no new deletion primitive and change no MemoryCore delete semantics.
- Document the command, the preview transcript, and a real-stack E2E script in the Pi adapter README.

## Refinements after self-review

Four follow-up commits (a net −212 lines) simplify the first landing:

- the pending store keeps three states (`ready` / `executing` / `completed`) instead of five; `cancelled` and `uncertain` did not provide a guarantee the other states did not already give;
- the second preview round-trip and its discovery cache are gone — the first search hands every candidate its own action id, so the client confirms directly;
- the per-session "current action" table is gone, so one session can hold several independent pending actions;
- `impact` is no longer shipped from the server: it is a pure function of the object kind, so the Pi UI derives it.

**One deliberate deviation from the issue text worth flagging.** The issue asks for a confirm/cancel gate "same pattern as `mem:create-task` / `mem:update-task`", and that pattern exposes a `cancel` endpoint. Here cancellation is client-side: declining the dialog simply never confirms, and the action expires by itself after five minutes. The acceptance criterion ("cancel path leaves memory untouched", with unit coverage) still holds — `previews and cancels without calling delete` and `prepares selectable actions without deleting during discovery` cover it — and removing the endpoint also removed the `cancelled` state that existed only to serve it. If you would rather keep the endpoint shape identical to the task commands, restoring it is a small addition.

## Redaction and Preview Bounds

- Candidate names are redacted inside the service **and** again when a preview or confirmation response is rendered, so an alternative route implementation cannot bypass the boundary.
- Skill previews are built from the full `description`, never from Core's already-clipped search snippet — otherwise a secret straddling the snippet boundary would survive as a fragment.
- Secrets are also matched by field name (`password`, `passwd`, `pwd`, `secret`, `api_key`, `access_key`, `client_secret`, `authorization`, `credential`, `private_key`) in JSON, colon and equals forms.
- Truncation is UTF-8 safe and marks clipped previews with `[truncated]`.

## Confirm Gate and Idempotency

- A five-minute pending action keyed by session moves `ready → executing → completed`.
- Exactly one caller starts the delete; concurrent confirmations await the same in-flight promise.
- A completed action answers later confirmations without deleting a second time.
- An expired action is rejected, and a failed delete removes the action so the next attempt must start from a fresh preview — Core remains the authority on what still exists.

## Scope and Isolation

- Identity is derived from the initialized session; Team/Agent/User IDs in the request body are ignored.
- Skills: current `(team_id, agent_id)` owner scope, active versions only.
- Memory Prompts: agent-private prompts effective for the current agent's L1/L2/L3 layers, built-ins excluded. A prompt shared with another agent is refused at discovery **and** re-checked at execution time.
- The command routes are not advertised in the injected prompt and are not added to either bridge allowlist; the proxy keeps all Core credentials.
- Pending actions are in-process, so a multi-replica proxy would need sticky routing; verified here on single-instance deployments.

## Validation

- `npm test` (MemoryProxy) — 4 files, 21 tests passed.
- `npm test` (MemoryCore/pi-plugin) — 2 files, 7 tests passed.
- `npm run build` (MemoryCore/pi-plugin) — `tsc` clean.
- `git diff --check` — clean.
- `MemoryCore/pi-plugin/scripts/forget-e2e.ts`, against a real Memory Core, a real Proxy, and a real Pi source checkout's extension loader (fixtures created in isolation and removed afterwards):
  - cancel keeps the skill; a redacted confirmation deletes it; a repeated confirmation returns 200 without a second delete;
  - every candidate carries its own action id, and tampered or cross-session action ids are rejected;
  - auth fails closed; one confirmation deletes only its own target;
  - two concurrent confirmations delete once.
- Installed **Pi 0.84.1** launched in RPC mode with this branch's extension (verified locally): Pi reports the command as registered by `MemoryCore/pi-plugin/index.ts`; the confirmation dialog it actually renders carries `[REDACTED]` and no raw secret; declining leaves the Skill in Core; accepting deletes exactly that Skill.

Closes #1393


kbob3687-hub:
Size note: 706 lines of implementation, 1016 lines of tests (596 unit + 420 real-stack E2E), 71 lines of docs. The follow-up commits reduce implementation by 18% versus the first landing; the line delta grew only because the E2E script is new.

Maxwell-Code07:

PR #1463, opened 2026-09-20
fix(memory-core): support string-only content in strict LLM endpoints

### LINKED ISSUE #1313: L2/L3 tool-loop fails with 400 "Bad input" on Cloudflare Workers AI (and other strict OpenAI-compatible endpoints)
### Summary

When `llm.baseUrl` points at **Cloudflare Workers AI**'s OpenAI-compatible endpoint (`/ai/v1/chat/completions`), every multi-round tool call fails with HTTP **400 Bad Request**. Since L2 scene extraction and L3 persona generation drive the LLM through file tools (`read_file` / `write_file`) over multiple rounds, **L2/L3 effectively never succeed** on this provider — scenes are barely created/updated and the persona stays thin, while L1 (single-shot, no tools) works fine.

### Environment

- Standalone / service (worker) mode, `llm.provider` = OpenAI-compatible, `llm.baseUrl` = Cloudflare Workers AI (`.../ai/v1`), a tool-capable model (e.g. `@cf/qwen/qwen3-30b-a3b-fp8`).
- LLM runner: `StandaloneLLMRunner` (Vercel AI SDK, `ai` + `@ai-sdk/openai`, `generateText` with tools).

### Symptom

`[extractor] extract() LLM runner failed: Bad Request` / `[standalone-runner] run() failed: Bad Request`. The first round (plain user message) succeeds; the failure starts on the **second** round, once the message history contains tool-call / tool-result turns.

### Root cause

The AI SDK's OpenAI provider serializes messages per the OpenAI spec:

- assistant tool-call messages carry `content: null`;
- message `content` is sent as a **structured array** of parts (e.g. `[{type:"text", text:"..."}]`), including tool-result messages.

Real OpenAI and most mature OpenAI-compatible servers accept both. **Cloudflare Workers AI's endpoint applies a stricter request schema and rejects them**, returning:

```
AiError: Bad input: Error: oneOf at '/' not met, 0 matches:
  required properties at '/' are 'prompt',
  Type mismatch of '/messages/0/content', 'array' not in 'string',
  Type mismatch of '/messages/2/content', 'string' not in 'null',
  required properties at '/messages/2' are 'role,content', ...
```

i.e. it wants `content` to be a **string** and does not allow `null`.

### Minimal repro

Point `llm.baseUrl` at CF Workers AI and issue any tool-using request that goes past one round (e.g. trigger one L2 scene extraction on a scope that already has scenes). A direct repro against the endpoint:

```jsonc
// 200 OK — single string content + tools
{"model":"…","messages":[{"role":"user","content":"hi"}],"tools":[…]}

// 400 — array content and/or null content (what the AI SDK emits mid-tool-loop)
{"model":"…","messages":[
  {"role":"system","content":[{"type":"text","text":"…"}]},
  {"role":"assistant","content":null,"tool_calls":[…]},
  {"role":"tool","content":[{"type":"text","text":"…"}],"tool_call_id":"…"}
]}
```

### Scope

This is **specific to strict OpenAI-compat endpoints** (Cloudflare Workers AI is the concrete case); it is not a general provider problem — real OpenAI, vLLM, DeepSeek, Together, Ollama's `/v1`, etc. accept the AI SDK's output. So it does not affect users on those backends, but it silently breaks L2/L3 for anyone using CF Workers AI as the LLM.

### Possible fixes (for discussion)

1. **Normalize outbound messages for strict endpoints**: coerce array `content` → concatenated text and `null` → `""` before the request (behind a provider flag / `strictOpenAICompat` option), keeping `tool_calls` intact. Lowest-risk if opt-in.
2. Document CF Workers AI as needing such a shim (a sidecar proxy that rewrites the request body works today and keeps the core runner untouched).
3. Consider a runner-level compatibility layer keyed off the base URL / a config switch.

Happy to send a PR for (1) as an opt-in normalization if the maintainers prefer that shape.

zxxwyt:
## Description | 描述

Strict OpenAI-compatible endpoints can reject the second request in an L2/L3 tool loop because the AI SDK serializes assistant tool calls with `content: null`. Add opt-in `llm.strictOpenAICompat` (gateway environment override: `TDAI_LLM_STRICT_OPENAI_COMPAT=true`) to convert null to an empty string and join text-only content arrays before sending the request.

Tool call IDs, arguments, tool results, headers and abort signals are preserved. Mixed or unknown content parts are left intact to avoid silently dropping multimodal data. The default leaves the SDK's fetch path unchanged; the option works with both streaming and non-streaming standalone runners.

Carry the setting through gateway/plugin parsing, runtime provider resolution, and explicit seed/skill runner construction. Preserve `llm.stream` when copying gateway settings into `memory.llm`, so selecting the compatibility runner does not lose the streaming setting. Document configuration and limitations in both READMEs and the example YAML.

## Related Issue | 关联 Issue

Fixes #1313

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

- [x] No existing features affected | 无影响现有功能 — default-disabled request behavior and multimodal preservation covered by tests.

Validation in `MemoryCore`:

- `npm test`: 2 files, 16 tests passed. Includes the real AI SDK's two-step tool loop against a strict mock fetch endpoint, streaming tool calls, content normalization, preservation of unsupported payloads, configuration defaults, environment precedence, and proxy resolution.
- Negative control: removing the runner's compatibility fetch hook makes the enabled two-step test fail with HTTP 400 (`content must be a string`); restoring it passes.
- `npm run build:plugin`: passed.
- `npx tsc --noEmit --strict --skipLibCheck --target es2022 --module nodenext src/adapters/standalone/strict-openai-compat.ts`: passed (new helper only).
- `git diff --check`: passed.

## Additional Notes | 其他说明

No live Cloudflare account was used; provider behavior is modeled by the strict mock endpoint. This addresses text-content serialization only, not every provider-specific schema restriction. No dependency changes.


Maxwell-Code07:

PR #1464, opened 2026-09-20
fix(core): enforce caller-scoped ACL on asset/get and asset/list

Xuruida:
## Description | 描述

`asset/get` and `asset/list` previously ran with an unrestricted admin view:
any team member holding a valid `user_key` could enumerate and read other
members' `visibility=private` assets, including sensitive
description/metadata fields (horizontal privilege escalation).

Fix at the kernel meta layer (defense in depth — no longer relies on the
caller-side whitelist alone):

- `getAssetForCaller`: owner always allowed; other callers go through
  `checkAssetPermission` (visibility → ACL); denied → null → 404 so asset
  existence is not leaked; `system_admin` stays exempt as the admin-plane
  break-glass path (same style as `user/list`)
- `listAssetsForCaller`: non-members get an empty page; team members get
  their own assets plus permission-filtered results, filtered before
  pagination so totals stay accurate; `system_admin` keeps the full
  management view

Internal (non-HTTP) callers of `getAssetById` / `listAssetsByTeam` are
untouched — the fix only wraps the two HTTP routes.

## Related Issue | 关联 Issue

Security hardening reported via external security review (horizontal
privilege escalation on meta asset read APIs).

## Change Type | 修改类型

- [x] Bug fix | Bug 修复
- [ ] New feature | 新功能

## Self-test Checklist | 自测清单

Verified with an end-to-end privilege-escalation matrix (two-member team,
one `private` asset owned by alice, attacker bob as a plain member):

| Case | Before | After |
|---|---|---|
| bob `asset/list` {team_id} | alice's private asset leaked | empty ✅ |
| bob `asset/get` {asset_id} | 200 with full description | 404 (existence hidden) ✅ |
| alice `asset/get` own private asset | 200 | 200 ✅ |
| alice `asset/list` own team | full | full ✅ |
| system_admin `asset/list` | full view | full view (exempt) ✅ |
| `asset/list-accessible` (member-safe API) | unchanged | unchanged ✅ |
| missing `x-tdai-user-key` | 400 | 400 ✅ |

Panel-side callers (`chat-memory` lists with `visibility:"team"` /
`owner_user_id:self` filters, knowledge detail on team-visible assets) are
semantically unaffected.

## Additional Notes | 其他说明

Behavior change is strictly a restriction: reads that previously succeeded
for non-owner/non-authorized callers on `private` assets now return 404 /
filtered results. Owner, team-visible, and admin-plane reads are unchanged.


Maxwell-Code07:

PR #1387, opened 2026-09-17
feat: PostgreSQL storage backend (pgvector + tsvector) with Skill module alignment

hlwang123:
## Summary

Adds a fourth storage backend, **`postgres`**, alongside `sqlite` / `tcvdb` / `mongodb`, so Memory Core can run fully on a self-hosted PostgreSQL 16 + pgvector instance — no cloud vector DB required. This also closes the Skill-module gap: previously skill wiring was skipped on non-SQLite backends; it now works on PostgreSQL with a store that is semantically 1:1 aligned with `SqliteSkillStore`.

Deployment scripts, DDL and E2E suites for this backend live in [hlwang123/TencentDB-Agent-Memory-PG](https://github.com/hlwang123/TencentDB-Agent-Memory-PG); this PR contains the MemoryCore source-side changes.

## Changes

| File | Change |
|------|--------|
| `store/postgres/memory-store.ts` | **NEW** — full `IMemoryStore` on PostgreSQL: pgvector IVFFlat (cosine) for vectors, `tsvector` + GIN for full-text (jieba pre-segmentation, `websearch_to_tsquery`), all L0/L1/entity/audit methods. Adds a `getPgPool()` escape hatch (exposes the `pg.Pool` + dimensions), role-equivalent to SQLite's `getRawDb()` |
| `store/postgres/skill-store.ts` | **NEW** — `PgSkillStore`: multi-version rows with `is_head`, per-skill write serialization via `pg_advisory_xact_lock`, FTS via generated `tsvector` column, vectors via `skill_vec` table, plus embedding / RRF hybrid retrieval paths |
| `core/tdai-core.ts` | Skill wiring: when the memory store exposes `getPgPool()`, construct `PgSkillStore` on the shared pool (same connection as memory data) |
| `config.ts` | `StoreBackend` adds `"postgres"`; new `PostgresConfig { connectionString }` + `memory.postgres` parsing |
| `core/skill/types.ts`, `skill-config.ts` | Skill-side store-backend unions accept `"postgres"` (it flows through config only — the store itself is wired via `getPgPool()`) |
| `store/factory.ts` | `case "postgres"`: application-layer embedding service (same wiring as sqlite) + store snapshot |
| `store/store-pool.ts` | `StoreMode` adds `"postgres"`, `createPostgresStore()`, `getStore` branch + config fingerprint |
| `utils/manifest.ts` | Manifest snapshot for postgres (connection string persisted with the **password redacted**) |
[...]

PR #1466, opened 2026-09-21
feat(metadata): PostgreSQL metadata store backend (schema per instance)

hlwang123:
## Summary

Follow-up to #1387 (postgres memory/skill store backend). With that PR, the memory plane can run fully on PostgreSQL, but the v3 metadata plane (users / user keys / teams / agents / tasks / assets / fixed-asset bindings / ACL / config params / upstream config) still silently fell back to SQLite. This PR closes the gap: setting `TDAI_METADATA_POSTGRES_URI` moves the metadata plane to PostgreSQL as well, so a PG deployment no longer depends on SQLite at all.

Stacked on #1387 — please merge that first; the diff here will then collapse to the metadata changes only.

### What's included

- **`PostgresMetadataStore`** (`src/metadata/store/postgres-adapter.ts`): implements all ~70 `IMetadataStore` methods, method-for-method aligned with `sqlite-adapter.ts`:
  - instance isolation via one schema per instance (`{prefix}_{sanitized_instance_id}`, truncated to PG's 63-byte identifier limit); `purgeInstance` = `DROP SCHEMA ... CASCADE`, mirroring the MongoDB db-per-instance model
  - SQL dialect port: `AUTOINCREMENT` → `GENERATED BY DEFAULT AS IDENTITY`, `INSERT OR IGNORE` → `ON CONFLICT DO NOTHING`, existing `ON CONFLICT ... DO UPDATE` upserts kept as-is
  - unique-violation handling via SQLSTATE `23505` + constraint names; `DuplicateUserKeyError` semantics (explicit `default_key_value` collision → HTTP 409) preserved; PK / relation-id collision retries preserved
  - composite writes (createUser + default key, createTeam + auto admin member, createTask + linkAgents, setAgentFixedAssets full replace) run in transactions on a single pooled connection
- **Factory / config wiring**:
  - `TDAI_METADATA_POSTGRES_URI` env (also gateway yaml `metadata.store.postgresUri`, backfilled via `metadata-env.ts`)
  - 3-way exclusivity: at most one of postgres / mongo / sqlite env vars may be set explicitly
  - `MetadataStorePool` shares one `pg.Pool` across instances (ownsPool=false), standalone `createMetadataStore` owns its pool
  - `deployMode=service` now accepts postgres in addition to mongodb
  - api-trace resolves the `full` profile for postgres (server-class backend), same as mongodb

## Test plan

- [x] Full `runMetadataStoreContract` suite against local `postgres:16`: **58/58 passed** (same suite that validates the sqlite/mongodb adapters)
- [x] Factory smoke: env resolution → `backend=postgres`; 3-way exclusivity error; same-instance store reuse; cross-instance schema isolation; `purgeInstance` drops only the target schema; `closeAll`
- [x] Strict `tsc` error-set diff vs clean worktree at base commit: no new type errors (repo has pre-existing strict-mode errors; verified identical sets)

Local verification helpers (contract wrapper + smoke script) are intentionally not committed, matching the repo's convention of keeping `*.test.ts` backend wrappers out of the OSS tree.


Maxwell-Code07:

